PR #101, opened 2015-09-08
Added ipython notebook to simulate data for allele-fraction model.

samuelklee:


samuelklee:
I'm going to close this without merging as the generative model is a bit out of date.  I believe @davidbenjamin has up-to-date code in his db_acs_prototype branch.

PR #132, opened 2015-10-07
Adjusting caller to take in log2 data from the targets.  Also, added …

### LINKED ISSUE #131: Caller is expecting targets in CR, but needs to be logCR.

LeeTL1220:
…some utility functions.

Closes #131


src/main/java/org/broadinstitute/hellbender/tools/exome/ReCapSegCaller.java
```diff
@@ -31,11 +33,12 @@
     public static final double DEFAULT_Z_SCORE_THRESHOLD = 2.0;
 
     //bounds on coverage we take as non-neutral a priori
-    public static final double NON_NEUTRAL_THRESHOLD_HIGH = 1.25;
-    public static final double NON_NEUTRAL_THRESHOLD_LOW = 0.75;
+    public static final double NON_NEUTRAL_THRESHOLD_HIGH = ParamUtils.log2(1.25);
+    public static final double NON_NEUTRAL_THRESHOLD_LOW = ParamUtils.log2(.75);
```
davidbenjamin: 1.25 and 0.75 have no inherent significance.  Originally there was a single threshold parameter (call it x) in log space.  The symmetric range [-x, x] maps to [2^(-x), 2^x] in real space, which requires two parameters to specify since it's not symmetric about zero.  Now that we're going back to log-space, we should restore this threshold to a single parameter with a range that is symmetric about zero.  (It could well be that the distribution in log space is not symmetric about zero, but *a priori* we don't know this).

src/main/java/org/broadinstitute/hellbender/tools/exome/ReCapSegCaller.java
```diff
@@ -31,11 +33,12 @@
...
 
     //bounds on coverage for high-confidence neutral segments
-    public static final double COPY_NEUTRAL_CUTOFF = 1.1;
+    public static final double COPY_NEUTRAL_CUTOFF_HIGH = ParamUtils.log2(1.1);
```
davidbenjamin: Ditto.

src/main/java/org/broadinstitute/hellbender/tools/exome/ReCapSegCaller.java
```diff
@@ -65,7 +68,7 @@ private ReCapSegCaller() {} // prevent instantiation
 
         // Get the standard deviation of targets belonging to high-confidence neutral segments
         final double[] nearNeutralCoverage = segments.stream()
-                .filter(s -> SegmentUtils.meanTargetCoverage(s, targets)/neutralCoverage < COPY_NEUTRAL_CUTOFF && SegmentUtils.meanTargetCoverage(s, targets)/neutralCoverage > 1/COPY_NEUTRAL_CUTOFF)
+                .filter(s -> (SegmentUtils.meanTargetCoverage(s, targets) - neutralCoverage) < COPY_NEUTRAL_CUTOFF_HIGH && (SegmentUtils.meanTargetCoverage(s, targets) - neutralCoverage) > COPY_NEUTRAL_CUTOFF_LOW)
```
davidbenjamin: Once you have a single cutoff parameter as above, the filter can be written as a single predicate using an absolute value.

src/main/java/org/broadinstitute/hellbender/utils/param/ParamUtils.java
```diff
@@ -105,4 +105,27 @@ public static double isFinite(final double val, final String message) {
...
+
+    /**
+     * log the value in base 2
+     *
+     * @param val -- The value to be logged.
+     * @return the logged value.  If val is zero, returns Double.NEGATIVE_INFINITY.  If val
+     *  is negative, returns Double.NaN.
+     */
+    public static double log2(final double val) {
+        return Math.log(val)/Math.log(2.0);
```
davidbenjamin: I leave it up to you to decide whether this will be used heavily enough to justify caching 1/log(2).

davidbenjamin:
@LeeTL1220 Review complete.

src/main/java/org/broadinstitute/hellbender/tools/exome/ReCapSegCaller.java
LeeTL1220: Rolled it back to previous version.

Done.

src/main/java/org/broadinstitute/hellbender/tools/exome/ReCapSegCaller.java
LeeTL1220: Rolling it back addressed this.

src/main/java/org/broadinstitute/hellbender/tools/exome/ReCapSegCaller.java
LeeTL1220: Rolling it back addressed this.

src/main/java/org/broadinstitute/hellbender/utils/param/ParamUtils.java
LeeTL1220: It's already in there with passing unit tests.  I think we'll use it in the future.

LeeTL1220:
@davidbenjamin comments addressed and pushed.

PR #138, opened 2015-10-08
Added GibbsSampler and univariate, unimodal SliceSampler for Markov-Chain Monte-Carlo sampling, along with test using simple copy-ratio model.

### LINKED ISSUE #115: Implement slice sampler.
Could be possible to use this for other Bayesian MCMC analyses in hellbender, too.

samuelklee:
Also fixed miscellaneous typos.

Closes #115.


samuelklee:
@davidbenjamin actually, you can go ahead and start reviewing this if you want.  Again, most of the meat is in GibbsSampler and SliceSampler (and their corresponding test classes); changes in the rest of the classes are mostly minor.

davidbenjamin:
@samuelklee I know you're in the middle of refactoring so I'll hold off on the line-by-line, but here are some thoughts.

* It seems like `GibbsSampler` is responsible for too much.  I think there are one or two classes to be extracted, such as one corresponding to a node of the graphical model.  From the new code in the last 24 hours it looks like you're already going in this direction.
*  (subjective!) It would feel more right to me if the nodes, not the `GibbsSampler` own the method for updating themselves.
* (still subjective) Likewise, the `List<List<List<` doesn't sit well with me and one solution would be if the nodes own their history of samples, and also a bool for whether to keep samples at all.  the `GibbsSampler` could query the nodes for samples.
* (Ditto about the subjectivity-fest) I feel like the distinction between global and local sampling won't be flexible enough for many hierarchical models.  My hunch is that nodes should know only about other nodes in their Markov blanket and that the size of this Markov blanket essentially contains the global/local distinction.  There are difficulties with this approach that I don't even remotely claim to have answers for, however. . .
* After reading a lot of MCMC APIs, I find Mamba.jl (written in Julia) to be quite admirable and relevant to our purposes.

samuelklee:
all good ideas, thanks!  might try to get a bit more refactoring done tomorrow.  i do think it'd be nice to get the ball rolling on some of the likelihood classes with a working-but-hacky version, but it's probably better to refactor now rather than later.  thanks for your patience!

samuelklee:
getting closer to its final form (well, at least for a first cut).  still need to add javadoc.  sorry for all the annoying emails from coveralls...

samuelklee:
closing this one too, so coveralls can chill out

PR #154, opened 2015-10-19
Added GibbsSampler and univariate, unimodal SliceSampler for Markov-Chain Monte-Carlo sampling, along with tests using Gaussian and simple copy-ratio models.

### LINKED ISSUE #115: Implement slice sampler.

samuelklee:
Also fixed miscellaneous typos.

Closes #115.

@davidbenjamin just checking out coverage for now.  Will add some more tests covering various exceptions tonight and then you can proceed to review!


src/main/java/org/broadinstitute/hellbender/utils/mcmc/AbstractParameterizedState.java
```diff
@@ -0,0 +1,127 @@
...
+ * Represents a collection of named {@link Parameter} objects, which may hold values of mixed type.
+ * See GibbsSamplerSingleGaussianUnitTest and GibbsSamplerCopyRatioUnitTest for examples of use.
+ *
+ * @author Samuel Lee &lt;slee@broadinstitute.org&gt;
+ */
+public abstract class AbstractParameterizedState {
+    private final Map<String, Parameter<?>> parameterMap = new HashMap<>();
+
+    /**
+     * Constructs an AbstractParameterizedState from a List of {@alink Parameter} objects with values of mixed type.
```
davidbenjamin: javadoc typo `@link`

src/main/java/org/broadinstitute/hellbender/utils/mcmc/AbstractParameterizedState.java
```diff
@@ -0,0 +1,127 @@
...
+     * @param parameters    List of Parameters
+     * @throws IllegalArgumentException if {@code parameters} is null or contains duplicate parameter names
+     */
+    public AbstractParameterizedState(final List<Parameter<?>> parameters) {
```
davidbenjamin: Is there an MCMC state that is *not* parameterized?  If not, I feel `AbstractMCMCState` etc would be a more evocative name for this class.

src/main/java/org/broadinstitute/hellbender/utils/mcmc/AbstractParameterizedState.java
```diff
@@ -0,0 +1,127 @@
...
+     * Returns the value of a {@link Parameter} contained in the collection held by the AbstractParameterizedState,
+     * given the parameter name and type.
+     * @param parameterName         name of Parameter
+     * @param parameterValueClass   class of Parameter value
+     * @param <T>                   type of Parameter value
+     * @return                      Parameter value
+     * @throws IllegalArgumentException if {@code parameterName} does not correspond to a Parameter in the collection or
+     *                                  if {@code parameterValueClass} does not match that of the Parameter in the collection
+     */
+    public <T> T get(final String parameterName, final Class<T> parameterValueClass) {
```
davidbenjamin: As a check of my understanding: the ugly but concise way of doing this would be to have this return `Object`, thereby obviating the second argument but forcing you to use a bunch of casts downstream.  A compromise would be to basically wrap `Object` in a `ParameterValue` class, but you opted against this because it wouldn't be type-safe regarding different types of parameter values.  Do I grok?

samuelklee: That's right.  I tried to minimize the number of casts (and more generally, the number of times the parameterized type of any value (parameter or data) that you're trying to set or get had to be specified).  But since these types get erased when specifying generic getters/setters/etc., you have to either explicitly pass the class or suppress warnings.  I chose the former, but I'm not sure this is a complete solution---e.g., if you try to make the parameter value something like a List<Integer> you're going to run into issues...(although maybe this is a feature, not a bug.

I think these sorts of issues are unavoidable if we try to make collections of mixed type like ParameterizedState and DataCollection, but perhaps there's some more clever way to do it.  We also do lose a bit of performance because of all the casting.

src/main/java/org/broadinstitute/hellbender/utils/mcmc/AbstractParameterizedState.java
samuelklee: That's right, you can use them to represent parameter blocks.  Not a big deal to change up the name, though (that goes for the other classes as well).

src/main/java/org/broadinstitute/hellbender/utils/mcmc/GibbsSampler.java
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * Returns a list of samples for a specified model parameter, discarding the first {@code numBurnIn} samples.
+     * @param parameterName         name of parameter
+     * @param parameterValueClass   class of parameter value
+     * @param numBurnIn             number of burn-in samples to discard from beginning of chain
+     * @param <U>                   type of parameter value
+     * @return                      List of parameter samples
+     */
+    public <U> List<U> getSamples(final String parameterName, final Class<U> parameterValueClass, final int numBurnIn) {
```
davidbenjamin: Any reason you're calling the generic `U` and not `T` here?

src/main/java/org/broadinstitute/hellbender/utils/mcmc/AbstractParameterizedState.java
samuelklee: fixed in latest commit, thanks!

src/main/java/org/broadinstitute/hellbender/utils/mcmc/GibbsSampler.java
samuelklee: tried to be consistent about using S/S1/S2 for AbstractParameterizedState, T/T1/T2 for DataCollection, and U for parameter values...if this is confusing and doesn't add to the readability i can change this to T

samuelklee:
OK, @davidbenjamin, have at it!

src/test/java/org/broadinstitute/hellbender/utils/mcmc/AbstractParameterizedStateTest.java
```diff
@@ -0,0 +1,62 @@
+package org.broadinstitute.hellbender.utils.mcmc;
```
davidbenjamin: This class needs a few tests without expected exceptions.

src/test/java/org/broadinstitute/hellbender/utils/mcmc/GibbsSamplerSingleGaussianUnitTest.java
```diff
@@ -0,0 +1,181 @@
...
+ *     python package emcee (see http://dan.iel.fm/emcee for details) and numerical evaluation in Mathematica
+ *     of the analytic forms of the posteriors.
+ * </p>
+ *
+ * @author Samuel Lee &lt;slee@broadinstitute.org&gt;
+ */
+public final class GibbsSamplerSingleGaussianUnitTest extends BaseTest {
+    private static final String TEST_SUB_DIR = publicTestDir + "org/broadinstitute/utils/mcmc/";
+
+    private static final File COVERAGES_FILE = new File(TEST_SUB_DIR
```
davidbenjamin: Instead of hard-coding a test file, why not generate draws from a normal distribution on the fly?  The cost of sampling is probably less than I/O, so the test wouldn't take any longer.  In my opinion a bigger advantage is to make this test self-contained.

samuelklee: Good point, I can change this.  Hmm, but maybe one argument for loading from a file is that this simple example of the "DataCollection construction step" will be analogous to the slightly more complicated case in the CopyRatio test (where, I would argue, generating random data on the fly would be complex enough to merit its own tests).  On the other hand, in practice, I'm not sure if the naive Data.loadData method will actually be used (probably construction by passing Lists derived from files with headers, columns, etc. and loaded by other file-reading methods will be the typical use case).  Thoughts?

samuelklee:
Closes #160 (in the first commit).

src/main/java/org/broadinstitute/hellbender/utils/mcmc/GibbsSampler.java
davidbenjamin: Your convention sounds good.

src/test/java/org/broadinstitute/hellbender/utils/mcmc/GibbsSamplerSingleGaussianUnitTest.java
davidbenjamin: In my opinion the responsibility for pedagogy belongs to `DataCollection`'s unit tests and the (eventual) `ACSModel`'s integration test.  Also, I agree with your second paragraph.

src/test/java/org/broadinstitute/hellbender/utils/mcmc/SliceSamplerUnitTest.java
```diff
@@ -0,0 +1,55 @@
...
+     * by 20000 samples to a relative error of 0.05% and 2%, respectively.
+     */
+    @Test
+    public void testSliceSamplingOfNormalDistribution() {
+        final double mean = 5.;
+        final double standardDeviation = 0.75;
+        final NormalDistribution normalDistribution = new NormalDistribution(mean, standardDeviation);
+        final Function<Double, Double> normalLogPDF = x -> normalDistribution.logDensity(x);
+
+        final double xInitial = 4.5;
```
davidbenjamin: Make xInitial several standard deviations away from the mean because a slice sampler should be robust against bad initial guesses.  To be honest I can't think of any coding error that would fail in this way without failing in general, but it just seems right to stack the test against the code.

src/test/java/org/broadinstitute/hellbender/utils/mcmc/GibbsSamplerCopyRatioUnitTest.java
```diff
@@ -0,0 +1,412 @@
+package org.broadinstitute.hellbender.utils.mcmc;
```
davidbenjamin: I suppose this is going to become the `CopyRatioModelIntegrationTest` some time soon?

davidbenjamin: I would say that the data should be generated on the fly and not loaded, but for the future integration test that I'm assuming loading data seems more right, so it makes sense to keep it as it is here.

davidbenjamin:
@samuelklee This looks great.  My only substantive comment is generating data for the Gaussian model unit test on the fly.

src/test/java/org/broadinstitute/hellbender/utils/mcmc/GibbsSamplerCopyRatioUnitTest.java
samuelklee: I think I'll keep this separate as a simple example of the API usage and will have a separate test for the full copy-ratio model.  I considered lumping both the SingleGaussian and CopyRatio Tests together into GibbsSamplerUnitTest but that was quite long and sprawling...I know it's not standard to have two separate Test classes but I think it's warranted here.

samuelklee:
Thanks, back to you @davidbenjamin!  Note I realized that I was a bit confused about deep vs. shallow copies; see the updated javadoc.

PR #162, opened 2015-10-21
ACS docs for collapsed sampling of hyperparameters

davidbenjamin:


davidbenjamin:
@samuelklee Just a few new equations in the proposed methods.  Should be a quick review.

samuelklee:
looks good!

PR #173, opened 2015-11-02
Moving from hg19 to b37

### LINKED ISSUE #172: Need to reconcile b37 and hg19, per Geraldine's request
Details are found in the discussion on  #167  ...

Basically, we take in hg19, but we act as if it was b37.  This is how Broad CGA operates, but we do not want to do that for tools that we will expose externally.  

@vdauwera Anything to add?

LeeTL1220:
Closes #172


vdauwera:
:+1: 

FYI I'm looking into having the production ref file name corrected separately.

LeeTL1220:
Travis fail is due to Travis, not anything in this pull request.

PR #174, opened 2015-11-02
Added ACS MCMC copy-ratio modeller.

### LINKED ISSUE #127: Implement copy-ratio model for ACS.
Simple model that fits Gaussians to coverages in each segment, with a global parameter controlling the variances. This is basically already done in GibbsSamplerUnitTest and just needs to be repackaged.

### LINKED ISSUE #170: Refactor classes for MCMC sampling.
The classes abstracting the data and the parameterized states are somewhat overengineered.  Some possible changes:

1) It could be possible to remove DataCollection in favor of an empty interface, leaving it to the user to correctly specify the class.

2) Right now, two different types of classes inherit from AbstractParameterizedState; those that represent "parameterized blocks" and those that represent the parameterized state that is updated by the GibbsSampler.  Only the latter require the implementation of the "copy" method; the former might be better abstracted as "ParameterCollection" or something similar, which the latter can then extend.

Probably best to get the copy-ratio and allele-fraction models working using the current implementation then plan on going back and refactoring everything at once (along with the relevant test code).

samuelklee:
Also refactored DataCollection and corresponding tests in the first commit.

Closes #127.
Closes #170.


davidbenjamin:
Just one annoying Java grammar comment on the first commit, moving on to the second.

davidbenjamin:
Back to @samuelklee

samuelklee:
thanks, i think that's everything---back to you, @davidbenjamin

PR #179, opened 2015-11-04
(Don't merge until test pass) Adds a TargetWalker for those tools that iterate thru targets.

vruano:
Adds a toy target read coverage which evenatual will do the counting for the germline implementation
Updates the GATK4 public dependency to bring in some bug fixes.


vruano:
@samuelklee  please review.

droazen:
Woohoo, finally some new walker types :) Would love to see the new walker base classes eventually migrate into https://github.com/broadinstitute/gatk so that they can be an official part of the GATK engine.

samuelklee:
OK, done with my review.  mostly superficial things.  however, i think coverage could be higher (since travis has been failing lately, i'm not sure what the exact drop was from master)---could use some more tests.  back to @vruano!

vruano:
Back to @samuelklee 
Test failing but possibly due to travis hiccup.

vruano:
Back to @samuelklee

samuelklee:
just a couple more minor comments, otherwise good to merge after squashing, tests pass, etc.!

vruano:
Test pass locally and in gsa5. Will give travis a chance to complete and the merge even if it does not due to due to a build fail before `gradle test`.

Issue #142, opened 2015-10-14
Per target complexity calculation.


vruano:
Calculates the complexity of a target.

The complexity is measure as the number of positions in the target that are masked as repeated in the genome

The details can be obtain from following up the instruction to calculate complexity in XHMM tutorial page:
 
https://atgu.mgh.harvard.edu/xhmm/tutorial.shtml

Issue #141, opened 2015-10-14
Per target GC bias.


vruano:
Tool to calculate the GC bias per target.

PR #190, opened 2015-11-17
(Don't merge before test pass) Adds a new tool to annotate targets with target specific properties.

vruano:
Adds a new tool to annotate targets with target specific properties.

This addresses issues #141 and #142.

The first two annotations supported are

* GC_BIAS = Fraction of GC / Total number of sense bases in the target reference sequence.

  - The user need to provide a reference.

* REPEAT_FRACTION = Fraction of base in the target masked as repeated somewhere else in the referece.

  - The user need to provide a BED file (indexed) with the repeated regions.

Annotations whose dependence is not provided won't be present in the output.

Syntax:

  java -jar hellbender-all.jar AnnotateTargets -R my-reference.fasta --repeats my-repeated-regions.bed -T targets.tab -o annotated-targets.tab

The output file will contain additional columns, one for each annotation with the corresponding value per each target in the input.

Other changes:

* Fixed an issue with progressMeter not being updated in the FeatureWalker.
* Changed TargetColumn enum name to TargetTableColumn.
* Added a TargeTableWriter analogous to the TargetTableReader.
* Target now has a reference to an annotation collection.
* TargetTableReader automatically will read in annotation present in the input target table.


vruano:
@davidbenjamin please review.
@achevali I think you had an interest in GC bias annotation, correct? Here is the code.

src/main/java/org/broadinstitute/hellbender/tools/exome/AnnotateTargets.java
```diff
@@ -0,0 +1,252 @@
...
+ *     target table into the output target table.
+ * </p>
+ * <p>
+ *     There must be at least one annotation whose dependencies are satisfied, otherwise the
+ *     tool will return an error.
+ * </p>
+ * <p>
+ *     These are the currently supported annotations and their dependencies:
+ *     <dl>
+ *         <dt>{@link TargetAnnotation#GC_BIAS GC_BIAS}</dt>
```
davidbenjamin: This annotation should be called "GC_CONTENT" or "PERCENT_GC" etc, not  GC bias.  There's only bias if changes in content lead to changes in coverage.

src/main/java/org/broadinstitute/hellbender/tools/exome/AnnotateTargets.java
```diff
@@ -0,0 +1,252 @@
...
+ *         <dd><p>GC content in the target expressed as the fraction of concrete
```
davidbenjamin: Is "concrete base" a standard term?

src/main/java/org/broadinstitute/hellbender/tools/exome/AnnotateTargets.java
```diff
@@ -0,0 +1,252 @@
...
+ *         bases (A, C, G, T) that are G or C in the reference sequence that
+ *         overlap the target.</p>
+ *         <p> It can take values from 0 to 1
+ *         or NaN in the case special case that there is no concrete bases
```
davidbenjamin: "are no"

src/main/java/org/broadinstitute/hellbender/tools/exome/AnnotateTargets.java
```diff
@@ -0,0 +1,252 @@
...
+ * @author Valentin Ruano-Rubio &lt;valentin@broadinstitute.org&gt;
+ */
+@CommandLineProgramProperties(
+        oneLineSummary = "Annotate targets",
+        summary = "Annotate targets with various properties such as GC Bias and repeat fraction",
+        programGroup = ExomeAnalysisProgramGroup.class
+)
+public class AnnotateTargets extends TargetWalker {
+
+    public static final String REPEAT_REGION_FULL_NAME = "repeats";
```
davidbenjamin: regions should be plural

src/main/java/org/broadinstitute/hellbender/tools/exome/AnnotateTargets.java
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    private interface TargetAnnotator {
+        Object apply(final Target target, final ReadsContext readsContext, final ReferenceContext referenceContext, final FeatureContext featureContext);
+    }
+
+    /**
+     * Calculate the GC-Bias target annotation.
+     */
+    private class GCBiasAnnotator implements TargetAnnotator {
```
davidbenjamin: GCContent eg, not GCBias

src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
```diff
@@ -103,8 +104,7 @@ public void onTraversalStart() {
         readFilter = makeReadFilter();
 
         try  {
-            final Writer outputWriter = createOutputWriter(outputFile);
-            outputTableWriter = ReadCountCollectionUtils.writerWithIntervals(outputWriter, sampleList);
+            final Writer outputWriter = createOutputWriter(outputFile);outputTableWriter = ReadCountCollectionUtils.writerWithIntervals(outputWriter, sampleList);
```
davidbenjamin: two lines?

src/main/java/org/broadinstitute/hellbender/tools/exome/HashTargetAnnotationCollection.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.*;
+
+/**
+ * Simple target annotation collection based on a hash map from the annotation to its value as an string.
+ *
+ * @author Valentin Ruano-Rubio &lt;valentin@broadinstitute.org&gt;
+ */
+final class HashTargetAnnotationCollection implements TargetAnnotationCollection {
+
+    private final Map<TargetAnnotation, String> values;
```
davidbenjamin: Since `TargetAnnotation` is an `Enum` couldn't this be implemented as an `EnumMap`?

src/main/java/org/broadinstitute/hellbender/tools/exome/TargetAnnotation.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+public enum TargetAnnotation {
+
+    /**
+     * Fraction of bases that are G or C in the reference overlapping the target.
+     *
+     * <p>This value is always a number between 0 and 1 except in very rare cases with small targets with all, undefined bases (N or X),
+     * in this case it value will be {@link Double#NaN NaN}</p>.
+     */
+    GC_BIAS(TargetTableColumn.GC_BIAS),
```
davidbenjamin: CONTENT, not BIAS

src/main/java/org/broadinstitute/hellbender/tools/exome/TargetTableAnnotationManager.java
```diff
@@ -0,0 +1,101 @@
...
+ * <p>
+ *     Holds information on what annotations are present in the target table file, what columns contain them and how to
+ *     construct {@link TargetAnnotationCollection} instances to query these per target.
+ * </p>
+ *
+ * @author Valentin Ruano-Rubio &lt;valentin@broadinstitute.org&gt;
+ */
+final class TargetTableAnnotationManager {
+
+    private Set<TargetAnnotation> annotationSet;
```
davidbenjamin: What about `EnumSet`?

src/main/java/org/broadinstitute/hellbender/tools/exome/TargetTableAnnotationManager.java
```diff
@@ -0,0 +1,101 @@
...
+     * @param dataLine the target data-line.
+     * @return never {@code null}.
+     */
+    public TargetAnnotationCollection createTargetAnnotationCollection(final DataLine dataLine) {
+
+        if (annotationSet.size() == 0) {
+            return TargetAnnotationCollection.NO_ANNOTATION;
+        }
+
+        return new TargetAnnotationCollection() {
```
davidbenjamin: What's the purpose of returning an anonymous class instead of using your implementation of `TargetAnnotationCollection`?

src/main/java/org/broadinstitute/hellbender/utils/Nucleotide.java
```diff
@@ -0,0 +1,125 @@
+package org.broadinstitute.hellbender.utils;
+
+import java.util.Arrays;
+
+/**
+ * Represents the nucleotide alphabet.
+ *
+ * <p>
+ *    Thi enumerator not only contains concrete nucleotides, but also
```
davidbenjamin: This

src/main/java/org/broadinstitute/hellbender/utils/Nucleotide.java
```diff
@@ -0,0 +1,125 @@
...
+        baseToValue['t'] = baseToValue['T'] = T;
+        baseToValue['u'] = baseToValue['U'] = T;
+        baseToValue['x'] = baseToValue['X'] = X;
+        baseToValue['n'] = baseToValue['N'] = N;
+    }
+
+    /**
+     * Returns the base that corresponds to this nucleotide.
+     * <p>
+     *    The base is return in uppercase.
```
davidbenjamin: returned

src/main/java/org/broadinstitute/hellbender/utils/Nucleotide.java
```diff
@@ -0,0 +1,125 @@
...
+        public Counter() {
+            counts = new long[Nucleotide.values().length];
+        }
+
+        /**
+         * Increases by 1 the count for a nucleotide.
+         * @param nucleotide the target nucleotide.
+         * @throws IllegalArgumentException if nucleotide is {@code null}.
+         */
+        public void addAll(final Nucleotide nucleotide) {
```
davidbenjamin: Why is this method called `addAll`?

src/test/java/org/broadinstitute/hellbender/tools/exome/AnnotateTargetsIntegrationTest.java
```diff
@@ -0,0 +1,286 @@
...
+        for (int i = interval.getStart() - 1; i < interval.getEnd(); i++) {
+            if (REPEATED_ARRAY[sequenceIndex][i]) {
+                repeatedBases++;
+            }
+        }
+        final double actual = repeatedBases / (double) (interval.getEnd() - interval.getStart() + 1);
+        Assert.assertEquals(observed, actual, 0.0001);
+    }
+
+    private void checkOutputGCBias(final ReferenceFileSource reference, final SimpleInterval interval, final double observed) {
```
davidbenjamin: content, not bias

src/test/java/org/broadinstitute/hellbender/tools/exome/AnnotateTargetsIntegrationTest.java
```diff
@@ -0,0 +1,286 @@
...
+                case 'c':
+                    gc++; total++; break;
+                case 'a':
+                case 't':
+                case 'u':
+                    total++; break;
+                default:
+            }
+        }
+        final double expectedGCbias = total == 0 ? Double.NaN : ((double) gc / (double) total);
```
davidbenjamin: ditto

davidbenjamin:
Back to @vruano

src/main/java/org/broadinstitute/hellbender/tools/exome/AnnotateTargets.java
vruano: changing to "there is no unambiguous base"

src/main/java/org/broadinstitute/hellbender/tools/exome/HashTargetAnnotationCollection.java
vruano: Good one! didn't know about their existence.

src/main/java/org/broadinstitute/hellbender/tools/exome/TargetTableAnnotationManager.java
vruano: Same.

vruano: However here is still better to just say "Set" as there is no need to indicate the actual implenetation used in the member declaration.

vruano: This also applies to the "EnumMap" vs "Map" above.

src/main/java/org/broadinstitute/hellbender/tools/exome/TargetTableAnnotationManager.java
vruano: Is best to use a inner class, regardless whether anonymous or named, because the implementantions will all share the set of annotations available and their index in each data-line, which is the same for every data-line. 

Use a general top-level implementation would result in duplicated storage even when these annotation may never be accessed in some cases. 

One could declare the inner class explicitly rather than use an annonymous one. I can do that if you like it better.

vruano: Will do that.

src/main/java/org/broadinstitute/hellbender/utils/Nucleotide.java
vruano: No good reason, changing it to ```add```.

vruano:
back to @davidbenjamin

vruano:
The last commit should do the trick. Will rebase and merge once tests pass.

vruano:
Seems that travis is acting up... test work locally and gsa5. So they are mergeable.

PR #192, opened 2015-11-19
adding gradle jvm memory limit

lbergelson:
adding a 1gb limit to the gradle jvm to try and stop the intermittent out of memory errors

[Review on Reviewable](https://reviewable.io/reviews/broadinstitute/hellbender-protected/192)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: java
+sudo: required
+dist: trusty
 jdk:
 - oraclejdk8
 env:
@@ -6,6 +8,12 @@ env:
   - TERM=dumb
   global:
   - JAVA_LIBRARY_PATH=/usr/lib/jni
+
+  #needed to avoid a bug in the hdf5 installer
+  - PS1=">"
+
+  #limit gradle jvm memory usage
+  - GRADLE_OPTS=-Xmx512m
 
   # If this changes, you will need to update the gradle build
   - HDF5_DIR=/hdf/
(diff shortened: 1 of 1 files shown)


lbergelson:
@LeeTL1220 This looks like it may actually fix the problem now.  I discovered that the beta infrastructure running ubuntu 14 instead of 12 offers more ram, 7gb, and switched the build to run on that.  It seems to pass.

lbergelson:
I'm getting a test failure:

```java
Gradle suite > Gradle test > org.broadinstitute.hellbender.tools.exome.allelefraction.AlleleFractionInitializerUnitTest.testInitialize FAILED
    java.lang.AssertionError: expected [0.0] but found [-0.02041761966572464]
```

I don't think it's the fault of my changes though.

lbergelson:
hmn, I get a different teset failure on rerurn:
```
 org.broadinstitute.hellbender.tools.exome.CalculateCoverageComponentsIntegrationTest.testSimpleMatrix FAILED
    java.lang.AssertionError: expected [4] but found [3]
        at org.testng.Assert.assertEquals(Assert.java:125)
        at org.testng.Assert.assertEquals(Assert.java:372)
        at org.testng.Assert.assertEquals(Assert.java:382)
        at org.broadinstitute.hellbender.utils.pca.PCAUnitTest.assertEquals(PCAUnitTest.java:201)
        at org.broadinstitute.hellbender.tools.exome.CalculateCoverageComponentsIntegrationTest.testSimpleMatrix(CalculateCoverageComponentsIntegrationTest.java:54)
```

vruano:
Yup I have reported that problem in #239 and @davidbenjamin is looking into it.

vruano:
That is for the one in AlleleFractionBlah... the second one I never seen it before, I guess that is for me to take care of... although test have always worked locally.

lbergelson:
@vruano Ah, thanks for pointing that out.  

This is what happens when we can't run our test infrastructure :(  Rerunning again.  I think we should merge this even if it doesn't pass, since its now at least running the tests.

vruano:
Ok.

PR #200, opened 2015-11-30
Modified CalcualteTargetCovarage:

vruano:
* Now CalculatTargetCoverage can calculate coverage using two different units:
    - OVERALAPPING_READs : the clasic one
    - AVERAGE_DEPTH : number of overalapping base-calls / target-width.

    The latter is closer to what is used by XHMM (still need to avoid duplicated counting of overlapping pair-read bases).

    Also have added a minBQ filter to discard poor quality base-calls from the count.

* Added a minBQ filter to discard poor quality base-calls from being counted.


vruano:
@LeeTL1220 please review.

src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
```diff
@@ -176,32 +344,163 @@ private static Writer createOutputWriter(final File outputFile) {
...
+                    .forEach(read -> {
+                        final int columnIndex = readToColumn.applyAsInt(read);
+                        if (columnIndex >= 0) {
+                            counts[columnIndex] += countOverlappingQualifyingBaseCalls(read, interval);
+                        }
+                    });
+            return counts;
+        }
+
+        private long countOverlappingQualifyingBaseCalls(final GATKRead read, final SimpleInterval interval) {
```
LeeTL1220: This is a bit of a long function, please add a simple comment.

src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
```diff
@@ -52,10 +72,43 @@
...
+ * <h4>Qualifying base-calls</h4>
+ * <p>When the user selects {@link CoverageUnit#AVERAGE_DEPTH} as the
+ * coverage unit, only qualifying base-calls will be taken into account.</p>
+ * A base is qualifying iff:
+ * <ul>
+ *     <li>is enclosed in a qualifying read,</li>
+ *     <li>aligned with a base in the reference (i.e. not part for an insertion nor a clip) and</li>
+ *     <li>its not a poor quality call.</li>
+ * </ul>
+ * <p>The user can indicate the minimum base call quality to qualify
```
LeeTL1220: Why not use the filter infrastructure we already have?

src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
```diff
@@ -156,6 +262,62 @@ public void onTraversalStart() {
...
+     *
+     * <p>
+     *     In addition it warns the user if he sets a value for the {@link #minimumBaseQuality}
+     *     argument yet base-call qualities are not relevant given the {@link #coverageUnit}
+     *     selected.
+     * </p>
+     *
+     * @throws UserException.BadArgumentValue if any of the argument has a bad value.
+     */
+    private void checkThresholdArgumentValues() {
```
LeeTL1220: Again, we have a read filtering infrastructure that we can use.

src/test/resources/org/broadinstitute/hellbender/tools/exome/exome-read-counts-base-calls.tsv
```diff
@@ -0,0 +1,6 @@
+CONTIG	START	END	NAME	NA12778	NA12872	NA12878
```
LeeTL1220: Can we get rid of the trailing ``.0`` for these integer values?

LeeTL1220:
@vruano Minor comments.

vruano:
@LeeTL1220 please review, sorry for the long delay. Even if this has overlap with the existing read-counting tool in gatk-public I would like to get it merged and get the XHMM base-call counts implemented asap. We can merge these too tools eventually.

LeeTL1220:
@vruano I would like to merge as well, but these tools should not have the same name (for now).  Won't this cause problems when invoking the tool that used to be ExomeReadCounts?  Can you rename the gatk-protected ``CalculateTargetCoverage``?

vruano:
Ah yup sorry. Will change it to something else.... say CalculateTargetBaseCallCoverage making base-call counting the default. @LeeTL1220  Do you agree?

LeeTL1220:
@vruano Yes, I do.

Also, I really appreciate the copious comments.  I do not have any other comments.  Back to you.

Issue #165, opened 2015-10-27
Standard argument definitions


davidbenjamin:
hellbender has a class called StandardArgumentDefinitions for common command line arguments.  We should either make our own (for segments, targets, etc) or add to the hellbender class.

davidbenjamin:
Fixed in PR #202.

PR #202, opened 2015-12-01
Consolidating standard command line arguments. Addresses issue #165.

davidbenjamin:
@achevali Could you review this?


src/main/java/org/broadinstitute/hellbender/engine/TargetWalker.java
```diff
@@ -24,11 +31,5 @@ protected boolean isAcceptableFeatureType(final Class<? extends Feature> feature
...
 
     @Override
-    public File getDrivingFeatureFile() {
-        final File targetsFile = targetArguments.getTargetsFile();
-        if (targetsFile == null) {
-            throw new UserException.BadArgumentValue(TargetArgumentCollection.TARGET_FILE_FULL_NAME, "you must specify a target file");
-        }
-        return targetsFile;
-    }
+    public File getDrivingFeatureFile() { return targetsFile; }
```
achevali: Why is this check no longer necessary?

achevali:
Looks good, you can merge when convinced tests are passing.

src/main/java/org/broadinstitute/hellbender/engine/TargetWalker.java
davidbenjamin: This walker used to have a `TargetArgumentCollection`, which has an optional targets file along with other arguments.  Since 1) these other arguments were not used and 2) the targets file is never optional for this walker, I decided to use the straightforward approach of simply having a non-optional command line argument for the targets file.

Issue #51, opened 2015-07-10
Test files in the wrong subdirectories.


vruano:
In Hellbender protected we have not being go when it comes to placing the test resources in the appropriate subdirectory. 

This should follow the package name structure of the class or classes whose test the file was added for. 

E.g. NormalizeSomaticsReadCounts testing files should be under src/test/resouces/org/broadinginsitute/hellbender/tools/exome but the "hellbender" section was omitted. 

Another e.g. the Segmenter testing files sit directly under src/test/segmenter with ignore the package name. 

So this task is about moving the files to the appropriate location and update the testing code to refer to the new location.

vruano:
Addressed in recent pull request.

PR #203, opened 2015-12-01
moved some test resources.  Addresses issue #51.

davidbenjamin:
@vruano Is this change what you had in mind?


vruano:
Back to @davidbenjamin

vruano:
I think you might be forgetting *.utils.* to -> *.hellbender.utils.* and also the segmenter and other test files that are placed directly under src/test/resources. 

You don't need to address them all here if you don't feel like investing time on it right now... In that case update the issue indicating what is pending.

davidbenjamin:
@vruano I moved the other resources, too.

vruano:
Looks good., I tried in gsa5 and test pass so I will merge it.

Issue #237, opened 2015-12-09
Compilation warnings coming from AlleleFractionState.


vruano:
Not a tragedy but annoying:
<pre>
/Users/valentin/Developer/hellbender-protected/src/main/java/org/broadinstitute/hellbender/tools/exome/allelefraction/AlleleFractionState.java:25: warning: [rawtypes] found raw type: List
        public MinorFractions(final List other) {
                                    ^
  missing type arguments for generic class List<E>
  where E is a type-variable:
    E extends Object declared in interface List
/Users/valentin/Developer/hellbender-protected/src/main/java/org/broadinstitute/hellbender/tools/exome/allelefraction/AlleleFractionState.java:26: warning: [unchecked] unchecked method invocation: constructor <init> in class ArrayList is applied to given types
            super(new ArrayList<>(other));
                  ^
  required: Collection<? extends E>
  found: List
  where E is a type-variable:
    E extends Object declared in class ArrayList
/Users/valentin/Developer/hellbender-protected/src/main/java/org/broadinstitute/hellbender/tools/exome/allelefraction/AlleleFractionState.java:26: warning: [unchecked] unchecked conversion
            super(new ArrayList<>(other));
                                  ^
  required: Collection<? extends E>
  found:    List
  where E is a type-variable:
    E extends Object declared in class ArrayList
/Users/valentin/Developer/hellbender-protected/src/main/java/org/broadinstitute/hellbender/tools/exome/allelefraction/AlleleFractionState.java:26: warning: [unchecked] unchecked method invocation: constructor <init> in class ArrayList is applied to given types
            super(new ArrayList<>(other));
                 ^
  required: Collection<? extends E>
  found: ArrayList
  where E is a type-variable:
    E extends Object declared in class ArrayList
/Users/valentin/Developer/hellbender-protected/src/main/java/org/broadinstitute/hellbender/tools/exome/allelefraction/AlleleFractionState.java:26: warning: [unchecked] unchecked conversion
            super(new ArrayList<>(other));
</pre>

vruano:
```git blame```s you @davidbenjamin

akiezun:
i think you guys should run the compiler with `'-Werror'` and eliminate all warnings

vruano:
Does is what is done in gatk (public)?

akiezun:
yes

vruano:
Sounds good, will take care.

vruano:
Why you do '-proc:none'?

[...]

PR #238, opened 2015-12-09
Fixes compilation warnings and make all warnings to be considered error…

vruano:
…s from this point on

Resolves issue #237


vruano:
Please take a look @davidbenjamin.

Issue #239, opened 2015-12-09
Tests in travis fail for AlleleFractionInitializerUnitTest.


vruano:
Locally testing does ok but when run in Travis you get an error consistently.

The stack trace reads:

<pre>
org.broadinstitute.hellbender.tools.exome.allelefraction.AlleleFractionInitializerUnitTest.testInitialize FAILED
    java.lang.AssertionError: expected [0.0] but found [-0.023368743794425884]
        at org.testng.Assert.fail(Assert.java:94)
        at org.testng.Assert.failNotEquals(Assert.java:496)
        at org.testng.Assert.assertEquals(Assert.java:209)
        at org.testng.Assert.assertEquals(Assert.java:222)
        at org.broadinstitute.hellbender.tools.exome.allelefraction.AlleleFractionInitializerUnitTest.testInitialize(AlleleFractionInitializerUnitTest.java:41)
</pre>

vruano:
@davidbenjamin perhaps a hostname/architecture random seed? Native libraries involved? Non determinism caused by multi-core Spark?

davidbenjamin:
@vruano I'll get to the bottom of this.

davidbenjamin:
@vruano I forgot to seed a few random generators in one class.  This is fixed in PR #241, which is currently under review.

davidbenjamin:
I snuck it into PR #202 actually in order to get it to pass.

samuelklee:
Got another one in travis build for #351:

    Gradle suite > Gradle test > org.broadinstitute.hellbender.tools.exome.allelefraction.AlleleFractionInitializerUnitTest.testInitialize FAILED
    java.lang.AssertionError: expected [0.0] but found [0.02006594020751029]
        at org.testng.Assert.fail(Assert.java:94)
        at org.testng.Assert.failNotEquals(Assert.java:496)
        at org.testng.Assert.assertEquals(Assert.java:209)
        at org.testng.Assert.assertEquals(Assert.java:222)
        at org.broadinstitute.hellbender.tools.exome.allelefraction.AlleleFractionInitializerUnitTest.testInitialize(AlleleFractionInitializerUnitTest.java:34)

samuelklee:
I think this was closed in #427.

PR #241, opened 2015-12-10
Faster, simpler, more accurate initialization of allele fraction model

davidbenjamin:
@samuelklee inspired by your suspicion about the poor tolerances for bias variance.

@LeeTL1220 @vruano If speed on wgs is ever an issue, this code could easily be extended to get fast and accurate Bayesian estimates without MCMC via the Laplace approximation.  The initialization gets the mode, and the posterior error bars are found in the initial step size guesses for the samplers.


samuelklee:
just a reminder here that i was still getting intermittent failures on travis in #254.  this already incorporated #202 (which tried to address #239).  so there may still be a bug somewhere...probably best to fix it in this PR if it crops up again.

    org.broadinstitute.hellbender.tools.exome.allelefraction.AlleleFractionInitializerUnitTest.testInitialize FAILED
    java.lang.AssertionError: expected [0.0] but found [-0.02183397937695908]

src/main/java/org/broadinstitute/hellbender/tools/exome/allelefraction/AlleleFractionInitializer.java
```diff
@@ -19,105 +19,48 @@
...
-        responsibilities = data.getAllelicCounts().stream().map(c -> new Responsibility(c)).collect(Collectors.toList());
-        state = new AlleleFractionState(BIAS_MEAN_INITIAL, BIAS_VARIANCE_INITIAL, INITIAL_OUTLIER_PROBABILITY, estimateMinorFractions());
-        for (int iteration = 0; iteration < NUMBER_OF_EM_ITERATIONS; iteration++) {
-            performSingleEMIteration();
-        }
+        state = new AlleleFractionState(INITIAL_MEAN_BIAS, INITIAL_BIAS_VARIANCE, INITIAL_OUTLIER_PROBABILITY, initialMinorFractions(data));
+        double previousIterationLogLikelihood;
+        double nextIterationLogLikelihood = Double.NEGATIVE_INFINITY;
+        int iteration = 1;
+        System.out.format("Initializing allele fraction model.  Iterating until log likelihood converges to within %.3f.%n",
```
samuelklee: use logger instead of system out

src/main/java/org/broadinstitute/hellbender/tools/exome/allelefraction/AlleleFractionInitializer.java
```diff
@@ -19,105 +19,48 @@
...
+                LOG_LIKELIHOOD_CONVERGENCE_THRESHOLD);
+        do {
+            previousIterationLogLikelihood = nextIterationLogLikelihood;
+            state = new AlleleFractionState(estimateMeanBias(data), estimateBiasVariance(data),
+                    estimateOutlierProbability(data), estimateMinorFractions(data));
+            nextIterationLogLikelihood = AlleleFractionModeller.logLikelihood(state, data);
+            System.out.format("Iteration %d, model log likelihood = %.3f.%n", iteration, nextIterationLogLikelihood);
```
samuelklee: same here

src/main/java/org/broadinstitute/hellbender/tools/exome/allelefraction/AlleleFractionModeller.java
```diff
@@ -103,25 +102,29 @@ public Double sample(final RandomGenerator rng, final AlleleFractionState state,
...
         private final int segmentIndex;
 
-        public PerSegmentMinorFractionSampler(final int segmentIndex, final double initialMinorFraction) {
-            sampler = new AdaptiveMetropolisSampler(initialMinorFraction, INITIAL_MINOR_FRACTION_STEP_SIZE, 0, MAX_MINOR_FRACTION);
+        public PerSegmentMinorFractionSampler(final int segmentIndex, final double initialMinorFraction, final double initialStepSize) {
+            sampler = new AdaptiveMetropolisSampler(initialMinorFraction, initialStepSize, 0, MAX_MINOR_FRACTION);
             this.segmentIndex = segmentIndex;
         }
 
         public Double sample(final RandomGenerator rng, final AlleleFractionState state, final AlleleFractionData data) {
```
samuelklee: note when rebasing that i added the following here to handle missing data:

    if (data.numHetsInSegment(segmentIndex) == 0) {
        return Double.NaN;
    }

src/main/java/org/broadinstitute/hellbender/tools/exome/allelefraction/AlleleFractionInitializer.java
```diff
@@ -19,105 +19,48 @@
...
+            iteration++;
+        } while (nextIterationLogLikelihood - previousIterationLogLikelihood > LOG_LIKELIHOOD_CONVERGENCE_THRESHOLD);
```
samuelklee: perhaps set a maximum number of iterations, just to be safe.

src/main/java/org/broadinstitute/hellbender/tools/exome/allelefraction/AlleleFractionInitializer.java
```diff
@@ -126,84 +69,79 @@ public AlleleFractionInitializer(final AlleleFractionData data) {
...
-    // reads in minor allele divided by total reads, ignoring outliers
-    private AlleleFractionState.MinorFractions estimateMinorFractions() {
+    /**
+     *  initialize minor fractions assuming no allelic bias <p></p>
+     *
+     *  we integrate over f to get posterior probabilities (responsibilities) of alt / ref minor
+     * that is, responsibility of alt minor is int_{0 to 1/2} f^a (1-f)^r df
+     *          responsibility of ref minor is int_{0 to 1/2} f^r (1-f)^a df
+     * these are proportional to I(1/2, a + 1, r + 1) and I(1/2, r + 1, a + 1),
+     * respectively, where I is the (incomplete) regularized Beta function
```
samuelklee: add a period to end the sentence (runs together when viewed as javadoc)

src/main/java/org/broadinstitute/hellbender/tools/exome/allelefraction/AlleleFractionInitializer.java
```diff
@@ -126,84 +69,79 @@ public AlleleFractionInitializer(final AlleleFractionData data) {
...
         return result;
     }
 
-    // M step for outlierProbability -- set it to the average of all OUTLIER responsibilities
-    private double estimateOutlierProbability() {
-        return responsibilities.stream().mapToDouble(r -> r.get(AlleleFractionIndicator.OUTLIER)).average().getAsDouble();
+    private double estimateOutlierProbability(final AlleleFractionData data) {
+        final UnivariateObjectiveFunction objective = new UnivariateObjectiveFunction(outlierProbability ->
+                AlleleFractionModeller.logLikelihood(state.shallowCopyWithProposedOutlierProbability(outlierProbability), data));
+        final SearchInterval searchInterval = new SearchInterval(0.0, 0.1, state.outlierProbability());
```
samuelklee: extract the min and max of the search interval

src/main/java/org/broadinstitute/hellbender/tools/exome/allelefraction/AlleleFractionInitializer.java
```diff
@@ -126,84 +69,79 @@ public AlleleFractionInitializer(final AlleleFractionData data) {
...
+        return new BrentOptimizer(0.01, 0.001).optimize(objective, GoalType.MAXIMIZE, searchInterval, new MaxEval(1000)).getPoint();
```
samuelklee: might want to extract rel and abs, too, but i'll leave it up to you

src/main/java/org/broadinstitute/hellbender/tools/exome/allelefraction/AlleleFractionInitializer.java
```diff
@@ -126,84 +69,79 @@ public AlleleFractionInitializer(final AlleleFractionData data) {
...
+        return new BrentOptimizer(0.001, 0.001).optimize(objective, GoalType.MAXIMIZE, searchInterval, new MaxEval(1000)).getPoint();
+    }
 
-        //update gamma hyperparameters
-        final double meanBias = estimateAverageBias();
-        final double meanSquaredBias = estimateAverageSquaredBias();
-        final double variance = meanSquaredBias - meanBias*meanBias;
-        state = new AlleleFractionState(meanBias, variance, state.outlierProbability(), state.minorFractions());
+    private AlleleFractionState.MinorFractions estimateMinorFractions(final AlleleFractionData data) {
+        AlleleFractionState.MinorFractions result = new AlleleFractionState.MinorFractions();
```
samuelklee: final (or this can all be made a one liner with streams, if you like)

src/main/java/org/broadinstitute/hellbender/tools/exome/allelefraction/AlleleFractionModeller.java
```diff
@@ -25,25 +26,27 @@
...
+ *</p>
+ * <p>
  * Each segment has a minor allele fraction f, and for each het within the locus
  * the number of alt reads is drawn from a binomial distribution with total count
  * n = #alt reads + #ref reads and alt probability f/(f + (1-f)*bias ratio) if the
  * locus is alt minor and (1-f)/(1-f + f*bias ratio) if the locus is ref minor.
- *
+ *</p>
+ * <p>
  * Conceptually, the model contains latent variable corresponding to the bias ratio
```
samuelklee: variable -> variables

src/main/java/org/broadinstitute/hellbender/tools/exome/allelefraction/AlleleFractionModeller.java
```diff
@@ -25,25 +26,27 @@
...
  * and an indicator for alt minor/ref minor at each het locus.  However, we integrate them
```
samuelklee: an indicator -> indicators

src/test/java/org/broadinstitute/hellbender/tools/exome/allelefraction/AlleleFractionInitializerUnitTest.java
```diff
@@ -42,39 +37,4 @@ public void testInitialize() {
...
-        Assert.assertEquals(resp1.get(AlleleFractionIndicator.ALT_MINOR) + resp1.get(AlleleFractionIndicator.REF_MINOR) + resp1.get(AlleleFractionIndicator.OUTLIER), 1.0);
-
-        resp5.setFromLogPosteriors(Double.NEGATIVE_INFINITY, 1, Double.NEGATIVE_INFINITY);
-        Assert.assertEquals(resp5.get(AlleleFractionIndicator.REF_MINOR), 1.0);
-        Assert.assertEquals(resp1.get(AlleleFractionIndicator.ALT_MINOR) + resp1.get(AlleleFractionIndicator.REF_MINOR) + resp1.get(AlleleFractionIndicator.OUTLIER), 1.0);
-
-        resp5.setFromLogPosteriors(Math.log(1.0), Math.log(3.0), Double.NEGATIVE_INFINITY);
-        Assert.assertEquals(resp5.get(AlleleFractionIndicator.ALT_MINOR), 1.0/4.0);
-        Assert.assertEquals(resp1.get(AlleleFractionIndicator.ALT_MINOR) + resp1.get(AlleleFractionIndicator.REF_MINOR) + resp1.get(AlleleFractionIndicator.OUTLIER), 1.0);
-    }
```
samuelklee: will need to optimize imports here after rebasing

samuelklee:
looks good, back to you @davidbenjamin!  we can probably extract some of the Laplace approximation/likelihood maximization stuff later, if needed for WGS (or we could even use it for quickly refitting the model between iterations of similar-segment merging, if we decide to change up that step)

davidbenjamin:
Responded to all comments.  Note the change of randoms seeds in `AlleleFractionSimulatedData`.  Several random samplers (two Poisson, a uniform, and a gamma) were sharing a single random generator object.  I don't immediately see how that would cause non-deterministic tests, but it seemed fishy and I fixed it.  However, I'm not sure that this was te culprit.

davidbenjamin:
Back to @samuelklee, but could a Java connoisseur like @vruano comment on whether sharing of random generators (see above comment) might have been the problem?

vruano:
IMO there is nothing intrinsically fishy about using a single random number generator for everything; it will depend on the situation what approach is more appropriate. Lack of determinism comes from different execution order or parallel execution... is there a chance the any of these two is happening? 

Even if the test run in a single thread, I wonder whether the order the unit or integration test are run each time might be different; Is it possible that 'gradle test' / TestNG runs test in different order each time or even it may run them in parallel?

samuelklee:
OK by me to rebase and merge.

Issue #244, opened 2015-12-11
Fix segment-union step.


samuelklee:
Currently strands hets/targets due to fencepost errors, leading to oversegmentation downstream.  Parsimonious interval should also be refactored to be based on target ends.

samuelklee:
Some working notes.  @davidbenjamin and @LeeTL1220 might want to comment.

1) Fencepost errors were occurring when a new segment was being created from e.g. a target-segment start breakpoint and a SNP-segment start breakpoint.  The single SNP at the start of the SNP segment was being included in the new segment, leading to difficulties in similar-segment merging downstream.  I think that this problem is solved by keeping track of the type of each breakpoint (i.e., start or end), and decreasing the end of the new segment by one if the corresponding breakpoint was originally a start.

2) Naive unioning of breakpoints is introducing new segments that contain only targets (e.g., when two SNP segments occur within a single target segment, 3 such segments at the start, middle, and end are created).  If the target coverages in these new segments are all high or low by chance, these spurious segments may not get remerged during similar-segment merging, even if the merge threshold is quite aggressive (e.g., 5 sigma).

I think an easy solution is to check all new segments containing only targets against the original CBS target segments, and discard those that were not preexisting.  This can be done in small-segment merging.

3) Making the unioned segments parsimonious (i.e., making sure that their endpoints corresponding to target starts/ends or SNP sites) is also problematic.  For example, if a SNP-segment breakpoint breaks up a target segment into two new segments but also happens to fall *within* a target, which of the two segments should that target belong to?  Right now, I believe that the end of the first new segment is extended to the end of that target, but this is problematic when the next breakpoint also falls within that target.

Note that the output of GATK CNV defines target segments by the target ends, so perhaps we should follow this convention here as well.

samuelklee:
Another thought: rather than just merging the "middle" segments described above, model comparison over all possible breakpoints contained within them could be performed.

davidbenjamin:
Fixed in PR #254.

Issue #116, opened 2015-09-25
Implement AlleleicCapsegger workflow class.


samuelklee:
We are still porting over the back half of python ACS, but I'll start putting together the workflow up until small-segment merging so we can make sure things are working correctly so far.

samuelklee:
Closed in #254.

Issue #64, opened 2015-07-22
Revise ACS likelihood model and segment-merging algorithm once ported.


samuelklee:
Right now, there are some questionable steps in the segment-merging algorithm in python ACS.  For example, small segments containing less than 3 targets are forcibly merged with adjacent segments, where the choice of which adjacent segment (left or right) to merge with is determined by an ad hoc allelic fraction-copy ratio Manhattan distance.

For now, we'll port over the python algorithm faithfully, but afterwards we should re-evaluate and revise  if necessary.

samuelklee:
We're sticking with a variation on the original similar-segment merging procedure for now.  The basic idea is to use the posterior summary statistics (i.e., mean and standard deviation) for the segment mean and minor allele fraction to decide whether to merge adjacent segments.

We merge if the absolute difference in the parameter mean of the segments is less than a constant times the parameter standard deviation in each segment.  Merging is done starting from left to the right on the genome, and a segment is merged repeatedly until it is no longer similar with the segment on the right.

Summary statistics for the merged segments are naively updated using inverse-weighted variance during each round of merging over the segments.  After each round is complete, the model is completely refit using MCMC.  Additional rounds of merging continue until convergence (i.e., no more merges are found).

Using the above procedure, the copy-ratio--only ACS takes ~20 seconds to run on HCC1143T from start to finish (not including het pulldown, which on its own takes ~15 minutes for the non-Spark version), starting with ~15k tumor hets, ~125k targets, and 384 segments (from an old version of recapseg).  There are 959 segments after segment union, 538 after small-segment merging, and similar-segment merging with the constant set to 2 finds 68 similar segments and converges after 3 rounds of MCMC with 25 burn-in and 50 sample iterations.

Note that this procedure is nevertheless ad hoc---a more appropriate thing to do would be model comparison on each pair of adjacent segments (w/ and w/out the breakpoint), but this would require several likelihood evaluations and would probably yield a similar result.

samuelklee:
Closed in #254.

PR #254, opened 2015-12-17
Added allelic CNV workflow.

### LINKED ISSUE #49: Port SegmentMerger from python AllelicCapseg
starting with merge_small_segments

samuelklee:
Also:
* Added similar-segment merging based on posterior summaries.
* Reworked segment-union step and added spurious-segment merging. 
* Refactored reading/writing of segment files.
* Changed AllelicCount to use SimpleInterval and added column arrays to SegmentTableColumns.
* General refactoring and code cleanup. 

Closes #49, #64, #116, #244.


samuelklee:
Sorry for the length, @davidbenjamin!  Although there are many changed files, the changes in a large fraction of them are minor refactors that should be fairly straightforward to review.  The majority of the code in the new classes, (e.g., AllelicCapSegger, ACSModeller, etc.) is also fairly straightforward.  Probably the only nontrivial code is in SegmentUtils (in which I made a lot of changes, which might be difficult to parse from the diff---probably easier to look at the code in IntelliJ to review it) and SegmentMergeUtils, although that in the latter is not too complicated either.

samuelklee:
OK, think i fixed the AlleleFractionSimulatedData-related failures.  @vruano, note that I did see one intermittent PCAUnitTest failure that went away after restarting the build.

src/main/java/org/broadinstitute/hellbender/tools/exome/ACSCopyRatioModeller.java
```diff
@@ -129,21 +135,25 @@ public ACSCopyRatioModeller(final SegmentedModel segmentedModel) {
...
-                    double ll = 0.;
-                    for (int target = 0; target < numTargetsInSegment; target++) {
-                        final double coverage = dataCollection.getCoveragesInSegment(segment).get(target);
-                        if (!state.getOutlierIndicator(targetStartInSegment + target)) {
-                            ll -= normalTerm(coverage, newMean, state.variance());
+                if (numTargetsInSegment == 0) {
+                    means.add(Double.NaN);
+                } else {
+                    final Function<Double, Double> logConditionalPDF = newMean -> {
+                        double ll = 0.;
```
davidbenjamin: Perhaps I have drunk the Java 8 Kool-Aid, but I have started thinking of `double/int <variable> = 0` followed by a `for` loop involving `<variable>` as a possible code smell.  IMO the following is more readable:
```
final Function<Double, Double> logConditionalPDF = newMean -> IntStream.range(0, numTargetsInSegment)
			.filter(target -> !state.getOutlierIndicator(targetStartInSegment + target))
			.map(target -> dataCollection.getCoveragesInSegment(segment).get(target))
			.map(coverage -> -normalTerm(coverage, newMean, state.variance()))
			.sum();
```

samuelklee:
@davidbenjamin Spoke too soon, looks like I just got another failure on travis-ci/push, although travis-ci/pr passed:
    
    org.broadinstitute.hellbender.tools.exome.allelefraction.AlleleFractionInitializerUnitTest.testInitialize FAILED
    java.lang.AssertionError: expected [0.0] but found [-0.02183397937695908]

src/main/java/org/broadinstitute/hellbender/tools/exome/ACSCopyRatioModeller.java
```diff
@@ -129,21 +135,25 @@ public ACSCopyRatioModeller(final SegmentedModel segmentedModel) {
...
-                };
+                        return ll;
+                    };
 
-                final SliceSampler sampler =
-                        new SliceSampler(rng, logConditionalPDF, state.getMeanInSegment(segment),
-                                meanSliceSamplingWidth);
-                means.add(sampler.sample());
+                    final SliceSampler sampler =
+                            new SliceSampler(rng, logConditionalPDF, state.getMeanInSegment(segment),
```
davidbenjamin: This doesn't fit on one line, but I think it would look better on two than on three.

src/main/java/org/broadinstitute/hellbender/tools/exome/ACSModeller.java
```diff
@@ -0,0 +1,150 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Represents an ACS segmented model for copy ratio and allele fraction.
+ *
+ * @author Samuel Lee &lt;slee@broadinstitute.org&gt;
+ */
+public final class ACSModeller {
+    protected static final String INITIAL_SEG_FILE_TAG = "sim-0";
```
davidbenjamin: If you could make these tags follow alphabetical order the results files would look nicer in a terminal.

src/main/java/org/broadinstitute/hellbender/tools/exome/ACSModeller.java
```diff
@@ -0,0 +1,150 @@
...
+        List<ACSModeledSegment> mergedSegments = new ArrayList<>(segments);
+
+        //write initial model fit to file
+        final File initialModeledSegmentsFile = new File(outputPrefix + "-" + INITIAL_SEG_FILE_TAG + ".seg");
+        SegmentUtils.writeACSModeledSegmentFile(initialModeledSegmentsFile, mergedSegments, segmentedModel.getGenome());
+
+        //perform iterations of similar-segment merging until all similar segments are merged
+        int numIterations = 1;
+        int prevNumSegments;
+        while (numIterations <= MAX_SIMILAR_SEGMENT_MERGE_ITERATIONS) {
```
davidbenjamin: This should be a `for` loop.

src/main/java/org/broadinstitute/hellbender/tools/exome/ACSModeller.java
samuelklee: good point, i thought about this issue but didn't come up with a good scheme.  the natural order is:

1) seg file from RCS
2) MAF SNP segments (-MAF, currently)
3) segment union (-union)
4) after small-segment merging (-no-small)
5) initial fit (-sim-0)
6) fits after each iteration of similar-segment merging (-sim-#)
7) final fit (-sim-final)

src/main/java/org/broadinstitute/hellbender/tools/exome/AllelicCapSegger.java
```diff
@@ -0,0 +1,207 @@
...
+        segmentedModel.writeSegmentFileWithNumTargetsAndNumSNPs(segmentedModelFile);
+
+        //initial model fitting
+        logger.info("Fitting initial model...");
+        final ACSModeller modeller = new ACSModeller(segmentedModel, outputPrefix,
+                numSamplesCopyRatio, numBurnInCopyRatio, numSamplesAlleleFraction, numBurnInAlleleFraction);
+
+        //similar-segment merging
+        logger.info("Merging similar segments...");
+        modeller.mergeSimilarSegments(sigmaThresholdCopyRatio, sigmaThresholdAlleleFraction,
```
davidbenjamin: It deserves a comment that output of segment files (including the final one) is a side effect of `mergeSimilaregments` since output is explicit up to this point.

src/main/java/org/broadinstitute/hellbender/tools/exome/HetPulldownCalculator.java
```diff
@@ -156,18 +160,18 @@ public Pulldown getTumor(final File tumorBAMFile, final IntervalList normalHetIn
...
-             final ReferenceSequenceFileWalker refWalker = new ReferenceSequenceFileWalker(this.refFile)
-        ) {
+             final ReferenceSequenceFileWalker refWalker = new ReferenceSequenceFileWalker(this.refFile)) {
             if (bamReader.getFileHeader().getSortOrder() != SAMFileHeader.SortOrder.coordinate) {
                 throw new UserException.BadInput("BAM file " + bamFile.toString() + " must be coordinate sorted.");
             }
 
             final Pulldown hetPulldown = new Pulldown(bamReader.getFileHeader());
 
+            final int snpTotalNumber = snpIntervals.size();
```
davidbenjamin: As long as you're making this more human-readable, may as well name the variable `totalNumberOfSnps`.

src/main/java/org/broadinstitute/hellbender/tools/exome/ModeledSegment.java
```diff
@@ -7,7 +7,8 @@
 
 import java.util.Arrays;
 
-/** Class for storing a segment that was generated by a model (e.g. RCS, ACS), though we do not necessarily know ahead of time
+/**
+ * Class for storing a segment that was generated by a model (e.g., GATK CNV), though we do not necessarily know ahead of time.
```
davidbenjamin: "Do not necessarily know *which model* ahead of time" is clearer.

src/main/java/org/broadinstitute/hellbender/tools/exome/SNPSegmenter.java
```diff
@@ -16,23 +17,27 @@
...
+                                        final String sampleName, final File outputFile) {
+        try {
+            final File targetsFromSNPCountsFile = File.createTempFile("targets-from-snps", ".tsv");
 
-        List<TargetCoverage> targetsFromSNPCounts = snpCounts.stream()
-                .map(count -> count.toMinorAlleleFractionTargetCoverage("snp-target" + count.getContig() + ":" +
-                        count.getStart() + "_" + count.getEnd())).collect(Collectors.toList());
+            List<TargetCoverage> targetsFromSNPCounts = snps.targets().stream()
+                    .map(count -> count.toMinorAlleleFractionTargetCoverage("snp-target" + count.getContig() + ":" +
+                            count.getStart() + "_" + count.getEnd())).collect(Collectors.toList());
```
davidbenjamin: Do you mean for that underscore to be a dash?

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentMergeUtils.java
```diff
@@ -1,6 +1,7 @@
 package org.broadinstitute.hellbender.tools.exome;
```
davidbenjamin: Note to self and anyone else who may be reading these comments in a linear fashion: I'm skipping this for now in favor of lighter code; will return later.

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
```diff
@@ -13,292 +15,452 @@
...
+        return myTargets.stream().mapToDouble(TargetCoverage::getCoverage).average().getAsDouble();
     }
 
     /**
-     * read a list of segments (with calls, if possible) from a file.  Creates a list of modeled segments
-     *
-     * Segment Means in the input file are assumed to be in CR space (not log2).  Please note that a Modeled segment
-     *  stores the segment mean as log2CR.
+     * Get difference between a segment mean and the overlapping targets.
+     * This will select the overlapping targets from the input collection, for your convenience.
```
davidbenjamin: No need to say "for your convenience."

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
```diff
@@ -13,292 +15,452 @@
...
 
-        final List<TargetCoverage> myTargets = targets.targets(loc);
-
-        if (myTargets.size() == 0) {
-            throw new IllegalStateException("Empty segment -- no overlapping targets.");
+        final List<SimpleInterval> targetSegmentsFixed = new ArrayList<>();
+        for (final SimpleInterval segment : targetSegments) {
+            try {
+                final Locatable firstTarget = targets.targets(segment).get(0);
+                if (firstTarget.getEnd() == segment.getStart()) {
```
davidbenjamin: Why is the `if` necessary?

samuelklee: hmm, i guess i should also add an else throw exception, just in case there are targets in the segment but the segment start doesn't correspond to the first target.  good catch!

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
```diff
@@ -13,292 +15,452 @@
...
             start = snps.target(snpRange.from).getStart();
             end = snps.target(snpRange.to - 1).getEnd();
-        } else if (snpsInInterval == 0) {
-            start = targets.target(targetRange.from).getStart();
-            end = targets.target(targetRange.to - 1).getEnd();
-        } else {
-            start = Math.min(targets.target(targetRange.from).getStart(), snps.target(snpRange.from).getStart());
-            end = Math.max(targets.target(targetRange.to - 1).getEnd(), snps.target(snpRange.to - 1).getEnd());
+        } else if (numTargetsInInterval > 0) {
+            if (!(targets.target(targetRange.from).getStart() <= interval.getStart() &&
```
davidbenjamin: Extract variables for `targets.target(targetRange.from)` and  `targets.target(targetRange.to - 1)`.

davidbenjamin: Why not rewrite `!( a <= b && c <= d)` as ` b < a || d < c`?

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
```diff
@@ -13,292 +15,452 @@
...
+                    interval.getStart() <= targets.target(targetRange.from).getEnd())) {
+                //if interval start does not fall within first target, use start of first target as start of trimmed interval
+                start = targets.target(targetRange.from).getStart();
+            }
+            if (!(targets.target(targetRange.to - 1).getStart() <= interval.getEnd() &&
```
davidbenjamin: Analogous rewrite with deMorgan's laws.

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
davidbenjamin: Even better, use `SimpleInterval`'s `overlaps` method: `if(! new SimpleInterval(interval).overlaps(firstTarget))`.

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
```diff
@@ -13,292 +15,452 @@
...
         final IndexRange snpRange = snps.indexRange(interval);
 
-        final int targetsInInterval = targetRange.size();
-        final int snpsInInterval = snpRange.size();
+        final int numTargetsInInterval = targetRange.size();
+        final int numSNPsInInterval = snpRange.size();
 
-        int start;
-        int end;
+        int start = interval.getStart();
```
davidbenjamin: For some of the comments below, these variables would instead be `SimpleInterval start = new SimpleInterval(interval.getContig(), interval.getStart(), interval.getStart())` and likewise for `end`.

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
davidbenjamin: Or `if(end.overlaps(firstTarget))`

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
davidbenjamin: I'm confused -- the code checks whether the interval starts before the first target or after the first target ends.  But isn't the latter impossible since the first target was derived from `targets.indexRange(interval).from`?  i.e. if a target ends before the interval starts, how can the target be part of the interval's associated index range?  So, it seems that you only need a single condition: `if (start < firstTarget.getStart())`.

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
davidbenjamin: As above, it seems like `if(end > lastTarget.getEnd())` would suffice.

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
```diff
@@ -13,292 +15,452 @@
...
-     * at the start of {@code interval}.
-     *
-     * For example, if {@code interval} is 1:100-200 and the only target or snp it overlaps is 1:60-180, then the result
-     * is 1:60-180.
-     *
-     * Note that if {@code targets} contains overlapping targets we can get an inconsistent result where
-     * parsimoniousInterval returns an interval that is not parsimonious by its own definition.  For example, if {@code targets}
-     * contains 1:40-90 and 1:80-110 and {@code interval} is 1:100-200, then parsimoniousInterval yields 1:80-110, which
-     * falls in the middle of a target.  Since sites in {@code snps} are length-1, there is no such concern for them.
+     * Given an interval and collections of targets and SNPs, returns a trimmed interval produced by removing the empty
```
davidbenjamin: Just to check, this method will not expand an interval to the start of the first overlapping target and the end of the last overlapping target; it will only shrink the interval or leave it alone.  Is this the desired behavior?

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
samuelklee: oops, i guess you're right---the second check is redundant!

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
samuelklee: that's right, because i think we want to avoid overlapping segments (which would occur if a SNP breakpoint falls within a target and we expanded intervals, for example).  i can add some documentation to clear this up.

davidbenjamin: That would be good.

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
```diff
@@ -13,292 +15,452 @@
...
+     * portions at the start and the end of the original interval that do not overlap the targets and SNPs that overlap
+     * with the original interval.  If this procedure would remove the entire interval, the original interval is
+     * returned instead.
      */
-    public static SimpleInterval parsimoniousInterval(final Locatable interval, final TargetCollection<? extends Locatable> targets,
-                                                                     final TargetCollection<? extends Locatable> snps) {
-        Utils.nonNull(interval, "Interval may not be null");
-        Utils.nonNull(targets, "Targets may not be null");
-        Utils.nonNull(snps, "SNPs may not be null");
+    public static SimpleInterval trimInterval(final Locatable interval,
```
davidbenjamin: Nice renaming.  I feel like a doofus for calling it `parsimonious`.

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
```diff
@@ -13,292 +15,452 @@
...
+    //then returns only those untrimmed segments that are non-empty
+    private static List<SimpleInterval> constructUntrimmedSegments(final Genome genome,
+                                                                   final SortedMap<String, List<Breakpoint>> breakpointsByContig) {
+        final List<SimpleInterval> segments = new ArrayList<>();
+        for (final String contig : breakpointsByContig.keySet()) {
+            final List<Breakpoint> breakpoints = breakpointsByContig.get(contig);
+            int start = breakpoints.get(0).getStart();
+            for (int i = 1; i < breakpoints.size(); i++) {
+                final Breakpoint breakpoint = breakpoints.get(i);
+                int end;
```
davidbenjamin: How about  `final int end = breakpoint.getType() - (breakpoint.getType() == START ? 1 : 0)`?

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
```diff
@@ -13,292 +15,452 @@
...
+                                                                   final List<SimpleInterval> segments) {
+        //get original target-segment starts and ends
+        final Set<SimpleInterval> targetSegmentStarts =
+                targetSegments.stream().map(s -> new SimpleInterval(s.getContig(), s.getStart(), s.getStart()))
+                        .collect(Collectors.toSet());
+        final Set<SimpleInterval> targetSegmentEnds =
+                targetSegments.stream().map(s -> new SimpleInterval(s.getContig(), s.getEnd(), s.getEnd()))
+                        .collect(Collectors.toSet());
+
+        final List<SimpleInterval> segmentsWithoutSpuriousEnds = new ArrayList<>();
```
davidbenjamin: To avoid confusion due to the two meanings of "end", name this `segmentsWithoutSpuriousBreakpoints`.

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
```diff
@@ -13,292 +15,452 @@
...
-                    dataLine.append(sample, ms.getContig()).append(ms.getStart(), ms.getEnd())
-                            .append(ms.getOriginalProbeCount())
-                            .append(ms.getSegmentMeanInCRSpace())
-                            .append(ms.getCall());
-                })) {
-            for (final ModeledSegment ms : segments) {
-                writer.writeRecord(Utils.nonNull(ms, "Segments list contains a null."));
+    //remerge spurious target-only segments introduced at starts and ends of the
+    //original target-coverage segments by union of breakpoints
+    private static List<SimpleInterval> mergeSpuriousStartsAndEnds(final List<SimpleInterval> targetSegments,
```
davidbenjamin: `segments` should be the first argument because it contains the segments whose spurious starts and ends are being merged.

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
```diff
@@ -13,292 +15,452 @@
...
 
-        final double segMean = segment.getSegmentMeanInCRSpace();
-        final List<TargetCoverage> myTargets = targets.targets(segment);
-        return myTargets.stream().map(t -> ( Math.pow(2, t.getCoverage()) - segMean)).collect(Collectors.toList());
+                if (end < start) {  //this could happen if there are adjacent breakpoints
+                    continue;
+                }
+
+                final SimpleInterval segment = new SimpleInterval(contig, start, end);
+                if (genome.getTargets().targetCount(segment) > 0 || genome.getSNPs().targetCount(segment) > 0) {
```
davidbenjamin: Could the `mergeSpuriousStartsAndEnds` logic go here?  For spurious segments at the beginning of a target, if `segment` has no SNPs, you could `continue` without adding the segment.  This would keep `start` where it is, effectively merging with the next segment.  For spurious segments at the end of a target, you could merge with the last target added, which would only require a cheap `pop` from `segments`.

davidbenjamin:
@samuelklee review complete.

src/main/java/org/broadinstitute/hellbender/tools/exome/ACSCopyRatioModeller.java
samuelklee: done here and in outlierProbabilitySampler

src/main/java/org/broadinstitute/hellbender/tools/exome/ACSCopyRatioModeller.java
samuelklee: done

src/main/java/org/broadinstitute/hellbender/tools/exome/ACSModeller.java
samuelklee: done

src/main/java/org/broadinstitute/hellbender/tools/exome/AllelicCapSegger.java
samuelklee: done

src/main/java/org/broadinstitute/hellbender/tools/exome/HetPulldownCalculator.java
samuelklee: done

src/main/java/org/broadinstitute/hellbender/tools/exome/ModeledSegment.java
samuelklee: done

src/main/java/org/broadinstitute/hellbender/tools/exome/SNPSegmenter.java
samuelklee: this is just an arbitrary target name so it doesn't matter, but i'll go ahead and change it

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
samuelklee: done (although i don't think this was my text, originally)

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
samuelklee: done

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
samuelklee: cleaned up this code (hopefully enough that extracting variables isn't necessary)

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
samuelklee: done

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
samuelklee: done

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
samuelklee: i'd say most doofuses (doofusi?) don't use words like parsimonious!

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
samuelklee: good call, done

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
samuelklee: went ahead and called it mergedSegments, as in other methods

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
samuelklee: done for this and mergeSpuriousMiddles

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
samuelklee: hmm, probably, but i think it's a bit more clear to have the

construct untrimmed -> merge start/ends -> merge middles 

steps broken up explicitly, since performance isn't an issue here.

src/main/java/org/broadinstitute/hellbender/tools/exome/ACSModeller.java
samuelklee: also didn't change this up, will file an issue

samuelklee:
back to @davidbenjamin.  if everything looks OK, feel free to send it back to me to squash.  not sure what to do if tests with random seeds/race conditions keep failing (i didn't get a chance to look at those yet), should i go ahead and merge?  @vruano and @LeeTL1220 can comment.

davidbenjamin:
@samuelklee Squash and merge!

Issue #265, opened 2015-12-23
Tool for extracting targets and weights from an existing PoN.


LeeTL1220:
We need a tool that can create a target BED file and a target weights file from an existing PoN.

For the target BED:
``pon.getRawTargets()`` for the targets (alternatively, we can produce the targets that remain after filtering with ``pon.getPanelTargets()``)

For the target weights:

``` java
// code snippet for generating the weights from a PoN
// this function probably already exists by the time someone addresses this issue.
final double[] targetVariances = pon.getTargetVariances();
final double[] targetWeights = new double[targetVariances.length];
for (int i = 0; i < targetVariances.length; i++) {
     targetWeights[i] = 1/targetVariances[i];
}
```

davidbenjamin:
Good idea!  If we implement #102 and have an allelic PoN then we would have parallel methods for getting per-target priors on allelic bias.

davidbenjamin:
Self-assigning because I might need `pon.getTargetVariances` for the CN-LOH caller.

LeeTL1220:
@davidbenjamin Assigning to me, since I've already done this in the PoN QC branch.

LeeTL1220:
@davidbenjamin Actually, nevermind.  I've just made it so that the information is there.  We will still need a tool to actually extract this information.  HOWEVER, this should not be addressed until that branch is merged.  And it may make more sense for me to do it.

davidbenjamin:
@LeetTL agreed.

davidbenjamin:
Whoever handles this should probably also address #249.

LeeTL1220:
@davidbenjamin We should also have this tool be used in our workflow scripts to automatically generate weights before launching a segmentation.....  See pull request for PoN QC (comments in issue #283 )

PR #270, opened 2016-01-01
Get targets and target variances from the PoN. Addresses issue #265.

davidbenjamin:


davidbenjamin:
@LeeTL1220 Is this what you had in mind for #265?

davidbenjamin:
Redundant with @LeeTL1220's branch.  Whoops!

PR #273, opened 2016-01-03
Finite HMM api

vruano:
HMM api.
@davidbenjamin  please review.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/hmm/ForwardBackwardAlgorithm.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/hmm/ForwardBackwardAlgorithm.java
@@ -1,0 +1,425 @@
+package org.broadinstitute.hellbender.utils.hmm;
+
+import it.unimi.dsi.fastutil.objects.Object2IntMap;
+import it.unimi.dsi.fastutil.objects.Object2IntOpenHashMap;
+import org.broadinstitute.hellbender.utils.MathUtils;
+import org.broadinstitute.hellbender.utils.Utils;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.IntStream;
+
+/**
+ * Performs the Forward-backward algorithm for
+ * {@link HiddenMarkovModel}
+ * on a sequence of data points and positions.
+ *
+ * <p>This done by calling {@link #apply} that returns an
+ * {@link ForwardBackwardAlgorithm.Result} instance which
+ * allow for subsequent queries on the posterior probability of the
+ * hidden states at arbitrary points of the input sequence</p>
+ *
+ * @author Valentin Ruano-Rubio &lt;valentin@broadinstitute.org&gt;
+ */
+public final class ForwardBackwardAlgorithm {
+
+    /**
+     * The forward-backward algorithm result query interface.
+     *
+     * @param <D> the observed data type.
+     * @param <T> the observation time/position type.
+     * @param <S> the hidden state type.
+     */
+    public interface Result<D, T, S> {
+
+        /**
+         * Returns the data passed to the forward-backward algorithm.
+         * @return never {@code null}.
+         */
+        List<D> data();
+
+        /**
+         * Returns the list of positions passed to the forward-backward algorithm
+         * @return never {@code null}.
+         */
+        List<T> positions();
+
+        /**
+         * Returns the model passed to the foward-backward algorithm.
+         * @return never {@code null}.
+         */
+        HiddenMarkovModel<D, T, S> model();
+
+        /**
+         * Returns the forward probability of the hidden state at certain position.
+         * @param state the query hidden state.
+         * @param position the query position.
+         * @return a valid probability in log scale (from -Inf to 0).
+         *
+         * @throws IllegalArgumentException if either the {@code state} or the {@code position} are not
+         *  recognized by the model or were not passed to the forward-backward algorithm.
+         */
+        double logForwardProbability(final S state, final T position);
+
+        /**
+         * Returns the backward probability of the hidden state at certain position.
+         * @param state the query hidden state.
+         * @param position the query position.
+         * @return a valid probability in log scale (from -Inf to 0)
+         * @throws IllegalArgumentException if either the {@code state} or the {@code position} are not
+         *  recognized by the model or were not passed to the forward-backward algorithm.
+         */
+        double logBackwardProbability(final S state, final T position);
+
+        /**
+         * Returns the posterior probability of the hidden state at certain position.
+         * @param state the query hidden state.
+         * @param position the query position.
+         * @return a valid probability in log scale (from -Inf to 0)
+         * @throws IllegalArgumentException if either the {@code state} or the {@code position} are not
+         *  recognized by the model or were not passed to the forward-backward algorithm.
+         */
+        double logProbability(final S state, final T position);
+
+        /**
+         * Returns the posterior probability of a full sequence of hidden states given the original data
+         * and position sequences.
+         * @param states the list of states.
+         * @return a valid probability in log scale (from -Inf to 0)
+         * @throws IllegalArgumentException if {@code states}, is {@code null}, contains some value not recognized
+         * as a hidden state by the model, or is not of the same length as the original data.
+         */
+        double logProbability(final List<S> states);
+
+        /**
+         * Returns the likelihood of the original data given the model.
+         *
+         * @return a valid probability in log scale (from -Inf to 0)
+         */
+        double logDataLikelihood();
+    }
+
+    /**
+     * Runs the forward-backward algorithm on a data-positions list pair given a type-compatible model.
+     * <p>
+     *     Calling code can then use the returned {@link Result} instance to perform various queries based on
+     *     the calculations performed by the algorithm. Refer to {@link Result} documentation for details.
+     * </p>
+     * <p>
+     *     This method invocation may be computationally intensive with large data-sets. The complexity of the
+     *     algorithm is {@code O(L*(N^2))} where {@code L is the length of the data/position sequence and
+     *     {@code N} is the number of hidden states. The memory size of the result is of the order of {@code O(L*N)}.
+     * </p>
+     *
+     * @param data the observed data sequence.
+     * @param positions the observation time/position points.
+     * @param model the HMM model.
+     * @param <D> the observed data type.
+     * @param <T> the observation time/position type.
+     * @param <S> the hidden state type.
+     * @return never {@code null}.
+     * @throws IllegalArgumentException if any of the arguments is {@code null}, {@code data} and {@code positions}
+     *   have different length, or if the {@code model} does not recognized any of the values in {@code data} or {@code}
+     *   positions.
+     */
+    public static <D, T, S> Result<D, T, S> apply(final List<D> data, final List<T> positions,
+                                 final HiddenMarkovModel<D, T, S> model) {
+        Utils.nonNull(data, "the input data sequence cannot be null.");
+        Utils.nonNull(positions, "the input position sequence cannot be null.");
+        Utils.nonNull(model, "the input model cannot be null");
+
+        final List<D> dataList = Collections.unmodifiableList(new ArrayList<>(data));
+        final List<T> positionList = Collections.unmodifiableList(new ArrayList<>(positions));
+        if (dataList.size() != positionList.size()) {
+            throw new IllegalArgumentException("the data sequence and position sequence must have the same number of elements");
+        }
+
+        final double[][] forwardProbabilities = calculateLogForwardProbabilities(model, dataList, positionList);
+        final double[][] backwardProbabilities = calculateLogBackwardProbabilities(model, dataList, positionList);
+
+        return new ArrayResult<>(dataList, positionList, model, forwardProbabilities, backwardProbabilities);
+    }
+
+    /**
+     * Calculates the forward probabilities (the forward phase of the algorithm). These are in log scale.
+     * <p>
+     *     The returned array is uni-dimensional. Forward probabilities are arranged by position, thus the
+     *     forward probability for the ith hidden state and jth position is stored in the
+     *     index: i + j * number-of-hidden-states.
+     * </p>
+     * @param model the HMM model.
+     * @param data the observed data sequence.
+     * @param positions the observation time/position points.
+     * @param <D> the observed data type.
+     * @param <T> the observation time/position type.
+     * @param <S> the hidden state type.
+     * @return never {@code null}, the calling code can modify this array at will.
+     * @throws IllegalArgumentException if the {@code model} does not recognize any of the values in {@code data}
+     *   or {@code positions}.
+     */
+    private static <D, T, S> double[][] calculateLogForwardProbabilities(
+            final HiddenMarkovModel<D, T, S> model,
+            final List<D> data,
+            final List<T> positions) {
+
+        final List<S> states = model.hiddenStates();
+        final int numStates = states.size();
+        final int length = data.size();
+        final double[][] result = new double[length][numStates];
+        // Empty data? then we just return the empty array:
+        if (length == 0) {
+            return result;
+        }
+        // We first initialize the probabilities at the first position.
+        final T position0 = positions.get(0);
+        final D datum0 = data.get(0);
+        for (int stateIndex = 0; stateIndex < states.size(); stateIndex++) {
+            final S state = states.get(stateIndex);
+            result[0][stateIndex] =
+                    model.logPriorProbability(state, position0)
+                + model.logEmissionProbability(datum0, state, position0);
+        }
+
+        // Then we do the rest t_1, t_2, ... and so on.
+        // Array re-used to hold the elements of log-sum-exp operations:
+        final double[] logSumBuffer = new double[states.size()];
+        for (int thisPositionIndex = 1; thisPositionIndex < data.size(); thisPositionIndex++) {
+            final int previousPositionIndex = thisPositionIndex - 1;
+            final T previousPosition = positions.get(previousPositionIndex);
+            final T thisPosition = positions.get(thisPositionIndex);
+            for (int thisStateIndex = 0; thisStateIndex < numStates; thisStateIndex++) {
+                final S thisState = states.get(thisStateIndex);
+                for (int previousStateIndex = 0; previousStateIndex < numStates; previousStateIndex++) {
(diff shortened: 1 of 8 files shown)


src/main/java/org/broadinstitute/hellbender/utils/hmm/ForwardBackwardAlgorithm.java
```diff
@@ -0,0 +1,432 @@
...
+import java.util.stream.IntStream;
+
+/**
+ * Performs the Forward-backward algorithm for
+ * {@link HiddenMarkovModel}
+ * on a sequence of data points and positions.
+ *
+ * <p>This done by calling {@link #apply} that returns an
+ * {@link ForwardBackwardAlgorithm.Result} instance which
+ * allow for subsequent queries on the posterior probability of the
```
davidbenjamin: Nice design decision.

src/main/java/org/broadinstitute/hellbender/utils/hmm/ForwardBackwardAlgorithm.java
```diff
@@ -0,0 +1,432 @@
...
+ * hidden states at arbitrary points of the input sequence</p>
+ *
+ * @author Valentin Ruano-Rubio &lt;valentin@broadinstitute.org&gt;
+ */
+public final class ForwardBackwardAlgorithm {
+
+    /**
+     * The forward-backward algorithm result query interface.
+     *
+     * @param <D> the data observation type.
```
davidbenjamin: "observed data type" reads better.

src/main/java/org/broadinstitute/hellbender/utils/hmm/HiddenMarkovModel.java
```diff
@@ -0,0 +1,108 @@
+package org.broadinstitute.hellbender.utils.hmm;
+
+import java.util.List;
+
+/**
+ * Hidden Markov Model interface with hidden-state transition probabilities and data
+ * emission that might depends on the position.
+ *
+ * <p>
+ * This class models a process that emits a fix length data sequence.
```
davidbenjamin: fixed-length

src/main/java/org/broadinstitute/hellbender/utils/hmm/HiddenMarkovModel.java
```diff
@@ -0,0 +1,108 @@
...
+ * @param <T> represent the observation position type.
+ */
+public interface HiddenMarkovModel<D, T, S> {
+
+    /**
+     * Returns all the hidden states.
+     * <p>
+     *     Contract:
+     *     <ul>
+     *         <li>This method must always return the same state every time is called.</li>
```
davidbenjamin: same *states* every time *it* is called.

src/main/java/org/broadinstitute/hellbender/utils/hmm/ForwardBackwardAlgorithm.java
```diff
@@ -0,0 +1,432 @@
...
+        Utils.nonNull(data);
+        Utils.nonNull(positions);
+        Utils.nonNull(model);
+        final List<D> dataList = Collections.unmodifiableList(new ArrayList<>(data));
+        final List<T> positionList = Collections.unmodifiableList(new ArrayList<>(positions));
+        if (dataList.size() != positionList.size()) {
+            throw new IllegalArgumentException("the data sequence and position sequence must have the same number of elements");
+        }
+        final List<S> states = model.hiddenStates();
+        Utils.nonNull(states, "the model hidden state list cannot be null");
```
davidbenjamin: Shouldn't `HiddenMarkovModel` itself (or its implementations) enforce this?

src/main/java/org/broadinstitute/hellbender/utils/hmm/ForwardBackwardAlgorithm.java
```diff
@@ -0,0 +1,432 @@
...
+        if (states.size() == 0) {
```
davidbenjamin: Likewise.

src/main/java/org/broadinstitute/hellbender/utils/hmm/ForwardBackwardAlgorithm.java
```diff
@@ -0,0 +1,432 @@
...
+     * @param dataList the observed data sequence.
+     * @param positionList the observation time/position points.
+     * @param <D> the observed data type.
+     * @param <T> the observation time/position type.
+     * @param <S> the hidden state type.
+     * @return never {@code null}, the calling code can modify this array at will.
+     * @throws IllegalArgumentException if the {@code model} does not recognize any of the values in {@code dataList}
+     *   or {@code positionList}.
+     */
+    private static <D, T, S> double[] calculateLogBackwardProbabilities(final HiddenMarkovModel<D, T, S> model,
```
davidbenjamin: Why not `double[][]`?

src/main/java/org/broadinstitute/hellbender/utils/hmm/ForwardBackwardAlgorithm.java
```diff
@@ -0,0 +1,432 @@
...
+     * @param data the observed data sequence.
+     * @param positions the observation time/position points.
+     * @param <D> the observed data type.
+     * @param <T> the observation time/position type.
+     * @param <S> the hidden state type.
+     * @return never {@code null}, the calling code can modify this array at will.
+     * @throws IllegalArgumentException if the {@code model} does not recognize any of the values in {@code data}
+     *   or {@code positions}.
+     */
+    private static <D, T, S> double[] calculateLogForwardProbabilities(
```
davidbenjamin: Forward probabilities should go above backward probabilities in the code.

src/main/java/org/broadinstitute/hellbender/utils/hmm/ForwardBackwardAlgorithm.java
```diff
@@ -0,0 +1,432 @@
...
+        final int width = states.size();
+        final int length = dataList.size();
+        final double[] result = new double[width * length];
+
+        // Empty data? then we just return the empty array:
+        if (length == 0) {
+            return result;
+        }
+
+        // We first initialize the probabilities at the first position.
```
davidbenjamin: This comment should be very explicit about the fact that we initialize the last position of the chain, which is the first position of the backward pass.

src/main/java/org/broadinstitute/hellbender/utils/hmm/ForwardBackwardAlgorithm.java
```diff
@@ -0,0 +1,432 @@
...
+        // result already contains appropriate values for the backward-probabilities
```
davidbenjamin: "Result already contains" implies that this follows from a calculation.    Better just to say we initialize to 1.

src/main/java/org/broadinstitute/hellbender/utils/hmm/ForwardBackwardAlgorithm.java
```diff
@@ -0,0 +1,432 @@
...
+     * @return never {@code null}, the calling code can modify this array at will.
+     * @throws IllegalArgumentException if the {@code model} does not recognize any of the values in {@code dataList}
+     *   or {@code positionList}.
+     */
+    private static <D, T, S> double[] calculateLogBackwardProbabilities(final HiddenMarkovModel<D, T, S> model,
+                                                                   final List<D> dataList,
+                                                                   final List<T> positionList) {
+
+        final List<S> states = model.hiddenStates();
+        final int width = states.size();
```
davidbenjamin: `numStates` would be clearer.

src/main/java/org/broadinstitute/hellbender/utils/hmm/ForwardBackwardAlgorithm.java
```diff
@@ -0,0 +1,432 @@
...
+        final double[] result = new double[width * length];
+
+        // Empty data? then we just return the empty array:
+        if (length == 0) {
+            return result;
+        }
+
+        // We first initialize the probabilities at the first position.
+        // result already contains appropriate values for the backward-probabilities
+        // of the last state = 1, (0 in log scale).
```
davidbenjamin: Also, make it explicit that declaring the `double[]` automatically initializes to zero.

src/main/java/org/broadinstitute/hellbender/utils/hmm/ForwardBackwardAlgorithm.java
```diff
@@ -0,0 +1,432 @@
...
+
+        final double[] logSumBuffer = new double[states.size()];
+        for (int t = length - 2; t >= 0; --t) {
```
davidbenjamin: `t` and `u` need clearer names, like `thisPosition` and `nextPosition` or `here` and `next` etc.

src/main/java/org/broadinstitute/hellbender/utils/hmm/ForwardBackwardAlgorithm.java
```diff
@@ -0,0 +1,432 @@
...
+            final int u = t + 1;
+            final int tOffset = width * t;
+            final int uOffset = tOffset + width;
+            final T positionT = positionList.get(t);
+            final T positionU = positionList.get(u);
+            for (int sT = 0; sT < width; sT++) {
```
davidbenjamin: Similarly, more evocative names like `thisState` and `nextState` etc.  (Not saying that these particular suggestions are good, but you get the idea).

src/main/java/org/broadinstitute/hellbender/utils/hmm/ForwardBackwardAlgorithm.java
```diff
@@ -0,0 +1,432 @@
...
+     * @throws IllegalArgumentException if the {@code model} does not recognize any of the values in {@code data}
+     *   or {@code positions}.
+     */
+    private static <D, T, S> double[] calculateLogForwardProbabilities(
+            final HiddenMarkovModel<D, T, S> model,
+            final List<D> data,
+            final List<T> positions) {
+
+        final List<S> states = model.hiddenStates();
+        final int width = states.size();
```
davidbenjamin: Similar comments to those in backward probability code about variable names.

src/main/java/org/broadinstitute/hellbender/utils/hmm/ForwardBackwardAlgorithm.java
```diff
@@ -0,0 +1,432 @@
...
+        }
+
+        /**
+         * Calculates the data likelihood in log scale.
+         * <p>
+         *     This value can be obtain by adding up the posterior probabilities at any position; they are all supposed to
+         *     be the same.
+         * </p>
+         * <p>
+         *     This implementation uses always the first position.
```
davidbenjamin: always uses

src/main/java/org/broadinstitute/hellbender/utils/hmm/ForwardBackwardAlgorithm.java
```diff
@@ -0,0 +1,432 @@
...
+        private final List<D> data;
+
+        private final List<T> positions;
+
+        private final HiddenMarkovModel<D, T, S> model;
+
+        private final Object2IntMap<T> positionIndex;
+        private final Object2IntMap<S> stateIndex;
+
+        private final int width;
```
davidbenjamin: `numStates`

src/main/java/org/broadinstitute/hellbender/utils/hmm/ForwardBackwardAlgorithm.java
```diff
@@ -0,0 +1,432 @@
...
+         * @param width number of hidden states.
+         * @return a valid probability in log scale (between -Inf and 0 inclusive).
+         */
+        private static double calculateLogDataLikelihood(final double[] logForwardProbabilities,
+                                                         final double[] logBackwardProbabilities,
+                                                         final int width) {
+            if (logForwardProbabilities.length == 0) {
+                return 0;
+            } else {
+                final double[] values = new double[width];
```
davidbenjamin: If probabilities are 2-D arrays this code becomes `return logSumLog(arraySum(forward[0], backward[0]))`.

src/main/java/org/broadinstitute/hellbender/utils/hmm/HiddenMarkovModel.java
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * <p>
+ * This class models a process that emits a fix length data sequence.
+ * </p>
+ * <p>
+ * At each data point (or position) the process is in one of a discrete set
+ * of hidden states which are shared by all positions.
+ * </p>
+ * <p>
+ * The emission probability of a datum given the hidden state may depend on the current hidden state itself and also the position in the sequence.
```
davidbenjamin: omit "itself"

src/main/java/org/broadinstitute/hellbender/utils/hmm/ViterbiAlgorithm.java
```diff
@@ -0,0 +1,178 @@
...
+
+/**
+ * Implements the Viterbi Algorithm.
+ *
+ * @author Valentin Ruano-Rubio &lt;valentin@broadinstitute.org&gt;
+ */
+public final class ViterbiAlgorithm {
+
+    /**
+     * Calculates the most likely the hidden state sequence that explains the observed data
```
davidbenjamin: extra "the"

src/main/java/org/broadinstitute/hellbender/utils/hmm/ViterbiAlgorithm.java
```diff
@@ -0,0 +1,178 @@
...
+            final Path<S>[] bestTPaths = bestPaths[1 & t]; // time t best paths; previous best paths.
+            final T positionT = positions.get(t);
+            final D dataU = data.get(u);
+            final T positionU = positions.get(u);
+            for (int sU = 0; sU < width; sU++) {
+                final S stateU = states[sU];
+                // Initialize best-t-state search setting the best so
+                // far to 0th indexed state:
+                int bestTStateIndex = 0;
+                double bestTStateLog10Prob = bestTPaths[0].log10Prob
```
davidbenjamin: There is a mix of `log` and `log10` variable names.

src/main/java/org/broadinstitute/hellbender/utils/hmm/ViterbiAlgorithm.java
```diff
@@ -0,0 +1,178 @@
...
+        // t and t + 1 (u).
+        // The array index for time x is determined by the last bit in x.
+        @SuppressWarnings({"rawtypes", "unchecked"})
+        final Path<S>[][] bestPaths = (Path<S>[][]) new Path[][] {
+            initializeBestPaths(data, positions, model, states), // best paths at time 0.
+            (Path<S>[]) new Path[width]                      // empty best paths for time 1.
+        };
+
+        for (int u = 1; u < length; u++) {
+            final int t = u - 1;
```
davidbenjamin: Same stuff about variable names.

src/test/java/org/broadinstitute/hellbender/utils/hmm/HMMUnitTest.java
```diff
@@ -0,0 +1,542 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+/**
+ * Unit test for HMM api classes including {@link ViterbiAlgorithm}, {@link ForwardBackwardAlgorithm} and
+ * {@link ForwardBackwardAlgorithm.Result}.
+ *
+ * @author Valentin Ruano-Rubio &lt;valentin@broadinstitute.org&gt;
+ */
+public final class HMMUnitTest extends BaseTest {
```
davidbenjamin: I'm uncomfortable about relying so heavily on R for these tests.  You should supplement it with a few HMMs that are a priori trivial.  For example:

* If priors and transitions are all equal, then every path must be equally likely.
* If one state has the likeliest prior and most likely transitions from any state are into that state, then the likeliest path is that state repeated over and over.
* If you have a long chain with (irreducible) transition matrix M, the state probabilities asymptotically approach the eigenvector with largest eigenvalue of M.  (I consider this test very important because it would be basically impossible to get the right answer by coincidence).

src/test/java/org/broadinstitute/hellbender/utils/hmm/TestHMModel.java
```diff
@@ -0,0 +1,232 @@
...
+
+        final double uniform = rdn.nextDouble() * total;
+        double accumulate = 0;
+        for (int i = 0; i < Datum.values().length; i++) {
+            accumulate += Math.exp(logEmissionProbs[i]);
+            if (uniform < accumulate) {
+                return Datum.values()[i];
+            }
+        }
+        throw new IllegalStateException("may be some of the emission probs is negative?");
```
davidbenjamin: are negative

davidbenjamin:
@vruano Back to you.  It's a nice design, very clear API.  My main suggestion is to supplement the R testing with a few analytic results.  Also, I wasn't fully comfortable with the performance hit  representing states by objects instead of integer indices, but then I saw your testing code's use of `ordinal()`  I can't say I love this solution, but I don't know of any better way and I do like the flexibility of using either an `Integer` or an `Enum`.

src/test/java/org/broadinstitute/hellbender/utils/hmm/HMMUnitTest.java
davidbenjamin: A simpler and perfectly adequate version of that last one is that if the initial probabilities vector is v and the (position-dependent) transition matrices are M_1, M_2. . . then the probabilities vector at position k is M_k M_{k-1} . . . M_2 M_1 v.

src/main/java/org/broadinstitute/hellbender/utils/hmm/ForwardBackwardAlgorithm.java
vruano: It might sound like premature optimization, but I rather don't have ~150,000 3-position arrays.

vruano:
@davidbenjamin you can always use Integer typed hidden-states if that is what you need.

vruano:
@davidbenjamin  back to you. 
I have addressed all the comments in three commits 
196acc5 : simple comments
271b82f : HMM enhanced unit testing.
7413b6a : FW tables from 1D to 2D arrays.

davidbenjamin:
Very small comments.  I love the HMM tests.  Feel free to merge when you're done.

PR #274, opened 2016-01-04
Add flag to plot X/Y chromosomes

achevali:


davidbenjamin:
The lone failure is an unrelated issue (and possibly fixed by an open pull request), so I will proceed with review.

src/main/java/org/broadinstitute/hellbender/cmdline/ExomeStandardArgumentDefinitions.java
```diff
@@ -38,6 +38,9 @@
     public static final String LOG2_LONG_NAME= "log2Input";
     public static final String LOG2_SHORT_NAME = "LOG";
 
+    public static final String INCLUDE_SEX_CHROMOSOMES_LONG_NAME= "sexChrs";
```
davidbenjamin: Write out the long name completely: `sexChromosomes`.

davidbenjamin:
@achevali Just lengthen that one variable name and then merge.  The travis failure is being dealt with elsewhere.

PR #283, opened 2016-01-08
Issue 217 pon qc -- big

### LINKED ISSUE #214: Support target weights as a parameter to ``PerformSegmentation``
Format is TBD, but should be a simple text file.  It can be the same as the CBS input.  

In order to close this issue, we must also have a way of generating the target weight file from a PoN file.

### LINKED ISSUE #217: PoN QC during creation
The PoN QC step from ReCapSeg is not in the GATK version.  Therefore, large events could get into the PoN.  This would be problematic as those events would be undetectable by GATK CNV.

### LINKED ISSUE #218: Automate the creation of a PoN without samples that fail PoN QC
Do not implement before issue #217 

From recapseg repo:
```


We can approximate:
Remove columns (samples) from fnt_control_matrix after target_factor medians, which means only redoing ~15 minutes of PoN creation

Or we can roll back all the way without the samples we want to remove, which means rerunning all of pon creation. This is probably necessary if we remove a large percentage (question), and will be fast for users with a small number of normals/targets.

```

### LINKED ISSUE #247: Many ParamUtils checks will pass even if input is NaN
For example, ``ParamUtils.inRange`` will pass if input is NaN

### LINKED ISSUE #264: File issue in recapseg-hb-eval for modifying workflow to include PoN weight file for segmentation.

### LINKED ISSUE #275: Store information for extracting targets and target variances in the PoN when it is created.
- Targets
- Target variances (panel)
- Raw targets (the targets used in the initial creation)
- Raw target names

Variance for each target is calculated by tangent normalizing the normals into the reduced hyperplane and measuring the variance.

LeeTL1220:
There are quite a few changes in this pull request.

- Added the PoN Arm-Level Event Detector, which runs, by default, when creating a PoN.  This is implemented in the PoNIssueDetector.  This class can be extended to include other PoN QC steps.
- PoNIssueDetector has arm-level event detection in both Spark and non-spark versions.
- CreatePanelOfNormals CLI refactored and made into a slimmer CLI class and created several backend objects (e.g. HDF5PoNCreator )
- CreatePanelOfNormals CLI has flag to skip QC step and create the PoN immediately with the input BAM files.
- CreatePanelOfNormals CLI will produce a blacklist file of the samples that are tagged by the PoNIssueDetector.
- CreatePanelOfNormals CLI will produce a weights file that can be used in PerformSegmentation.
- Default locations of blacklist and weights file can be overridden by the user in CreatePanelOfNormals CLI.
- PerformSegmentation CLI can now take optional weights file.
- Added TangentNormalizer class that can do several tangent normalization operations.  Will usually return an instance of TangentNormalizationResult.  Spark and non-spark workflows are implemented.  
- Added method to TangentNormalizer that can allow tangent normalization of all normals in a PoN into the PoN.  This is useful for arm-level event detection and calculating target variances.  
- Targets in reduced panel are stored in the PoN as String table (virtually -- see below).  
- Raw targets are stored in the PoN (totally unfiltered -- identical to input file during creation).
- The storage of targets allows for future tool to extract the targets (issue 265)
- Added generic ability to store String tables in the HDF5 PoN.  Since string tables are not actually supported (yet) in HDF5, the implementation stores a String array and a number of columns field.  The wrapping code hides this implementation detail.
- Target variances are stored in the PoN.  The weight file that is generated is 1/variance for each target.
- Created HDF5PoNCreator class that handles the PoN creation independently from any front-end.  Can also redo the calculation of the reduced PoN related fields without re-creating the entire PoN.  This class contains both Spark and non-spark workflow.
- PoN.java interface updated to reflect the availability of these new fields.
- Added/extended unit tests for comparison of all fields in PoNs.
- Fixed issue with modifiable/unmodifiable PoN field copies not being honored in implementation.
- Created RamPoN, which is a full PoN that is stored entirely in RAM.  This is used for redo'ing reduction in HDFPoNCreator and is useful for testing.
- NormalizeSomaticReadCounts CLI refactored and made into a slimmer CLI class and created several backend objects (e.g. TangentNormalizer)
- Several more steps in PoN creation are now implemented in Spark (non-spark workflow is preserved and still accessible by end users).  Documentation provided that labels these steps.
- Default non-Spark SVD to ojAlgo implementation.  Note that Apache Commons RealMatrix are still used throughout, so conversion is performed.  Also note that this conversion takes very little time (on 150k x 1374 matrix)
- Some testdata generation matlab scripts are provided to ease changes down the road.  These are not really supported and are meant for future developers.
- Case2PoNTargetMapper was split into separate class, since several steps in PoN creation (e.g. target variances calculation and PoN QC)
- More documentation added.
- ReadCountCollection now serializable to make it play nice with Spark.  Test added to make sure it can be serialized/deserialized.
- Target now serializable to make it play nice with Spark.  Test added to make sure it can be serialized/deserialized.
- Corrected some tests in ParamUtils that would behave incorrectly if given a NaN.
- Added methods for simple file reading/writing in ParamUtils.  
- Change RCBSSegmenter to (optionally) use a weights file and accept a minWidth parameter.  See CBS docs for more info.
- Added utility class to aid in the conversion of Spark distributed and local matrices and Apache RealMatrix.
- SVD now uses ojAlgo for the non-spark implementation of SVD.
- Updated the GATK CBS.R file to accept the weights file.  This was esecially necessary since there is a bug in CBS ``segment`` that will not allow weights=NULL to be specified, even though that is the default value.  This is because it checks whether the parameter is specified and if it is, will crash if given NULL.  (shaking fist at R)
- Added integration tests for CreatePanelOfNormals with/without Spark, with/without QC.
- Added ``assertBasicPoNAssumptions`` to ``CreatePanelOfNormalsIntegrationTest.java`` that tests some basic PoN coherence.
- Created PoNTestUtils, which contains some good utility methods for testing PoN logic in our unit/CLI tests.
- Added testing of RamPoN
- Added testing of segmentation with new weights
- Added tests that test whether TangentNormalizer can properly tangent normalize normals in a PoN.
- Moved many of the CreatePanelOfNormalsIntegrationTest tests into the unit test for HDF5PoNCreator.  Added a few more for the target variances.
- Added relevant test data
- Updated pon ground truth testdata with the new fields.  These were verified through manual means and through other tests (mentioned above) before being integrated into the test files.  Matlab scripts are provided.

closes #275
closes #264
closes #247
closes #218
closes #217
closes #214


LeeTL1220:
@davidbenjamin One comment to myself: Move ``docs/PoN`` into ``docs/CNV/PoN/`` and move the ``docs/PoN/testdataGeneration`` into ``scripts/CNV/PoN/testdataGeneration/``.  Seem reasonable?

davidbenjamin:
@LeeTL1220 seems reasonable.

LeeTL1220:
@davidbenjamin I need to rebase due to other PRs being merged.  One moment...

LeeTL1220:
@davidbenjamin running unit tests locally before I commit and force push.

LeeTL1220:
@davidbenjamin Updated code due to required new rebasing due to merging of simplify beta hats branch

docs/PoN/testdataGeneration/generate_normal_projections.m
```diff
@@ -0,0 +1,30 @@
...
+
+    % Tangent normalize the fnt_control_matrix
+    tangent_norm_results = zeros(size(reduced_panel,1), size(fnt_control_matrix, 2));
+    for i = 1:size(fnt_control_matrix, 2)
+        sample = fnt_control_matrix(:,i);
+        sample = log2(sample);
+        sample = sample - median(sample);
+        
+        beta_hat = sample' * reduced_pinv';
+        orig_plane_projection = reduced_panel * beta_hat';
```
davidbenjamin: this is equivalent to
```
beta_hat = reduced_pinv * sample
projection = reduced_panel * beta_hat
```

src/main/java/org/broadinstitute/hellbender/tools/exome/Case2PoNTargetMapper.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.stream.IntStream;
+
+/**
+ * Utility class use to map matrices rows between two different target lists.
+ * <p>
+ *     The "case" target list is provided by a {@link ReadCountCollection}, whereas the "pon" target list
+ *     is provided by a {@link PoN}.
+ * </p>
+ * <p>
+ *     This class contains method to re-organized a matrix rows that is based on one list to match the order
```
davidbenjamin: I know it was copy-and-paste, but fix the syntax on this line.

src/main/java/org/broadinstitute/hellbender/tools/exome/Case2PoNTargetMapper.java
```diff
@@ -0,0 +1,104 @@
...
+ *     impose by the other list.
+ * </p>
+ * <p>
+ *     It also provided methods to create a PoN re-organized count matrix directly from a {@link ReadCountCollection}.
```
davidbenjamin: provides

src/main/java/org/broadinstitute/hellbender/tools/exome/CreatePanelOfNormals.java
```diff
@@ -187,103 +182,95 @@
...
+    @Argument(
+            doc = "Output file for sample names that failed the QC check (one sample name per line; sample names drawn from input file).  This option is ignored if QC step is being skipped.  Final PoN will not include these samples, if QC is enabled.",
+            shortName = BLACKLIST_QC_SHORT_NAME,
+            fullName = BLACKLIST_QC_FULL_NAME,
+            optional = true
+    )
+    protected File blacklistOutFile = null;
+
+    @Argument(
+            doc = "Output file for the weights of each target.  This is the 1/var(post-projected targets for each normal)." +
```
davidbenjamin: Personally, I would rather store the weights in the PoN, even at the cost of forcing our command line tool generating a weights file to pass to CBS on the fly, than have an extra file floating around.  I assume the performance hit would be negligible compared to the runtime of CBS, right?

src/main/java/org/broadinstitute/hellbender/tools/exome/CreatePanelOfNormals.java
```diff
@@ -309,146 +296,18 @@ private OptionalInt calculatePreferredNumberOfEigenSamples() {
...
-        final double[] columnMedians = new double[columnCount];
-        final double[][] data = counts.getData();
-        final Median medianCalculator = new Median();
-        final double[] columnData = new double[data.length];
-        for (int i = 0; i < columnCount; i++) {
-            for (int j = 0; j < data.length; j++) {
-                columnData[j] = data[j][i];
-            }
-            columnMedians[i] = medianCalculator.evaluate(columnData);
+    private double calculateExtremeColumnMedianCountsPercentileThreshold() {
```
davidbenjamin: IMO it's preferable for method names to refer to targets and samples rather than rows and column -- the underlying matrix isn't part of the API.

davidbenjamin: `calculate` is misleading since it doesn't calculate anything.  Not exactly a getter either, but there's got to be a good word.

src/main/java/org/broadinstitute/hellbender/tools/exome/CreatePanelOfNormals.java
```diff
@@ -309,146 +296,18 @@ private OptionalInt calculatePreferredNumberOfEigenSamples() {
...
-        final Median medianCalculator = new Median();
-        for (int i = 0; i < columnCount; i++) {
-            final double[] colCounts = counts.getColumn(i);
-            final double median = medianCalculator.evaluate(colCounts);
-            final double invMedian = 1.0 / median;
-            for (int j = 0; j < colCounts.length; j++) {
-                colCounts[j] = Math.log(Math.max(EPSILON, colCounts[j] * invMedian)) * INV_LN_2;
-            }
-            counts.setColumn(i, colCounts);
+    private double calculateTargetFactorsPercentileThreshold() {
```
davidbenjamin: Likewise about "calculate"

src/main/java/org/broadinstitute/hellbender/tools/exome/CreatePanelOfNormals.java
```diff
@@ -462,386 +321,42 @@ private double checkCountTruncatePercentile() {
...
-        // Determine kept and dropped column sets.
-        final Set<String> columnsToKeep = new HashSet<>(readCounts.columnNames().size());
-        final Map<String, Double> columnsToDrop = new LinkedHashMap<>((int) (extremeColumnMedianCountPercentileThreshold * 4.0 / 100.0));
-        for (int i = 0; i < columnMedians.length; i++) {
-            if (columnMedians[i] >= bottomExtremeThreshold && columnMedians[i] <= topExtremeThreshold) {
-                columnsToKeep.add(columnNames.get(i));
-            } else {
-                columnsToDrop.put(columnNames.get(i), columnMedians[i]);
-            }
+    private double checkTargetPercentage() {
```
davidbenjamin: Aha! "check" is a good substitute for those "calculate"s above.

src/main/java/org/broadinstitute/hellbender/tools/exome/NormalizeSomaticReadCounts.java
```diff
@@ -140,45 +114,17 @@ protected Object doWork() {
...
-        final RealMatrix tangentNormalizedCounts = pon.tangentNormalization(tangentNormalizationInputCounts, tangentBetaHats, true);
-
-        // Output the tangent normalized counts.
-        final ReadCountCollection tangentNormalized = targetMapper.fromPoNtoCaseCountCollection(tangentNormalizedCounts, originalCounts.columnNames());
-        writeTangentNormalizedOutput(tangentNormalized);
-        final ReadCountCollection preTangentNormalized = targetMapper.fromPoNtoCaseCountCollection(tangentNormalizationInputCounts, originalCounts.columnNames());
-        writePreTangentNormalizationOutput(preTangentNormalized);
-        writeTangentBetaHats(tangentBetaHats, originalCounts.columnNames());
+    private void outputTangentNormalizationResult(final TangentNormalizationResult tangentNormalizationResult){
+        writeTangentNormalizedOutput(tangentNormalizationResult.getTangentNormalized());
```
davidbenjamin: Write files in order of their calculation, i.e. factor-normalized counts first.

src/main/java/org/broadinstitute/hellbender/tools/exome/PerformSegmentation.java
```diff
@@ -1,9 +1,11 @@
 package org.broadinstitute.hellbender.tools.exome;
 
-import org.broadinstitute.hellbender.cmdline.*;
+import org.broadinstitute.hellbender.cmdline.*;import org.broadinstitute.hellbender.cmdline.*;
```
davidbenjamin: Stray copy and paste?

src/main/java/org/broadinstitute/hellbender/tools/exome/PerformSegmentation.java
```diff
@@ -42,13 +47,21 @@
     )
     protected Boolean log = false;
 
+    @Argument(
+            doc = "File with target weights.  Typically, 1/var(target).  Listed one value per line in plain text.  Values of zero or less, Nan, Inf, and -Inf are not acceptable.  Must have the same number of values as there are in the tangentFile.",
```
davidbenjamin: Be explicit about what this variance is, same as you did above in PoN creation.

src/main/java/org/broadinstitute/hellbender/tools/exome/PoNIssueDetector.java
```diff
@@ -0,0 +1,163 @@
...
+     *
+     * @param singleSampleTangentNormalized Tangent normalized data for a single sample.
+     * @return never {@code null}
+     */
+    private static Boolean hasSuspiciousContigs(final ReadCountCollection singleSampleTangentNormalized, final Map<String, Double> contigToMedian) {
+
+        final List<String> allContigsPresent = retrieveAllContigsPresent(singleSampleTangentNormalized);
+        for (String contig: allContigsPresent) {
+            final ReadCountCollection oneContigReadCountCollection = singleSampleTangentNormalized.subsetTargets(singleSampleTangentNormalized.targets().stream().filter(t -> t.getContig().equals(contig)).collect(Collectors.toSet()));
+            final RealMatrix counts = oneContigReadCountCollection.counts();
```
davidbenjamin: Might as well work with a `RealVector` instead of a matrix starting from this line.

src/main/java/org/broadinstitute/hellbender/tools/exome/ReductionResult.java
```diff
@@ -0,0 +1,37 @@
+package org.broadinstitute.hellbender.tools.exome;
+
+import com.google.api.client.repackaged.com.google.common.annotations.VisibleForTesting;
+import org.apache.commons.math3.linear.RealMatrix;
+
+/**
+ * Structure to hold the result of the SVD/reduction steps.
```
davidbenjamin: steps of PoN creation.

src/main/java/org/broadinstitute/hellbender/tools/exome/ReductionResult.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+public final class ReductionResult {
+    final private RealMatrix pseudoInverse;
+    final private RealMatrix reducedCounts;
+    final private RealMatrix reducedInverse;
+    final private double[] allSingularValues;
+
+    public ReductionResult(final RealMatrix pseudoInverse, final RealMatrix reducedCounts, final RealMatrix reduceInverse, final double[] allSingularValues) {
```
davidbenjamin: argument should be reducedInverse with a 'd'

src/main/java/org/broadinstitute/hellbender/tools/exome/TangentNormalizer.java
```diff
@@ -0,0 +1,282 @@
...
+    private static TangentNormalizationResult tangentNormalize(final PoN pon, final ReadCountCollection targetFactorNormalizedCounts, JavaSparkContext ctx) {
+
+        Utils.nonNull(pon, "PoN cannot be null.");
+        Utils.nonNull(targetFactorNormalizedCounts, "targetFactorNormalizedCounts cannot be null.");
+        Utils.nonNull(targetFactorNormalizedCounts.columnNames(), "targetFactorNormalizedCounts column names cannot be null.");
+        ParamUtils.isGreaterThanZero(targetFactorNormalizedCounts.columnNames().size(), "targetFactorNormalizedCounts column names cannot be an empty list.");
+
+        final Case2PoNTargetMapper targetMapper = new Case2PoNTargetMapper(targetFactorNormalizedCounts.targets(), pon.getPanelTargetNames());
+
+        // The input counts with row (targets) sorted so that the match the PoN's matrix order.
```
davidbenjamin: rows with an 's'

src/main/java/org/broadinstitute/hellbender/tools/exome/TangentNormalizer.java
```diff
@@ -0,0 +1,282 @@
...
+        logger.info("Tangent normalizing the normals (normalized by target factors) ...");
+
+        // For each sample in the PoN, tangent normalize against the qc reduced PoN.
+        return TangentNormalizer.tangentNormalize(pon, coverageProfile, ctx);
+    }
+
+    /**
+     * Prepares the data to perform tangent factor normalization.
+     * <p>
+     * This is done count group or column:
```
davidbenjamin: by count group or column

src/main/java/org/broadinstitute/hellbender/tools/exome/TangentNormalizer.java
```diff
@@ -0,0 +1,282 @@
...
+     *   <ol>
+     *     </li>we change counts to their ratio versus the column mean,</li>
```
davidbenjamin: We divide counts by the column mean

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+     * When using this method in testing, see {@link CreatePanelOfNormals} for good default values.
+     *
+     * @param ctx If no spark context is available, specify {@code null}
+     * @param inputPCovFile  pcov file from {@link CombineReadCounts} with samples as columns.  Never {@code null}
+     * @param numberOfEigenSamples  desired number of eigen samples in the final PoN.  If missing, will apply a rule to determine.
+     * @param sampleNameBlacklist  sample names in {@code inputPCovFile} that should be ignored.  Names must match exactly.  Never {@code null}.  Entires in this parameter that are not in {@code inputPCovFile} are ignored.
+     * @param outputHDF5Filename  final filename for the output PoN HDF5 file
+     * @param isDryRun  whether this is a dry run.  If you wish to do diagnostics and skip the actual writing of a PoN file, set this to true.
+     * @param initialTargets  the set of targets that was used to generate the {@code inputPCovFile}.
+     * @param targetFactorPercentileThreshold  The percentile to of extreme target factor values to cut.
```
davidbenjamin: syntax

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+     * @param extremeColumnMedianCountPercentileThreshold Percentile to cut columns that after they are ranked by median value
```
davidbenjamin: syntax

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+     * @param countTruncatePercentile percentile (on either end) to truncate extreme values (i.e. extreme high or low values will be set to the high or low count truncation value, which is based on this percentlie)
+     * @param maximumPercentageZeroTargets  the maximum percentage zero values in a target (across samples) before the target is filtered.
```
davidbenjamin: percentage of

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+     * @param maximumPercentageZeroColumns  the maximum percentage zero values in a sample (across targets) before the sample is filtered.
```
davidbenjamin: percentage of

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+
+        // Normalize read-counts by removing the target factor component:
+        normalizeReadCountsByTargetFactors(readCounts, targetFactors);
+
+        // Create and write the first part of the PoN HDF5 file.
+        if (!isDryRun) {
+            writeTargetFactorNormalizeReadCountsAndTargetFactors(outputHDF5Filename, readCounts, targetFactors, initialTargets.targets());
+        }
+
+        // Remove column and targets with too many zeros and targets with
```
davidbenjamin: Comment fits on one line.

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+
+        // Impute zero counts to be same as the median for the same target.
+        // This happens in-place.
+        imputeZerosCounts(readCounts, logger);
+
+        // Truncate extreme count values.
+        // This happens in-place.
+        truncateExtremeCounts(readCounts, countTruncatePercentile, logger);
+
+        // Normalized by the median and log scale the read counts.
```
davidbenjamin: extra 'd'

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+        final RealMatrix logNormalsEigenV = logNormalsV.getSubMatrix(0, numberOfCountColumns - 1, 0, numberOfEigenSamples - 1);
+
+        final RealMatrix reducedCounts = logNormalCounts.multiply(logNormalsEigenV);
+
+        // The Pseudo inverse comes nearly for free from having run the SVD decomposition.
+        final RealMatrix logNormalsPseudoInverse = logNormalsSVD.getPinv();
+
+        logger.info("Calculating the reduced PoN inverse matrix...");
+        final long riStartTime = System.currentTimeMillis();
+        final RealMatrix reducedCountsInverse = SVDFactory.createSVD(reducedCounts, ctx).getPinv();
```
davidbenjamin: Rename to `reducedCountsPseudoInverse`

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+        final SVD logNormalsSVD = SVDFactory.createSVD(logNormals.counts(), ctx);
+        final long svdEndTime = System.currentTimeMillis();
+        logger.info(String.format("Finished the SVD decomposition of the log-normal counts. Elapse of %d seconds", (svdEndTime - svdStartTime) / 1000));
+
+        final int numberOfEigenSamples = determineNumberOfEigenSamples(requestedNumberOfEigenSamples, numberOfCountColumns, logNormalsSVD, logger);
+        logger.info(String.format("Including %d eigen samples in the reduced PoN", numberOfEigenSamples));
+        final RealMatrix logNormalsV = logNormalsSVD.getV();
+        final RealMatrix logNormalsEigenV = logNormalsV.getSubMatrix(0, numberOfCountColumns - 1, 0, numberOfEigenSamples - 1);
+
+        final RealMatrix reducedCounts = logNormalCounts.multiply(logNormalsEigenV);
```
davidbenjamin: This is slower than it needs to be.  The SVD of matrix M is M = U*sigma*V' (using MATLAB notation for transpose) and we are calculating `reducedCounts` as M*V, where V is restricted to E eigensamples.  By orthogonality of V, V'V is the ExE identity, with zeroes below.  Thus sigma*V'*V equals sigma truncated to the first E columns.  Since sigma is rectangular diagonal, the result U*sigma*V'*V is thus a matrix whose e-th column is the e-th column of U times the e-th diagonal entry of sigma.

davidbenjamin: The following code should produce an equivalent result.  The code itself is longer, but it corresponds to the mathematically clearer process of throwing out small singular values.
```
final double[] singularValues = SVD.getSingularValues();
final double U = SVD.getU().getSubMatrix(0, logNormalCounts.getRowDimension()-1, 0, numberOfEigenSamples - 1).copy();
final RealMatrix reducedCounts = U.walkInOptimizedOrder(new DefaultRealMatrixChangingVisitor() {
                @Override
                public double visit(final int row, final int column, final double value) {
                    return singularValues[column]*value;
                }
            });

Comment (davidbenjamin) [inline src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java]: To elaborate, here's the interpretation: the left eigenvectors, i.e. the columns of U, are scaled eigensamples (U is the only thing with T rows, so it's the only thing that could conceivably represent samples).  They're scaled in order to conform with the definition of SVD, which required that U and V be orthogonal.  The scale factors are the diagonal elements of sigma.  Thus we drop all but the most significant eigensamples of U and scale them.

File: src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java [hunk 25]
```diff
@@ -0,0 +1,712 @@
...
+        final int numberOfEigenSamples;
+        if (requestedNumberOfEigenSamples.isPresent()) {
+            if (requestedNumberOfEigenSamples.getAsInt() > numberOfCountColumns) {
+                logger.warn(String.format("The number of requested eigen samples (%d) is larger than the available number of read count columns after filtering (%d), thus we will have to use the latter.", requestedNumberOfEigenSamples.getAsInt(), numberOfCountColumns));
+            }
+            numberOfEigenSamples = Math.min(requestedNumberOfEigenSamples.getAsInt(), numberOfCountColumns);
+        } else {
+            final double[] singularValues = logNormalsSVD.getSingularValues();
+            final double mean = MathUtils.mean(singularValues, 0, singularValues.length);
+            final double eigenValueCutoff = mean * JOLLIFES_RULE_MEAN_FACTOR; // Jollife's less strict version of Kaiser' Rule.
```
davidbenjamin: Kaiser's

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+            public double visit(final int row, final int column, final double value) {
+                return value - medianOfMedians;
+            }
+        });
+        logger.info(String.format("Counts centered around the BGS center %.2f", medianOfMedians));
+        return medianOfMedians;
+    }
+
+    /**
+     * Final pre-panel normalization that consists on dividing all counts by the median of
```
davidbenjamin: consists of

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+     */
+    @VisibleForTesting
+    static void normalizeAndLogReadCounts(final ReadCountCollection readCounts, final Logger logger) {
+        final RealMatrix counts = readCounts.counts();
+        final int columnCount = counts.getColumnDimension();
+        final Median medianCalculator = new Median();
+        for (int i = 0; i < columnCount; i++) {
+            final double[] colCounts = counts.getColumn(i);
+            final double median = medianCalculator.evaluate(colCounts);
+            final double invMedian = 1.0 / median;
```
davidbenjamin: Caching an inverse in order to replace division with multiplication is premature optimization IMO.  Even with a million targets and a thousand samples, this saves at most a few seconds.

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+            for (int j = 0; j < colCounts.length; j++) {
```
davidbenjamin: This could be handled with a visitor:
```
final double[] medians = IntStream.range(0, counts.getColumnDimension()).map(col -> medianCalculator.evaluate(counts.getColumn(i));
counts.walkInOptimizedOrder(new DefaultRealMatrixChangingVisitor() {
            @Override
            public double visit(final int row, final int column, final double value) { return value / medians[column]; }
        });
```

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+                colCounts[j] = Math.log(Math.max(EPSILON, colCounts[j] * invMedian)) * INV_LN_2;
+            }
+            counts.setColumn(i, colCounts);
+        }
+        logger.info("Counts normalized by the column median and changed into bits.");
```
davidbenjamin: "Changed into bits" is a nifty way of saying log-2 transformed, but it's too clever.

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+
+    /**
+     * Impute zero counts to the median of non-zero values in the enclosing target row.
+     *
+     * <p>The imputation is done in-place, thus the input matrix is well be modified as a result of this call.</p>
+     *
+     * @param readCounts the input and output read-count matrix.
+     */
+    @VisibleForTesting
+    static void imputeZerosCounts(final ReadCountCollection readCounts, final Logger logger) {
```
davidbenjamin: Method name is ambiguous; not clear if you are imputing new values to zeros or imputing old values to zero.  I prefer to err on the side of clarity with eg `imputeZeroCountsAsTargetMedians`.

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+            int nonZeroCount = 0;
+            for (int j = 0; j <  rowCounts.length; j++) {
+                if (rowCounts[j] != 0.0) {
+                    nonZeroValues[nonZeroCount++] = rowCounts[j];
+                } else {
+                    zeroIndexes[zeroCount++] = j;
+                }
+            }
+            totalCounts += rowCounts.length;
+            if (zeroCount > 0) { // We only have to do anything if if there is any zeros in the row.
```
davidbenjamin: if if

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+        final double[] nonZeroValues = new double[columnCount];
+        final Median medianCalculator = new Median();
+        int totalCounts = 0;
+        int totalZeroCounts = 0;
+
+        // for each row aka target:
+        for (int i = 0; i < targetCount; i++) {
+            // We classify columns as having a zero or not.
+            // we accumulate nonZero values in nonZeroValues to calculate the median
+            // where as we accumulate zero indexes that need to be updated later.
```
davidbenjamin: whereas is one word

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
davidbenjamin: are any zeros

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+            if (zeroCount > 0) { // We only have to do anything if if there is any zeros in the row.
+                final double nonZeroMedian = medianCalculator.evaluate(nonZeroValues, 0, nonZeroCount);
+                for (int j = 0; j < zeroCount; j++) {
+                    counts.setEntry(i, zeroIndexes[j], nonZeroMedian);
+                }
+                totalZeroCounts += zeroCount;
+            }
+        }
+        if (totalZeroCounts > 0) {
+            logger.info(String.format("Some 0.0 counts, %d of %d (%.2f%%), where imputed to their enclosing target non-zero median", totalZeroCounts, totalZeroCounts, 100.0 * (totalZeroCounts / totalCounts)));
```
davidbenjamin: were

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+     * @param readCounts the input and output read-count matrix.
+     */
+    @VisibleForTesting
+    static void imputeZerosCounts(final ReadCountCollection readCounts, final Logger logger) {
+
+        final RealMatrix counts = readCounts.counts();
+        final int targetCount = counts.getRowDimension();
+        final int columnCount = counts.getColumnDimension();
+
+        // Declared once outside loop to reuse across rows:
```
davidbenjamin: IMO there's premature optimization going on.  At the cost of doing two passes instead of one, we could replace this with:
```
final double[] medians = IntStream(0, targetCount).map(t -> counts.getRow(t).stream().filter(c -> c != 0.0).median());
counts.walkInOptimizedOrder(new DefaultRealMatrixChangingVisitor() {
            @Override
            public double visit(final int row, final int column, final double value) { return value != 0 ? value : medians[row]; }
        });
```

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+            logger.info("No count is 0.0 thus no count needed to be imputed.");
+        }
+    }
+    /**
+     * Truncates the extreme count values in the input read-count collection.
+     * Values are forced to be bound by the percentile indicated with the input {@code percentile} which must be
+     * in the range [0 .. 50.0]. Values under that percentile and the complementary (1 - percentile) are set to the
+     * corresponding threshold value.
+     *
+     * <p>The imputation is done in-place, thus the input matrix is well be modified as a result of this call.</p>
```
davidbenjamin: is well --> is

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+     *
+     * @param readCounts the input and output read-count matrix.
+     */
+    @VisibleForTesting
+    static void truncateExtremeCounts(final ReadCountCollection readCounts, final double percentile, final Logger logger) {
+
+        final RealMatrix counts = readCounts.counts();
+        final int targetCount = counts.getRowDimension();
+        final int columnCount = counts.getColumnDimension();
+        final double[] values = new double[targetCount * columnCount];
```
davidbenjamin: How about something like `double[] values = IntStream.range(0, targetCount).flatMap(t -> counts.getRow(t))`?  (I'm sure I'm forgetting `boxed()` etc. but it still ought to be a one-liner).

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+        }
+        final Percentile bottomPercentileEvaluator = new Percentile(percentile);
+        final Percentile topPercentileEvaluator = new Percentile(100.0 - percentile);
+        final double bottomPercentileThreshold = bottomPercentileEvaluator.evaluate(values);
+        final double topPercentileThreshold = topPercentileEvaluator.evaluate(values);
+        long totalCounts = 0;
+        long bottomTruncatedCounts = 0;
+        long topTruncatedCounts = 0;
+
+        for (int i = 0; i < targetCount; i++) {
```
davidbenjamin: Use a `Visitor`.

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+    }
+
+    /**
+     * Creates a new read-count collection that is a copy of the input but
+     * dropping columns with extreme median coverage.
+     *
+     * @param readCounts the input read-counts.
+     * @param extremeColumnMedianCountPercentileThreshold bottom percentile to use as a exclusion threshold.
+     * @return never {@code null}. It might be a reference to the input read-counts if
+     * there is no columns to be dropped.
```
davidbenjamin: are

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+            logger.info(String.format("None of the %d counts where truncated as they all fall in the non-extreme range [%.2f, %.2f]", totalCounts, bottomPercentileThreshold, topPercentileThreshold));
+        } else {
+            final double truncatedPercentage = ((double)(topTruncatedCounts + bottomTruncatedCounts) / totalCounts) * 100;
+            logger.info(String.format("Some counts, %d out of %d (%.2f%%), where truncated as they fall out of the non-extreme range [%.2f, %.2f]",
+                    topTruncatedCounts + bottomTruncatedCounts, totalCounts, truncatedPercentage, bottomPercentileThreshold, topPercentileThreshold));
+        }
+    }
+
+    /**
+     * Creates a new read-count collection that is a copy of the input but
```
davidbenjamin: one line

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+     * dropping columns with extreme median coverage.
+     *
+     * @param readCounts the input read-counts.
+     * @param extremeColumnMedianCountPercentileThreshold bottom percentile to use as a exclusion threshold.
```
davidbenjamin: an exclusion

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+     * @return never {@code null}. It might be a reference to the input read-counts if
+     * there is no columns to be dropped.
+     */
+    @VisibleForTesting
+    static ReadCountCollection removeColumnsWithExtremeMedianCounts(final ReadCountCollection readCounts, final double extremeColumnMedianCountPercentileThreshold, final Logger logger) {
+        final List<String> columnNames = readCounts.columnNames();
+        final double[] columnMedians = new double[columnNames.size()];
+        final RealMatrix counts = readCounts.counts();
+        final Median medianCalculator = new Median();
```
davidbenjamin: `final double[] columnMedians = IntStream.range(0, counts.getColumnDimension()).map(col -> medianCalculator.evaluate(counts.getColumn(col)))`.

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+     * @param readCounts the input read counts.
+     * @param maximumTargetZeros maximum number of counts equal to 0. per target tolerated.
+     * @return never {@code null}. It might be a reference to the input read-counts if there is
+     *   is no target to be dropped.
+     */
+    @VisibleForTesting
+    static ReadCountCollection removeTargetsWithTooManyZeros(final ReadCountCollection readCounts, final int maximumTargetZeros, final Logger logger) {
+        final Set<Target> targetsToKeep = new HashSet<>(readCounts.targets().size());
+        final RealMatrix counts = readCounts.counts();
+        for (int i = 0; i < counts.getRowDimension(); i++) {
```
davidbenjamin: `
Predicate<double[]> keepRow = rowCounts -> rowCounts.stream().filter(x -> x == 0.0).count() <= maximumTargetZeros();
final Set<Target> targetsToKeep IntStream.range(0, counts.getRowDimension()).filter(row -> keepRow(counts.getRow(row))).map(readCounts.targets()::get).collect(Collectors.toSet());
```

File: src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java [hunk 43]
```diff
@@ -0,0 +1,712 @@
...
+     *     It will return a copy of the input read-count collection with such columns dropped.
+     * </p>
+     *
+     * @param readCounts the input read counts.
+     * @param maximumColumnZeros maximum number of counts equal to 0. per column tolerated.
+     * @return never {@code null}. It might be a reference to the input read-counts if there is
+     *   is no column to be dropped.
+     */
+    @VisibleForTesting
+    static ReadCountCollection removeColumnsWithTooManyZeros(final ReadCountCollection readCounts, final int maximumColumnZeros, final Logger logger) {
```
davidbenjamin: As with targets, Stream-ify this.

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+     * @param readCounts read-counts to normalize.
+     * @param targetFactors the target factors
+     */
+    @VisibleForTesting
+    static void normalizeReadCountsByTargetFactors(final ReadCountCollection readCounts, final double[] targetFactors) {
+        final RealMatrix counts = readCounts.counts();
+        final int targetCount = counts.getRowDimension();
+        final int columnCount = counts.getColumnDimension();
+        for (int i = 0; i < targetCount; i++) {
+            final double factorInverse = 1.0 / targetFactors[i];
```
davidbenjamin: Premature optimization.

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+     * </p>
+     * @param readCounts read-counts to normalize.
+     * @param targetFactors the target factors
+     */
+    @VisibleForTesting
+    static void normalizeReadCountsByTargetFactors(final ReadCountCollection readCounts, final double[] targetFactors) {
+        final RealMatrix counts = readCounts.counts();
+        final int targetCount = counts.getRowDimension();
+        final int columnCount = counts.getColumnDimension();
+        for (int i = 0; i < targetCount; i++) {
```
davidbenjamin: Use a `Visitor`.

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+    }
+
+    /**
+     * Calculate the target factors:
+     *
+     * <p>These are the median coverage per target</p>.
+     * @param readCounts the input read counts.
+     * @return never {@code null}, with as many elements as targets in {@code readCounts}.
+     */
+    private static double[] calculateTargetFactors(final ReadCountCollection readCounts) {
```
davidbenjamin: Since it comes up so much, you should extract methods for getting arrays of row medians and column medians from a matrix.

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+        final RealMatrix counts = readCounts.counts();
+        final Median medianCalculator = new Median();
+        for (int i = 0; i < result.length; i++) {
+            result[i] = medianCalculator.evaluate(counts.getRow(i));
+        }
+        return result;
+    }
+
+    /**
+     * Calculates the absolute number of zero tolerated given the user requested percentage
```
davidbenjamin: zeros

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
```diff
@@ -0,0 +1,712 @@
...
+    /**
+     * Calculate the variance for each row in a matrix.
+     *
+     * @param m Not {@code null}
+     * @return Not {@code null}.  double array with length {@code m.getRowDimension()}
+     */
+    static double[] calculateRowVariances(final RealMatrix m) {
+        Utils.nonNull(m, "Cannot calculate variances for a null matrix.");
+        final StandardDeviation std = new StandardDeviation();
+        final double[] result = new double[m.getRowDimension()];
```
davidbenjamin: Use a stream.

src/main/java/org/broadinstitute/hellbender/utils/hdf5/PoN.java
```diff
@@ -56,7 +73,7 @@
      * Log normal selected samples listed by their numerical index in the log-normal matrices.
      *
      * <p>
-     * The return matrix is a modifiable detached copy of the values in the
+     * The return matrix is a unmodifiable detached copy of the values in the
```
davidbenjamin: an

src/main/java/org/broadinstitute/hellbender/utils/hdf5/PoN.java
```diff
@@ -73,7 +90,7 @@
      * </p>
      *
      * <p>
-     * The return matrix is a modifiable detached copy of the values in the
+     * The return matrix is a unmodifiable detached copy of the values in the
```
davidbenjamin: an

src/main/java/org/broadinstitute/hellbender/utils/param/ParamUtils.java
```diff
@@ -81,21 +93,50 @@ public static long isPositiveOrZero(final long val, final String message) {
...
+    }
+
+    /**
+     * Checks that the  input is greater than zero and returns the same value or throws an {@link IllegalArgumentException}
+     * @param val value to check
+     * @param message the text message that would be pass to the exception thrown
+     * @return the same value
+     * @throws IllegalArgumentException
+     */
+    public static long isGreaterThanZero(final long val, final String message) {
```
davidbenjamin: Why not `isPositive`?

src/main/java/org/broadinstitute/hellbender/utils/param/ParamUtils.java
```diff
@@ -128,4 +169,58 @@ public static double logb(final double val, final double base) {
...
         return Math.log(val)/Math.log(2.0);
     }
+
+    /**
+     *  Writes a list of strings to the given filename.  The file will be overwritten if it exists.
+     *
+     * @param stringList Strings that will be written (one to a line)
+     * @param outputFile File to write
+     */
+    public static void writeStringListToFile(final List<String> stringList, final File outputFile) {
```
davidbenjamin: How about call this `writeStringsToFile` and take any `Iterable<String>` as input?

davidbenjamin:
@LeeTL1220 Review complete.

src/main/java/org/broadinstitute/hellbender/tools/exome/CreatePanelOfNormals.java
LeeTL1220: I agree, but requires an implementation of issue #265 , see comments there.

src/main/java/org/broadinstitute/hellbender/tools/exome/CreatePanelOfNormals.java
LeeTL1220: That was a leftover from previous code.  I'll change...

docs/PoN/testdataGeneration/generate_normal_projections.m
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/tools/exome/Case2PoNTargetMapper.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/tools/exome/Case2PoNTargetMapper.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/tools/exome/CreatePanelOfNormals.java
LeeTL1220: calculate has become "check" in both cases.

Done.

I agree about terminology, but this a leftover from when we wouldn't collapse read counts (which is still supported) -- no action.  We can file an issue if you are not satisfied.

src/main/java/org/broadinstitute/hellbender/tools/exome/CreatePanelOfNormals.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/tools/exome/CreatePanelOfNormals.java
LeeTL1220: Already done.

src/main/java/org/broadinstitute/hellbender/tools/exome/NormalizeSomaticReadCounts.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/tools/exome/PerformSegmentation.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/tools/exome/PerformSegmentation.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/tools/exome/PoNIssueDetector.java
LeeTL1220: Done... simplifies the code.

src/main/java/org/broadinstitute/hellbender/tools/exome/ReductionResult.java
LeeTL1220: I changed it to be a bit better, but not exactly what you wrote here.

Done

src/main/java/org/broadinstitute/hellbender/tools/exome/ReductionResult.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/tools/exome/TangentNormalizer.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/tools/exome/TangentNormalizer.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/tools/exome/TangentNormalizer.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: As a warning, I have already filed an issue to sparkify this piece of code, since it is one of the biggest bottlenecks in ``CreatePanelOfNormals``

That being said, I have put your code into this pull request.  Many many thanks for running the test already.

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Nope, Jollife...

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: The optimizer might see it regardless of where it is.

No action.

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Done, but modifications to this code were necessary.

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Had to make minor changes to preserve the log message.

LeeTL1220: And to fix syntax.

LeeTL1220: The code is a bit simpler since there are several small loops being done sequentially.  

Log message is preserved.

Done

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: These are not necessary changes since I have put the visit function for the RealMatrix.  Please confirm.

No action.

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
davidbenjamin: I meant in the comment -- "Kaiser's" was missing the 's'.

src/main/java/org/broadinstitute/hellbender/tools/exome/CreatePanelOfNormals.java
davidbenjamin: OK!

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
davidbenjamin: Confirmed.

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: One liner using guava

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: I'm a little worried about using visitors, since there could be a performance hit when preserving the log messages.  I did the conversion above, but have to add a second pass through the data just to calculate the number of values that will be imputed. 

Here I would have to do two additional passthroughs.  This may not be a problem, because the optimizer can probably figure out what is going on.  But if it can't this would be a performance hit.

Tell me if you disagree.

No action.

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Not sure it is done the best way, but ...

Done

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: IMO this code does not really simplify anything.

No action.

davidbenjamin: Fair enough.

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
davidbenjamin: I'm not really worried about performance since visiting each value of a matrix is much cheaper than matrix multiplications and SVDs.  I just want whatever is most readable, and if *two* additional passes are needed, then it's best to keep the code as-is.

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Never mind. 

Going to simplify, but possibly not the same way.

LeeTL1220: Hard using a predicate.  Just created a simple method and put that into a ``.filter(...)``
Done.

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Very similar to target filtering.
Done.

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Used a visitor

Done

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
davidbenjamin: For sure!  My suggested code was ugly.

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/hdf5/HDF5PoNCreator.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/hdf5/PoN.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/hdf5/PoN.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/utils/param/ParamUtils.java
LeeTL1220: Renamed

Done

src/main/java/org/broadinstitute/hellbender/utils/param/ParamUtils.java
LeeTL1220: Done

LeeTL1220:
@davidbenjamin Need to rebase before giving it back to you, since other branches were merged.  Probably done tonight

LeeTL1220:
@davidbenjamin Rebased.... ready for (hopefully) final review....  All unit tests should pass.  If not, please re-assign to me

Added a MatrixUtils class for those row/column calculations that are common.

PR #291, opened 2016-01-12
Add binsize as a parameter for SparkGenomeReadCounts

achevali:


LeeTL1220:
@achevali Have you tried running this on the command line with different values?  Can you just add a quick integration test that makes sure you get a different set of targets with a different binsize?

LeeTL1220:
@achevali I see no difference.... Did you address my comment?

achevali:
Added test for larger binsize

achevali:
Closes #290

PR #292, opened 2016-01-14
Added comment about target padding.

### LINKED ISSUE #206: Document the padding of targets in the README
Since the padding of targets seems to improve performance, we should add it to the documentation.

- [x] Add PadTargets usage to the README
- [x] Send along the updated Queue scripts to @vdauwera  (when @achevali has completed pull request reviews in recapseg-hb-eval repo and Issue #199 in this repo).  The Queue scripts need to be posted to the GATK website.
- [x] Bonus:  Update the necessary Broad VM dotkit ``use`` statements to include Java-1.8 and the .r-3.1.3-gatk only

LeeTL1220:
Edited  comment about target padding.

@vdauwera Can you review this simple update to the workflow prerequisites in the README?  Does it need to be included in later material?  At some point, contents of this README will be in the forum (broken up into separate posts).  However, I am having an issue with the forum now (see email).

Closes #206

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,8 +191,9 @@ The CNV case and PoN workflows (description and examples)
 1       35138   35174   target4
 ....snip....
 ```
-Before running the workflows, we recommend padding the target file by 250 bases with the ``PadTargets`` tool.  Example:  ``java -jar gatk-protected.jar --targets initial_target_file.bed --output initial_target_file.padded.bed --padding 250``
-The premade Queue scripts (see below) handle the target padding automatically. 
+Before running the workflows, we recommend padding the target file by 250 bases with the ``PadTargets`` tool.  Example:  ``java -jar gatk-protected.jar PadTargets --targets initial_target_file.bed --output initial_target_file.padded.bed --padding 250``
+This allows some off-target reads to be factored into the copy ratio estimates.  Our internal evaluations have shown that this improves results.  
+If you are using the premade Queue scripts (see below), you can specify the padding there and the workflow will generate the padded targets automatically (i.e. there is no reason to run PadTargets explicitly if you are using the premade Queue scripts).  
 
 7. PoN file (when running case samples only).  This file should be created using the Create PoN workflow (see below).
 <a name="SampleName"/>
--- a/README.md
+++ b/README.md
@@ -191,6 +191,9 @@ The CNV case and PoN workflows (description and examples)
 1       35138   35174   target4
 ....snip....
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -191,6 +191,9 @@ The CNV case and PoN workflows (description and examples)
 1       35138   35174   target4
 ....snip....
 ```
+Before running the workflows, we recommend padding the target file by 250 bases with the ``PadTargets`` tool.  Example:  ``java -jar gatk-protected.jar --targets initial_target_file.bed --output initial_target_file.padded.bed --padding 250``
```
vdauwera: This is a bit confusing -- you say to use the PadTargets tool but that name doesn't seem to figure in your example command.

vdauwera: Would be nice to say why padding is recommended, or what goes into determining the size should be 250.

README.md
```diff
@@ -191,6 +191,9 @@ The CNV case and PoN workflows (description and examples)
...
+The premade Queue scripts (see below) handle the target padding automatically.
```
vdauwera: What does this mean? That if you're using the Queue scripts you don't need to run the step mentioned above? If so that should be more explicit.

vdauwera:
Done with my review -- would be good to have a bit more clarity, see in line comments.

README.md
LeeTL1220: Done (both comments)

README.md
LeeTL1220: Done

LeeTL1220:
@vdauwera This pull request can be closed without merging.  Previous pull request obviates it.  I updated the forum post, so everything should be fine.

LeeTL1220:
Here was the added content...

Before running the workflows, we recommend padding the target file by 250 bases with the ``PadTargets`` tool.  Example:  ``java -jar gatk-protected.jar PadTargets --targets initial_target_file.bed --output initial_target_file.padded.bed --padding 250``
This allows some off-target reads to be factored into the copy ratio estimates.  Our internal evaluations have shown that this improves results.  
If you are using the premade Queue scripts (see below), you can specify the padding there and the workflow will generate the padded targets automatically (i.e. there is no reason to run PadTargets explicitly if you are using the premade Queue scripts).

LeeTL1220:
Forum post:  http://gatkforums.broadinstitute.org/gatk/discussion/6791/description-and-examples-of-the-steps-in-the-cnv-case-and-cnv-pon-creation-workflows

Issue #146, opened 2015-10-14
HMM based CNV discovery.


vruano:
Uses the HMM Copy number Tri-State to discover CNV segments.

davidbenjamin:
Closed by PR #313.

PR #313, opened 2016-01-20
Germline HMM based CNV segment discovery

vruano:
Addresses issue #146.


src/main/java/org/broadinstitute/hellbender/tools/exome/CopyNumberTriStateSegment.java
```diff
@@ -0,0 +1,125 @@
...
+
+    private final double eventQuality;
+
+    private final double stdev;
+
+    /**
+     * Returns the "exact quality" (EQ)
+     * @return any {@code double} value.
+     */
+    public double getExactQuality() {
```
davidbenjamin: Docs need to describe what all these qualities mean.  i.e.  exact quality = phred-scaled probability that this segment is exactly correct.

src/main/java/org/broadinstitute/hellbender/tools/exome/CopyNumberTriStateSegment.java
```diff
@@ -0,0 +1,125 @@
...
+        return exactQuality;
+    }
+
+    /**
+     * Returns the "some quality" (SQ)
+     * @return any {@code double} value.
+     */
+    public double getSomeQuality() {
```
davidbenjamin: docs: phred-scaled probability that some copy-number event of the same type overlaps this segment.

src/main/java/org/broadinstitute/hellbender/tools/exome/CopyNumberTriStateSegment.java
```diff
@@ -0,0 +1,125 @@
...
+        return someQuality;
+    }
+
+    /**
+     * Returns the "start quality" (LQ)
+     * @return any {@code double} value.
+     */
+    public double getStartQuality() {
```
davidbenjamin: docs

src/main/java/org/broadinstitute/hellbender/tools/exome/CopyNumberTriStateSegment.java
```diff
@@ -0,0 +1,125 @@
...
+        return startQuality;
+    }
+
+    /**
+     * Returns the "stop quality" (RQ)
+     * @return any {@code double} value.
+     */
+    public double getEndQuality() {
```
davidbenjamin: docs

src/main/java/org/broadinstitute/hellbender/tools/exome/CopyNumberTriStateSegment.java
```diff
@@ -0,0 +1,125 @@
...
+        return endQuality;
+    }
+
+    /**
+     * Returns the "event quality" (NDP).
+     * @return any {@code double} value.
+     */
+    public double getEventQuality() {
```
davidbenjamin: docs

vruano:
Sorry, I've just improved the test coverage a bit; please go ahead with the review. Won't touch it again until you are finished with it @davidbenjamin 

By the way 178,000 of the new lines are from a test target file. So don't worry about the size of the commit.

src/main/java/org/broadinstitute/hellbender/tools/exome/DiscoverCopyNumberTriStateSegments.java
```diff
@@ -0,0 +1,484 @@
...
+
+    public static final String INPUT_ARE_ZSCORES_FULL_NAME = "inputAreZscores";
+
+    public static final String INPUT_ARE_ZSCORES_SHORT_NAME = "zscores";
+
+    /**
+     * Maximum reportable output quality score; higher quality scores will
+     * capped to this value.
+     */
+    public static final double MAX_QUAL_SCORE = 199.9999;
```
davidbenjamin: Any reason for this particular value?

src/main/java/org/broadinstitute/hellbender/tools/exome/DiscoverCopyNumberTriStateSegments.java
```diff
@@ -0,0 +1,484 @@
...
+import org.broadinstitute.hellbender.utils.hmm.ViterbiAlgorithm;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.DoubleStream;
+
+/**
+ * Discover potential copy-number segments in input read-counts as the most likely segment sequences
```
davidbenjamin: Javadoc should indicate what step in workflow immediately precedes this (tangent normalization?)

src/main/java/org/broadinstitute/hellbender/tools/exome/DiscoverCopyNumberTriStateSegments.java
```diff
@@ -0,0 +1,484 @@
...
+    protected TargetArgumentCollection targetArguments = new TargetArgumentCollection(() -> inputFile);
+
+    @Override
+    protected Object doWork() {
+        final CopyNumberTriStateHiddenMarkovModel model = modelArguments.createModel();
+        final TargetCollection<Target> targets = targetArguments.readTargetCollection(false);
+        final ReadCountCollection inputCounts = readAndSortCountsByTargetCoordinates(targets);
+        checkForMissingTargetsInInputCounts(targets, inputCounts);
+
+        final Map<String, List<CopyNumberTriStateSegment>> allSegments =
```
davidbenjamin: variable name should indicate that the `String` keys are samples.

src/main/java/org/broadinstitute/hellbender/tools/exome/DiscoverCopyNumberTriStateSegments.java
```diff
@@ -0,0 +1,484 @@
...
+            while (allSegmentsSorted.hasNext()) {
+                writer.writeRecord(allSegmentsSorted.next());
+            }
+        } catch (final IOException ex) {
+            throw new UserException.CouldNotCreateOutputFile(outputFile, "problems writing into the output", ex);
+        }
+    }
+
+    /**
+     * Create a merging iterator that sort all the segments across samples based
```
davidbenjamin: sorts

davidbenjamin: "by their coordinates" sounds a bit better than "based on their coordinates"

src/main/java/org/broadinstitute/hellbender/tools/exome/DiscoverCopyNumberTriStateSegments.java
```diff
@@ -0,0 +1,484 @@
...
+     * on their coordinates.
+     * <p>
+     *     The order between contigs is determined by the input target collection.
+     * </p>
+     * <p>
+     *     The input map of sample name to segment list is assume to contain all
```
davidbenjamin: assumed

src/main/java/org/broadinstitute/hellbender/tools/exome/DiscoverCopyNumberTriStateSegments.java
```diff
@@ -0,0 +1,484 @@
...
+     *     segments within each value list already in the correct order.
+     * </p>
+     *
+     * @param targets the target collection.
+     * @param allSegments map with all segment list coming from different samples.
```
davidbenjamin: lists

src/main/java/org/broadinstitute/hellbender/tools/exome/DiscoverCopyNumberTriStateSegments.java
```diff
@@ -0,0 +1,484 @@
...
+                return startCmp;
+            }
+            return Integer.compare(s1.getEnd(), s2.getEnd());
+        };
+
+        return Iterators.mergeSorted(allSegmentRecordIterators, recordComparator);
+    }
+
+    private Map<String, List<CopyNumberTriStateSegment>> calculateBestPathSegments(CopyNumberTriStateHiddenMarkovModel model, ReadCountCollection inputCounts) {
+        final List<String> columnNames = inputCounts.columnNames();
```
davidbenjamin: `sampleNames` -- unless there is a case where columns don't represent samples.

src/main/java/org/broadinstitute/hellbender/tools/exome/DiscoverCopyNumberTriStateSegments.java
```diff
@@ -0,0 +1,484 @@
...
+        final Map<String, List<CopyNumberTriStateSegment>> allSegments = new LinkedHashMap<>(columnNames.size());
+        long recordCount = 0;
+        final List<Target> targets = inputCounts.targets();
+        for (int i = 0; i < columnNames.size(); i++) {
+            final String columnName = columnNames.get(i);
```
davidbenjamin: `sampleName` -- unless as above

src/main/java/org/broadinstitute/hellbender/tools/exome/DiscoverCopyNumberTriStateSegments.java
```diff
@@ -0,0 +1,484 @@
...
+     * Calculates the list of copy-number change segments as the one present in the
+     * most likely segment sequence given a {@link CopyNumberTriStateHiddenMarkovModel model}.
+     * @param model the HMM model to be used to discover copy-number change segments.
+     * @param targets the input targets sequence.
+     * @param inputValues the input coverage values.
+     * @return never {@code null}.
+     */
+    private List<CopyNumberTriStateSegment> getCopyNumberTriStateSegments(final CopyNumberTriStateHiddenMarkovModel model, final List<Target> targets, double[] inputValues) {
+        final double mean = MathUtils.mean(inputValues, 0, inputValues.length);
+        final double invStdev = 1.0 / MathUtils.stddev(inputValues, 0, inputValues.length);
```
davidbenjamin: No need to cache the inverse.  It saves one division per target, but even with 10^6 targets and 10^3 samples we're talking about an optimization that saves less than a second but makes the code harder to read.

src/main/java/org/broadinstitute/hellbender/tools/exome/DiscoverCopyNumberTriStateSegments.java
```diff
@@ -0,0 +1,484 @@
...
+     * @return never {@code null}.
+     */
+    private CopyNumberTriStateSegment composeSegment(final List<Target> targets,
+                                                     final CopyNumberTriState call,
+                                                     final Target precedingTarget, final Target followingTarget,
+                                                     final ForwardBackwardAlgorithm.Result<Double, Target, CopyNumberTriState> fbResult,
+                                                     final List<Double> inputValues) {
+        final int length = targets.size();
+        final Target firstTarget = targets.get(0);
+        final double mean = inputValues.stream().mapToDouble(d -> d).sum() / length;
```
davidbenjamin: `.average().getAsDouble()` has an annoying unboxing from the `OptionalDouble` but IMO it's still preferable because it conveys the point of the calculation more directly.

src/main/java/org/broadinstitute/hellbender/tools/exome/DiscoverCopyNumberTriStateSegments.java
```diff
@@ -0,0 +1,484 @@
...
+        final double stdev = inputValues.size() <= 1 ? 0 : Math.sqrt(inputValues.stream().mapToDouble(d -> (d - mean) * (d - mean)).sum() / (length - 1));
```
davidbenjamin: Use Apache Commons, which also does sample std dev.

src/main/java/org/broadinstitute/hellbender/tools/exome/DiscoverCopyNumberTriStateSegments.java
```diff
@@ -0,0 +1,484 @@
...
+     * are considered to indicate a log probability calculation problem that is
+     * worth to be reported as an error or warning.
+     */
+    private static final double MAX_LOG_PROB = 1e-3;
+
+    /**
+     * Transform a log scaled probability (x) into the Phred scaled
+     * equivalent or its complement (1-x) Phred scaled equivalent.
+     * <p>
+     *     This method is tolerant with probability slightly larger than 1.0
```
davidbenjamin: tolerates probabilities

src/main/java/org/broadinstitute/hellbender/tools/exome/DiscoverCopyNumberTriStateSegments.java
```diff
@@ -0,0 +1,484 @@
...
+            final EnumSet<CopyNumberTriState> allStatesExceptCall = EnumSet.complementOf(onlyCallState);
+            return fbResult.logConstrainedProbability(lastTarget, Arrays.asList(onlyCallState, allStatesExceptCall));
+        }
+    }
+
+    /**
+     * Calculates the probability of a hidden state switch at the beginning of the segment
+     * from a different state to the call state.
+     */
+    private double logStartProbability(final Target precedingTarget, final Target firstTarget,
```
davidbenjamin: This should come before `logEndProbability`

src/main/java/org/broadinstitute/hellbender/tools/exome/DiscoverCopyNumberTriStateSegments.java
```diff
@@ -0,0 +1,484 @@
...
+        if (length == 0 || call == CopyNumberTriState.NEUTRAL) {
+            return 0;
+        } else {
+            final Set<CopyNumberTriState> callOrNeutralStates =  EnumSet.of(call, CopyNumberTriState.NEUTRAL);
+            final List<Set<CopyNumberTriState>> callOrNeutralConstraints = Collections.nCopies(length, callOrNeutralStates);
+
+            final double logCallOrNeutralProbability = fbResult.logConstrainedProbability(firstTarget, callOrNeutralConstraints);
+            final double logAllNeutralProbability = fbResult.logProbability(firstTarget, Collections.nCopies(length, CopyNumberTriState.NEUTRAL));
+
+            // log-sub-exp trick: log(exp(a) - exp(b)) = a + log(1 - exp(b - a))
```
davidbenjamin: log-sum-exp

src/main/java/org/broadinstitute/hellbender/tools/exome/DiscoverCopyNumberTriStateSegments.java
```diff
@@ -0,0 +1,484 @@
...
+
+        // read the input read-counts.
+        final ReadCountCollection originalReadCounts = readInputCounts(targets);
+
+        // Get the input read count target positions in the input targets collection.
+        final int[] sortedPositions = originalReadCounts.targets().stream()
+                .mapToInt(t -> targets.index(t.getName())).toArray();
+
+        // Check whether the input is in the same order as the targets in the input target collection.
+        // If not were re-arrangement them accordingly.
```
davidbenjamin: we rearrange

src/main/java/org/broadinstitute/hellbender/tools/exome/CopyNumberTriStateSegment.java
davidbenjamin: Or just mention that these qualities are described in the segment table columns enum.

src/main/java/org/broadinstitute/hellbender/tools/exome/hmm/CopyNumberTriStateHiddenMarkovModel.java
```diff
@@ -0,0 +1,354 @@
...
+     */
+    private final NormalDistribution deletionDistribution;
+
+    /**
+     * Duplication emission distribution.
+     */
+    private final NormalDistribution duplicationDistribution;
+
+    /**
+     * Holds the mean number of bps between targets in the same event.
```
davidbenjamin: don't abbreviate "base pairs"

src/main/java/org/broadinstitute/hellbender/tools/exome/hmm/CopyNumberTriStateHiddenMarkovModel.java
```diff
@@ -0,0 +1,354 @@
...
+     *
+     * @throws IllegalArgumentException if any of the model parameters has an invalid value.
+     */
+    public CopyNumberTriStateHiddenMarkovModel(final double eventRate,
+                                               final double meanNumberOfTargetsPerEvent,
+                                               final double deletionMean,
+                                               final double duplicationMean,
+                                               final double meanEventTargetDistance) {
+        checkParameterRange(VALID_EVENT_RATE_RANGE, eventRate, "event rate");
+        checkParameterRange(VALID_MEAN_NUMBER_OF_TARGETS_PER_EVENT_RANGE, meanNumberOfTargetsPerEvent,
```
davidbenjamin: IMO `ParamUtils.isPositive(eventRate)` would be clearer.

src/main/java/org/broadinstitute/hellbender/tools/exome/hmm/CopyNumberTriStateHiddenMarkovModel.java
```diff
@@ -0,0 +1,354 @@
...
+                "mean number of targets per event");
+        checkParameterRange(VALID_DELETION_CENTER_RANGE, deletionMean, "deletion mean depth coverage shift");
```
davidbenjamin: IMO, `ParamUtils.isNegative` would be clearer.

src/main/java/org/broadinstitute/hellbender/tools/exome/hmm/CopyNumberTriStateHiddenMarkovModel.java
```diff
@@ -0,0 +1,354 @@
...
+        checkParameterRange(VALID_DUPLICATION_CENTER_RANGE, duplicationMean, "duplication mean depth coverage shift");
```
davidbenjamin: `ParamUtils.isPositive`

src/main/java/org/broadinstitute/hellbender/tools/exome/hmm/CopyNumberTriStateHiddenMarkovModel.java
```diff
@@ -0,0 +1,354 @@
...
+
+        P = eventRate;
+        logP = Math.log(eventRate);
+        log1Minus2P = Math.log1p(-2 * eventRate);
+        T = meanNumberOfTargetsPerEvent;
+        logQ = -Math.log(meanNumberOfTargetsPerEvent);
+        log1MinusQ = Math.log1p(-Math.exp(logQ));
+        D = meanEventTargetDistance;
+        invD = 1.0 / D;
+        deletionDistribution = new NormalDistribution(deletionMean, 1);
```
davidbenjamin: Although defining `neutralDistribution` to have unit variance is definitional since we normalize to Z-scores, when `deletionDistribution` and `duplicationDistribution` are given unit variance this is a magic constant and should be defined as a static member.  The number 1.0 is an unfounded assumption of XHMM.

src/main/java/org/broadinstitute/hellbender/tools/exome/hmm/CopyNumberTriStateHiddenMarkovModel.java
```diff
@@ -0,0 +1,354 @@
...
+        duplicationDistribution = new NormalDistribution(duplicationMean, 1);
+    }
+
+    /**
+     * Validates the value of an parameter.
```
davidbenjamin: a parameter

src/main/java/org/broadinstitute/hellbender/tools/exome/hmm/CopyNumberTriStateHiddenMarkovModel.java
```diff
@@ -0,0 +1,354 @@
...
+     *   will depends on the declared input range extreme values.
+     * </p>
+     *
+     * @param validRange the valid range for the parameter value.
+     * @param value the parameter value itself.
+     * @param definition a human friendly description of the parameter to be used in an explanatory exception.
+     *
+     * @throws IllegalArgumentException if the value provided is in-valid.
+     */
+    private void checkParameterRange(final DoubleRange validRange, final double value, final String definition) {
```
davidbenjamin: Move this to `ParamUtils`.

src/main/java/org/broadinstitute/hellbender/tools/exome/hmm/CopyNumberTriStateHiddenMarkovModel.java
```diff
@@ -0,0 +1,354 @@
...
+            return nextState == CopyNumberTriState.NEUTRAL ? log1Minus2P : logP;
+        } else {
+            final double distance = calculateDistance(fromTarget, nextPosition);
+            final double logF = -distance * invD;
+            final double log1MinusF = Math.log1p(-Math.exp(logF));
+            if (nextState == CopyNumberTriState.NEUTRAL) {
+                // End of event transition prob.:
+                return MathUtils.logSumLog(new double[] {logF + logQ, log1MinusF + log1Minus2P});
+
+                //return MathUtils.logSumLog(new double[] {logF + logQ, log1MinusF + log1Minus2P});
```
davidbenjamin: delete this commented out line

src/main/java/org/broadinstitute/hellbender/tools/exome/hmm/CopyNumberTriStateHiddenMarkovModel.java
```diff
@@ -0,0 +1,354 @@
...
+                // Event "sign switch" (del. <--> dup.) transition prob.:
+                return log1MinusF + logP;
+            }
+        }
+    }
+
+    /**
+     * Calculate the distance between two targets.
+     * <p>
+     * If the target provided do not contain intervals, the distance is set to 0 in order to revert
```
davidbenjamin: targets

src/main/java/org/broadinstitute/hellbender/tools/exome/hmm/CopyNumberTriStateHiddenMarkovModel.java
```diff
@@ -0,0 +1,354 @@
...
+     * @return any values between 0 and {@link Double#POSITIVE_INFINITY}.
+     * @throws NullPointerException if any of the targets is {@code null}.
+     */
+    @VisibleForTesting
+    double calculateDistance(final Target fromTarget, final Target toTarget) {
+        final SimpleInterval fromInterval = fromTarget.getInterval();
+        final SimpleInterval toInterval = toTarget.getInterval();
+        if (fromInterval == null || toInterval == null) {
+            return 0.0;
+        } else if (!fromInterval.getContig().equals(toInterval.getContig())) {
```
davidbenjamin: This is at least the third time in this PR that comparison of two targets' contigs has come up.  It's worth extracting a method somewhere.

src/main/java/org/broadinstitute/hellbender/tools/exome/hmm/CopyNumberTriStateHiddenMarkovModel.java
```diff
@@ -0,0 +1,354 @@
...
+            return Double.POSITIVE_INFINITY;
+        } else {
+            final double distance = toInterval.getStart() - fromInterval.getEnd() - 1;
+            return distance > 0 ? distance : Math.max(0, fromInterval.getStart() - toInterval.getEnd() - 1);
```
davidbenjamin: What's the logic behind the return value for overlapping targets?

src/main/java/org/broadinstitute/hellbender/tools/exome/hmm/CopyNumberTriStateHiddenMarkovModel.java
```diff
@@ -0,0 +1,354 @@
...
+        final NormalDistribution distribution = stateDistribution(Utils.nonNull(state));
+        return distribution.logDensity(data);
+    }
+
+    /**
+     * Returns the appropriate emission distribution given the copy-number state.
+     * @param state the requested state.
+     * @return never {@code null.}
+     */
+    private NormalDistribution stateDistribution(final CopyNumberTriState state) {
```
davidbenjamin: It might make sense to put all the state distributions in a single `EnumMap<CopyNumberTriState, NormalDistribution>` container.

src/main/java/org/broadinstitute/hellbender/tools/exome/hmm/CopyNumberTriStateHiddenMarkovModel.java
```diff
@@ -0,0 +1,354 @@
...
+            return NEUTRAL_DISTRIBUTION;
+        } else if (state == CopyNumberTriState.DELETION) {
+            return deletionDistribution;
+        } else {
+            return duplicationDistribution;
+        }
+    }
+
+    @VisibleForTesting
+    public Double randomDatum(final CopyNumberTriState state, final Random rdn) {
```
davidbenjamin: If it's `@VisibleForTesting` can't you give it no access modifier to default to package protected instead of making it `public`?

src/main/java/org/broadinstitute/hellbender/tools/exome/hmm/CopyNumberTriStateHiddenMarkovModelArgumentCollection.java
```diff
@@ -0,0 +1,93 @@
...
+    public static final String MEAN_DELETION_COVERAGE_SHIFT_FULL_NAME = "meanDeletionCoverageSigmaShift";
+    public static final String MEAN_DELETION_COVERAGE_SHIFT_SHORT_NAME = "deletionShift";
+    public static final String MEAN_DUPLICATION_COVERAGE_SHIFT_FULL_NAME = "meanDuplicationCoverageSigmaShift";
+    public static final String MEAN_DUPLICATION_COVERAGE_SHIFT_SHORT_NAME = "duplicationShift";
+
+    @Argument(doc = "Probability that a target is the first in a deletion or duplication segment",
+              fullName = EVENT_RATE_FULL_NAME,
+              shortName = EVENT_RATE_SHORT_NAME,
+              optional = true)
+    public double eventSegmentRate = 1.0e-10;
```
davidbenjamin: This default seems too small.  In the XHMM paper they settle on 1e-8.

src/main/java/org/broadinstitute/hellbender/tools/exome/hmm/CopyNumberTriStateHiddenMarkovModelArgumentCollection.java
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * Creates a new model argument collection taking on the default values suggested in
+     *  <a href="http://www.ncbi.nlm.nih.gov/pmc/articles/PMC3484655/">Fromer et al. 2012</a>.
+     */
+    public CopyNumberTriStateHiddenMarkovModelArgumentCollection() {
+
+    }
+
+    public void validate() {
```
davidbenjamin: See comments above about using `ParamUtils.isPositive` etc when possible.

src/main/java/org/broadinstitute/hellbender/tools/exome/hmm/CopyNumberTriStateHiddenMarkovModelArgumentCollection.java
```diff
@@ -0,0 +1,93 @@
...
+        checkParameterRange(CopyNumberTriStateHiddenMarkovModel.VALID_EVENT_RATE_RANGE, eventSegmentRate, EVENT_RATE_FULL_NAME);
+        checkParameterRange(CopyNumberTriStateHiddenMarkovModel.VALID_DELETION_CENTER_RANGE, meanDeletionCoverageShift, MEAN_DELETION_COVERAGE_SHIFT_FULL_NAME);
+        checkParameterRange(CopyNumberTriStateHiddenMarkovModel.VALID_DUPLICATION_CENTER_RANGE, meanDuplicationCoverageShift, MEAN_DUPLICATION_COVERAGE_SHIFT_FULL_NAME);
+        checkParameterRange(CopyNumberTriStateHiddenMarkovModel.VALID_MEAN_INTRA_EVENT_TARGET_DISTANCE_RANGE, meanDistanceBetweenTargetsInSegment, MEAN_DISTANCE_BETWEEN_TARGETS_IN_SEGMENT_FULL_NAME);
+        checkParameterRange(CopyNumberTriStateHiddenMarkovModel.VALID_MEAN_NUMBER_OF_TARGETS_PER_EVENT_RANGE, meanTargetsPerSegment, MEAN_TARGETS_PER_SEGMENT_FULL_NAME);
+    }
+
+    private void checkParameterRange(final DoubleRange validRange, final double value, final String argumentName) {
```
davidbenjamin: Move to `ParamUtils`

src/main/java/org/broadinstitute/hellbender/utils/hmm/ForwardBackwardAlgorithm.java
```diff
@@ -93,11 +94,57 @@
...
+         * position.
+         *
+         * <p>
+         * Although in theory one does not need to indicate a position to evaluate the data-likelihood
+         * in practice, lack of float point precision may those small differences important (e.g. you
+         * some computation with small probabilities may result in a probability just over 1.0 which makes
+         * no sense).
+         * </p>
+         * <p>
+         * As a rule of thumb, to avoid such problems you should request the data-likelihood at the target when you
```
davidbenjamin: the target from which you evaluate the backward probability

src/main/java/org/broadinstitute/hellbender/utils/hmm/ForwardBackwardAlgorithm.java
```diff
@@ -393,33 +442,107 @@ public double logProbability(final S state, final T position) {
...
+                        .mapToDouble(i -> logForwardProbabilities[startIndex][i])
+                        .toArray();
+                for (int statesOffset = 1, dataOffset = startIndex + 1; dataOffset <= lastIndex; statesOffset++, dataOffset++) {
+                    final double[] previousLikelihoods = currentLikelihoods;
+                    final List<S> previousStates = currentStates;
+                    final T previousPosition = positions.get(dataOffset - 1);
+                    final T thisPosition = positions.get(dataOffset);
+                    final D thisData = data.get(dataOffset);
+
+                    currentStates = new ArrayList<>(Utils.nonNull(stateConstraints.get(statesOffset)));
```
davidbenjamin: Put in a comment about how you're considering all hidden state transitions from the previous position's constrained set to the current position's.

src/test/java/org/broadinstitute/hellbender/tools/exome/DiscoverCopyNumberTriStateIntegrationTest.java
```diff
@@ -0,0 +1,365 @@
...
+ */
+public class DiscoverCopyNumberTriStateIntegrationTest extends CommandLineProgramTest {
+
+    private static final File REALISTIC_TARGETS_FILE = new File("src/test/resources/org/broadinstitute/hellbender/tools/exome/hmm/realistic-targets.tab");
+
+    private static final TargetCollection<Target> REALIST_TARGETS;
+
+    static {
+        try (final TargetTableReader reader = new TargetTableReader(REALISTIC_TARGETS_FILE)) {
+            REALIST_TARGETS = new HashedListTargetCollection<>(reader.stream().collect(Collectors.toList()));
```
davidbenjamin: REALISTIC

src/test/java/org/broadinstitute/hellbender/tools/exome/DiscoverCopyNumberTriStateIntegrationTest.java
```diff
@@ -0,0 +1,365 @@
...
+/**
+ * Integration tests for {@link DiscoverCopyNumberTriStateSegments}
+ *
+ * @author Valentin Ruano-Rubio &lt;valentin@broadinstitute.org&gt;
+ */
+public class DiscoverCopyNumberTriStateIntegrationTest extends CommandLineProgramTest {
+
+    private static final File REALISTIC_TARGETS_FILE = new File("src/test/resources/org/broadinstitute/hellbender/tools/exome/hmm/realistic-targets.tab");
+
+    private static final TargetCollection<Target> REALIST_TARGETS;
```
davidbenjamin: REALISTIC

davidbenjamin:
@vruano As you can see I don't have much to say.  This is very readable code.

src/main/java/org/broadinstitute/hellbender/tools/exome/DiscoverCopyNumberTriStateSegments.java
vruano: Not really, XHMM's is 99.0 so perhaps I will change it to that but there is no actual reason for it; I think.

src/main/java/org/broadinstitute/hellbender/tools/exome/DiscoverCopyNumberTriStateSegments.java
vruano: With the input values in a list of unboxed, I think is too much hassle to use Apaches'  (it needs the doubles in a primitive array).

src/main/java/org/broadinstitute/hellbender/tools/exome/DiscoverCopyNumberTriStateSegments.java
vruano: not, actually it means log-sub-exp as in ```log (exp(a) - exp(b))```

src/main/java/org/broadinstitute/hellbender/tools/exome/hmm/CopyNumberTriStateHiddenMarkovModel.java
vruano: checkParameterRange does a bit more than isPositive in terms of composing a detailed error message. I prefer to keep it as it is.

src/main/java/org/broadinstitute/hellbender/tools/exome/hmm/CopyNumberTriStateHiddenMarkovModel.java
vruano: Same.

src/main/java/org/broadinstitute/hellbender/tools/exome/hmm/CopyNumberTriStateHiddenMarkovModel.java
vruano: Same.

src/main/java/org/broadinstitute/hellbender/tools/exome/hmm/CopyNumberTriStateHiddenMarkovModel.java
vruano: It is quite of a basic comparison... I don't see a good gain / hassle ratio. You want to add .sameContig(X) method to the Locatable interface? that would be in htsdk... Create a IntervalsUtil class for a single simple method? I prefere to leave it as it is.

src/main/java/org/broadinstitute/hellbender/tools/exome/hmm/CopyNumberTriStateHiddenMarkovModel.java
vruano: Just be robust, should not happen in practice. Do you want me to throw an exception otherwise?

src/main/java/org/broadinstitute/hellbender/tools/exome/hmm/CopyNumberTriStateHiddenMarkovModel.java
vruano: No, is used from another package.

src/main/java/org/broadinstitute/hellbender/tools/exome/hmm/CopyNumberTriStateHiddenMarkovModelArgumentCollection.java
vruano: Sure, bad default, will change. 

Yet even 1e-8 is quite small if you think about it.  If you have around ~100,000 targets per sample that means that you expect to see only 1 event every 100 samples. So it makes little sense and I guess is its compensating for false positives that happen in practice; perhaps some kind of under-the-counter multiple-test correction?

Also I think is wrong to try to use targets as a unit here as different capture or panels require to recalculate this value. I think we should eventually move into using base-pairs rather than targets which we can maintain constant per species.

vruano:
@davidbenjamin please rereviews. Notice I put that 199.9999 back otherwise some integration test would fail and I didn't want to have to refresh those values using XHMM; is not worth the hassle as it is a very arbitrary value. 

About things that I have NOT done, I had expose my arguments in the reviews; hopefully you find it convincing.

src/main/java/org/broadinstitute/hellbender/tools/exome/DiscoverCopyNumberTriStateSegments.java
davidbenjamin: True, my mistake!

src/main/java/org/broadinstitute/hellbender/tools/exome/hmm/CopyNumberTriStateHiddenMarkovModel.java
davidbenjamin: Okay, that's fine with me.

src/main/java/org/broadinstitute/hellbender/tools/exome/hmm/CopyNumberTriStateHiddenMarkovModel.java
davidbenjamin: @vruano I'm not sure what the right behavior is, but I feel like the distance between overlapping targets should never be *more* than the distance between non-overlapping targets.  I think returning `0` would make sense, as would an exception, as would redefining distance to be the distance between the starts.  I truly have no preference and leave it up to you.

davidbenjamin:
@vruano I'm convinced by everything you said.  All that remains is the return value in `calculateDistance` -- see comment above.

vruano:
@davidbenjamin you like the change?

davidbenjamin:
@vruano you may squash and merge.

Issue #67, opened 2015-07-23
Use actual probabilities in het pulldown


davidbenjamin:
The current approach is to perform a p-value test under a null hypothesis that a site is het to decide whether to include it.  The problem with this is the problem of all frequentist hypothesis testing in that it is totally agnostic of the other hypothesis, which in this case we know is that the site is homozygous.  This is a fairly simple application of Bayes' Theorem and will avoid throwing out perfectly good het sites. It could in principle also avoid treating homozygous sites as hets, but this possibility is unlikely due to a filter for low-coverage sites.  Still, by using a probabilistic approach one can dispense with this ad hoc filter, which will in turn let us keep more het sites.

davidbenjamin:
One other danger in the current approach is that if some true het site has large allelic bias, it could have alt/ref counts that are skewed enough to give a very small p value for heterozygosity assuming no bias or small bias.  For example, alt/ref counts of 60/140 are obviously a het, even though their p-value under the null hypothesis of binomial counts is small.

LeeTL1220:
Does this happen in practice (in a normal)?  Enough that we care?

davidbenjamin:
There's a decent chance that we don't care. I'll do some back-of-the-envelope calculations to see under what conditions the p-value method loses hets.

davidbenjamin:
I did some simple calculations, and it's worth it to do something probabilistic.  However, the stuff I have in the proposed methods docs is overkill (what was I thinking???).  We can do a lot better than the p-value while keeping things simple.  Here are some highlights:

*By definition, a p-value of 0.05 means that we are guaranteed to throw out 5% of all true hets.
*But it gets worse if there is even quite moderate allelic bias.  For example, if the total depth is 50 reads, 19 alt reads is the threshold at which p = 0.05.  But with a modest allelic bias of 47% alt probability this many alt reads or fewer occurs 13% of the time, so we are quite likely to throw away this het.
*If there are 10 total reads and 2 alt reads, p = 0.054 and we accept it as a het, when it could easily be a hom ref with sequencing errors.
*Note that all of these calls are obvious to us without any sort of statistical model.

[...]

PR #314, opened 2016-01-21
Likelihood-based het pulldown. Closes issue #67.

davidbenjamin:
See comments in issue #67.  Bottom line is that this PR saves a small but non-negligible fraction of good hets and might prevent a few homs from slipping in at low coverage.


davidbenjamin:
@samuelklee This is very low-priority.  I would have done it later but I needed something to do while rocking baby back to sleep.

davidbenjamin:
Failing on that spark SVD bug.  Re-running Travis.

samuelklee:
since, like ExomeReadCounts, GetHetCoverage is by far the rate-limiting step (especially on fargo'd samples), i'm going to punt reviewing/merging this until the ACNV eval settles down a bit---i don't want to have to rerun all the pulldowns.  once we settle on the iteration for the back half of ACNV with the old pulldown, i'll evaluate that against this new pulldown (although i'm pretty sure that the latter can only improve the results).  however, just wanted to say that i've taken a glance through the code and things look generally good!

davidbenjamin:
That seems very sensible.

src/main/java/org/broadinstitute/hellbender/tools/exome/HetPulldownCalculator.java
```diff
@@ -88,37 +81,37 @@ public HetPulldownCalculator(final File refFile, final File snpFile) {
...
-        final double pval = new BinomialTest().binomialTest(totalBaseCount, majorReadCount, hetAlleleFraction,
-                AlternativeHypothesis.TWO_SIDED);
-
-        return pval >= pvalThreshold;
+    @VisibleForTesting
+    static boolean isHet(final int refReadCount, final int altReadCount,
+                                final double errorRate, final double likelihoodRatioThreshold) {
+        final int minorReadCount = Math.min(refReadCount, altReadCount);
+        final int majorReadCount = Math.max(refReadCount, altReadCount);
+
```
samuelklee: whitespace

src/main/java/org/broadinstitute/hellbender/tools/exome/HetPulldownCalculator.java
```diff
@@ -88,37 +81,37 @@ public HetPulldownCalculator(final File refFile, final File snpFile) {
...
      *
      * <p>
-     *     Compatibility is defined by a p-value threshold.  That is, compute the two-sided p-value of observing
-     *     a number of major read counts out of a total number of reads, assuming the given heterozygous
-     *     allele fraction.  If the p-value is less than the given threshold, then reject the null hypothesis
-     *     that the heterozygous allele fraction is as given (i.e., SNP is likely to be homozygous) and return false,
-     *     otherwise return true.
+     *     Compatibility is defined by a likelihood ratio threshold between two hypotheses.  First, that the site is
+     *     heterozygous in which case the likelihood is nCa * (1/2)^n.  Second, that the site is homozygous and minor
+     *     allele reads are sequencing errors, with likelihoo nCa * errorRate^a * (1 - errorRate)^(n-a).
```
samuelklee: likelihood

src/main/java/org/broadinstitute/hellbender/tools/exome/HetPulldownCalculator.java
```diff
@@ -36,6 +38,9 @@
     private static final int READ_DEPTH_THRESHOLD = 10;
     private static final int MAX_INTERVALS_FOR_INDEX = 25000;
 
+    @VisibleForTesting
+    static final Nucleotide[] BASES = {Nucleotide.A, Nucleotide.C, Nucleotide.G, Nucleotide.T};
```
samuelklee: just checking, this is meant to be package private?

src/main/java/org/broadinstitute/hellbender/tools/exome/HetPulldownCalculator.java
```diff
@@ -49,38 +54,26 @@ public HetPulldownCalculator(final File refFile, final File snpFile) {
...
-     * Returns map of base-pair counts at a given locus.  Only includes ACGT counts.
+     * Returns base-pair counts at a given locus.
      * @param locus locus
      * @return      map of base-pair counts
      */
-    public static Map<Character,Integer> getPileupBaseCounts(final SamLocusIterator.LocusInfo locus) {
-        int aCount = 0, cCount = 0, gCount = 0, tCount = 0;
-
+    @VisibleForTesting
+    static Nucleotide.Counter getPileupBaseCounts(final SamLocusIterator.LocusInfo locus) {
```
samuelklee: same here

src/main/java/org/broadinstitute/hellbender/tools/exome/HetPulldownCalculator.java
```diff
@@ -196,21 +188,16 @@ private Pulldown getHetPulldown(final File bamFile, final IntervalList snpInterv
...
+                final Pair<Integer, Integer> refAltCounts = refAndAltReadCounts(baseCounts, refBase);
+                final int refReadCount = refAltCounts.getLeft();
+                final int altReadCount = refAltCounts.getRight();
 
-                if (sampleType == SampleType.NORMAL &&
-                        !isPileupHetCompatible(baseCounts, totalBaseCount, hetAlleleFraction, pvalThreshold)) {
-                    continue;
+                if (sampleType == SampleType.TUMOR || isHet(refReadCount, altReadCount, errorRate, likelihoodRatioThreshold)) {
+                    final SimpleInterval interval = new SimpleInterval(locus.getSequenceName(), locus.getPosition(), locus.getPosition());
+                    hetPulldown.add(interval ,refReadCount, altReadCount);
```
samuelklee: whitespace before/after comma

src/test/java/org/broadinstitute/hellbender/tools/exome/HetPulldownCalculatorUnitTest.java
```diff
@@ -146,15 +154,15 @@ public void testIsPileupHetCompatible(final Map<Character, Integer> baseCounts,
...
-                {0.5, 0.05, normalHetPulldown1},
-                {0.45, 0.05, normalHetPulldown2}
+                {0.01, 1.0, normalHetPulldown1},
+                {0.1, 1.0, normalHetPulldown2}
         };
     }
 
     @Test(dataProvider = "inputGetNormalHetPulldown")
-    public void testGetNormalHetPulldown(final double hetAlleleFraction, final double pvalThreshold,
+    public void testGetNormalHetPulldown(final double errorRate, final double likelihooRatioThreshold,
```
samuelklee: likelihood

src/test/java/org/broadinstitute/hellbender/tools/exome/HetPulldownCalculatorUnitTest.java
```diff
@@ -146,15 +154,15 @@ public void testIsPileupHetCompatible(final Map<Character, Integer> baseCounts,
...
                                          final Pulldown expected) {
-        final Pulldown result = calculator.getNormal(NORMAL_BAM_FILE, hetAlleleFraction, pvalThreshold);
+        final Pulldown result = calculator.getNormal(NORMAL_BAM_FILE, errorRate, likelihooRatioThreshold);
```
samuelklee: same

samuelklee:
done with my review, just extremely minor issues.  i'm going to run an evaluation, but getting the pulldowns might take a while so you can go ahead and merge in the meantime.

src/main/java/org/broadinstitute/hellbender/tools/exome/HetPulldownCalculator.java
davidbenjamin: fixed

src/main/java/org/broadinstitute/hellbender/tools/exome/HetPulldownCalculator.java
davidbenjamin: fixed

src/main/java/org/broadinstitute/hellbender/tools/exome/HetPulldownCalculator.java
davidbenjamin: Yes, for the unit test.

src/main/java/org/broadinstitute/hellbender/tools/exome/HetPulldownCalculator.java
davidbenjamin: yes, same idea

samuelklee:
getting slightly different counts on HCC1143 pulldowns...need to investigate further

src/main/java/org/broadinstitute/hellbender/tools/exome/HetPulldownCalculator.java
davidbenjamin: fixed

src/test/java/org/broadinstitute/hellbender/tools/exome/HetPulldownCalculatorUnitTest.java
davidbenjamin: fixed

src/test/java/org/broadinstitute/hellbender/tools/exome/HetPulldownCalculatorUnitTest.java
davidbenjamin: fixed

davidbenjamin:
@samuelklee "getting slightly different counts on HCC1143 pulldowns...need to investigate further" -- should I still merge?

samuelklee:
let's hold off for now, i'm still trying to track down the reason.

samuelklee:
actually, i'm betting that it's probably just a difference in how the alt counts are defined.  ok to merge!

samuelklee:
hmm, actually, which reads should count towards the alt count?  here you only take the second-most common nucleotide.  however, there are several pileups where you will have reads for more than two nucleotides---shouldn't these be accounted for in the likelihood and the definition of the error rate?

davidbenjamin:
For input to the allelic fraction model I think counting only the dominant non-ref allele is the correct approach.  But for the error rate other reads matter.  However, perhaps the code as written is in effect referring to one third of the error rate -- one of every three substitution errors on a true ref (alt) yields the true alt (ref) while we discard the other two possibilities.

samuelklee:
about 35 of the PRAD samples have finished running, and it looks like MAE/RMSE concordance with HAPSEG is significantly decreased.  this could be because the extra hets are finding events that HAPSEG missed, but it's hard to say if this is the case.  the new events are mostly small (and hence have large uncertainty in MAF) but they do not seem to cluster.  let's discuss today!

davidbenjamin:
Wow, that's really interesting.  Good thing we didn't merge yet!

samuelklee:
After discussion, we decided to close this PR for the time being.  It's possible that the less conservative filter was letting in hets with greater allelic bias and negatively affecting concordance with HAPSEG under our allelic-bias model.  However, we should definitely revisit this pulldown after an allelic PoN or a per-site allelic-bias model has been implemented.

@davidbenjamin, perhaps make a separate branch if you want to keep the edits to the docs?

PR #330, opened 2016-01-29
Added calculation of MCMC credible intervals in PosteriorSummary using mllib kernel density estimation.

### LINKED ISSUE #315: Extra whitespace for NaN values in ACNV segment files.
In the segment-file output of ACNV, MAF posterior means and standard deviations are currently output as:

    String.format("%6.4f", segment.getMinorAlleleFractionPosteriorSummary().mean())
    String.format("%6.4f", segment.getMinorAlleleFractionPosteriorSummary().standardDeviation())

However, this adds 3 leading whitespaces when these values are Double.NaN (which occurs when a segment has no SNPs).  A minor issue, but a bit unexpected and annoying when parsing these files downstream and hence worth fixing up.  Mea culpa.

### LINKED ISSUE #320: Remove extra newline in logging for AlleleFractionInitializer.
Not a huge deal, but would make the logs look a bit cleaner.

samuelklee:
-Made corresponding minor changes to ACNV classes and tests.
-Cleaned up imports.
-Minor issues: closes #315, closes #320.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/ACNVModeller.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/ACNVModeller.java
@@ -2,6 +2,7 @@ package org.broadinstitute.hellbender.tools.exome;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.apache.spark.api.java.JavaSparkContext;
 import org.broadinstitute.hellbender.tools.exome.allelefraction.AlleleFractionModeller;
 import org.broadinstitute.hellbender.utils.SimpleInterval;
 import org.broadinstitute.hellbender.utils.mcmc.PosteriorSummary;
@@ -22,6 +23,9 @@ public final class ACNVModeller {
 
     private static final int MAX_SIMILAR_SEGMENT_MERGE_ITERATIONS = 25;
 
+    //use 95% HPD interval to construct {@link PosteriorSummary} for segment means and minor allele fractions
+    private static final double CREDIBLE_INTERVAL_ALPHA = 0.05;
+
     public static final Logger logger = LogManager.getLogger(ACNVModeller.class);
 
     private SegmentedModel segmentedModel;
@@ -33,6 +37,7 @@ public final class ACNVModeller {
     private final int numBurnInCopyRatio;
     private final int numSamplesAlleleFraction;
     private final int numBurnInAlleleFraction;
+    private final JavaSparkContext ctx;
 
     /**
      * Constructs a copy-ratio and allele-fraction modeller for a {@link SegmentedModel},
@@ -45,16 +50,19 @@ public final class ACNVModeller {
      * @param numBurnInCopyRatio        number of burn-in samples to discard for copy-ratio model MCMC
      * @param numSamplesAlleleFraction  number of total samples for allele-fraction model MCMC
      * @param numBurnInAlleleFraction   number of burn-in samples to discard for allele-fraction model MCMC
+     * @param ctx                       JavaSparkContext, used for kernel density estimation in {@link PosteriorSummary}
      */
     public ACNVModeller(final SegmentedModel segmentedModel, final String outputPrefix,
                         final int numSamplesCopyRatio, final int numBurnInCopyRatio,
-                        final int numSamplesAlleleFraction, final int numBurnInAlleleFraction) {
+                        final int numSamplesAlleleFraction, final int numBurnInAlleleFraction,
+                        final JavaSparkContext ctx) {
         this.segmentedModel = segmentedModel;
         this.outputPrefix = outputPrefix;
         this.numSamplesCopyRatio = numSamplesCopyRatio;
         this.numBurnInCopyRatio = numBurnInCopyRatio;
         this.numSamplesAlleleFraction = numSamplesAlleleFraction;
         this.numBurnInAlleleFraction = numBurnInAlleleFraction;
+        this.ctx = ctx;
         fitModel(numSamplesCopyRatio, numBurnInCopyRatio, numSamplesAlleleFraction, numBurnInAlleleFraction);
     }
 
@@ -63,7 +71,7 @@ public final class ACNVModeller {
      * specifying number of total samples and number of burn-in samples for Markov-Chain Monte Carlo model fitting
      * performed after each iteration.
      * <p>
-     *     First, the list of {@link ACNVModeledSegment ) (i.e., the list of {@link PosteriorSummary} for each
+     *     First, the list of {@link ACNVModeledSegment} (i.e., the list of {@link PosteriorSummary} for each
      *     segment copy-ratio mean and minor allele fraction) is taken from the initial model fit.
      *     Then, {@link SegmentMergeUtils#mergeSimilarSegments} is used to merge segments in this
      *     list given specified copy-ratio and minor-allele-fraction merging thresholds, after which another MCMC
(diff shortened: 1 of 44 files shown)


samuelklee:
please review, @davidbenjamin!  apologies for the many changed files, mostly from optimize imports.  most nontrivial code is that for calculating the credible intervals in PosteriorSummary and its test class, although there were some minor changes throughout the ACNV classes to make things consistent.

if @LeeTL1220 could also check that i'm bringing in spark mllib correctly, that would be appreciated.  i noticed that i had to set the log level in AllelicCNV to WARN, otherwise i would get a ridiculous amount of INFO messages when running queue jobs on the farm---not sure if this means i'm configuring something incorrectly?

src/main/java/org/broadinstitute/hellbender/tools/exome/ACNVModeller.java
```diff
@@ -72,15 +80,15 @@ public ACNVModeller(final SegmentedModel segmentedModel, final String outputPref
      * Segment files (with filenames specified by {@link ACNVModeller#outputPrefix} are output for the initial
      * model fit and the model fits after each merge iteration as a side effect.
      *
-     * @param sigmaThresholdSegmentMean         threshold number of standard deviations for segment-mean similarity
-     * @param sigmaThresholdMinorAlleleFraction threshold number of standard deviations for minor-allele-fraction similarity
+     * @param intervalThresholdSegmentMean         threshold number of credible intervals for segment-mean similarity
```
davidbenjamin: Clarify this javadoc.  `number` to me implies an integer.

davidbenjamin: Actually, I take it back, as long as it's described in `SegmentMergeUtils`.

src/main/java/org/broadinstitute/hellbender/utils/mcmc/PosteriorSummary.java
```diff
@@ -1,28 +1,117 @@
...
+        Utils.validateArg(samples.size() > 0, "Number of samples must be greater than zero.");
+        final double sampleMin = Collections.min(samples);
+        final double sampleMax = Collections.max(samples);
+        final double sampleMean = new Mean().evaluate(Doubles.toArray(samples));
+        final double sampleStandardDeviation = new StandardDeviation().evaluate(Doubles.toArray(samples));
+        if (sampleStandardDeviation == 0.0 || Double.isNaN(sampleMean)) {
+            return sampleMean;
+        }
+        //use Silverman's rule to set bandwidth for kernel density estimation from sample standard deviation
+        final double bandwidth = 1.06 * sampleStandardDeviation * Math.pow(samples.size(), -0.2);
```
davidbenjamin: extract a static constant

davidbenjamin: Or, if you wish, a static lambda containing the whole line.

src/test/java/org/broadinstitute/hellbender/utils/mcmc/PosteriorSummaryTest.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.Random;
+import java.util.stream.Collectors;
+
+/**
+ * Tests for {@link PosteriorSummary}.  Tests that posterior mode and highest-posterior-density credible interval
+ * are recovered for samples drawn from various posterior distributions.
+ *
+ * @author Samuel Lee &lt;slee@broadinstitute.org&gt;
+ */
+public final class PosteriorSummaryTest extends BaseTest {
```
davidbenjamin: Rename to `PosteriorSummaryUnitTest`.

src/test/java/org/broadinstitute/hellbender/utils/mcmc/PosteriorSummaryTest.java
```diff
@@ -0,0 +1,87 @@
...
+    private static final int RANDOM_SEED = 42;
+    private static final RandomGenerator rng = RandomGeneratorFactory.createRandomGenerator(new Random(RANDOM_SEED));
+
+    private static final List<Double> normalSamples = toList(new NormalDistribution(rng, 10., 1).sample(100));
+    private static final List<Double> normalSamplesSmall = toList(new NormalDistribution(rng, 10., 1).sample(10));
+    private static final List<Double> normalSamplesScaled = toList(new NormalDistribution(rng, 10., 0.01).sample(100));
+
+    private static final List<Double> betaSamples = toList(new BetaDistribution(rng, 2, 10).sample(100000));
+    private static final List<Double> betaSamplesEdgeMode = toList(new BetaDistribution(rng, 10, 1).sample(1000));
+
```
davidbenjamin: I would like to see one more test case (which ought to be very similar to the true posterior on MAFs) in which you take a Beta centered near 1/2 (so that it has large weight above 1/2) and reflect it so that f > 1/2 is mapped to 1 - f.

davidbenjamin:
@samuelklee  Just a couple little comments and a request for one more test case.

src/main/java/org/broadinstitute/hellbender/tools/exome/ACNVModeller.java
samuelklee: also, note that the command line --help will tell you that these arguments are Doubles and will give the default values.

src/test/java/org/broadinstitute/hellbender/utils/mcmc/PosteriorSummaryTest.java
samuelklee: done (also caught a couple of other tests that needed to be renamed)

src/test/java/org/broadinstitute/hellbender/utils/mcmc/PosteriorSummaryTest.java
samuelklee: done. also changed up the other beta tests slightly to reduce the required number of samples (i realized that spark warnings were being raised with sample sizes of 100000)

src/main/java/org/broadinstitute/hellbender/utils/mcmc/PosteriorSummary.java
samuelklee: i wouldn't expect users to look for an extracted method in this class if they were to need it elsewhere (say for histogram binning, etc.), so i guess i will go with extracting static constants.  i'm not sure if the case is that strong for extracting here (since i think magic number-ness is sufficiently demystified by the comment and i don't think these numbers will need to be used elsewhere or changed, e.g., to more sig figs, in this class in the future), but better safe than sorry, i guess!  i also added some more comments throughout this method.

src/test/java/org/broadinstitute/hellbender/utils/mcmc/PosteriorSummaryTest.java
samuelklee: also added some tests for samples containing NaN and identical values

samuelklee:
thanks, @davidbenjamin, back to you!  @LeeTL1220, please also check spark business when you get a chance!

src/main/java/org/broadinstitute/hellbender/utils/mcmc/PosteriorSummary.java
davidbenjamin: I agree that the case is not very strong.  I actually waffled on this one for a while.  For me the tiebreaker was consistency in coding style.

samuelklee:
also, once this goes into master, i'd like to run some samples through it before handing it to the guys downstairs to make absolutely sure nothing broke over the last couple of PRs.  since i won't have to rerun the pulldowns, this will be quick.

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentTableColumns.java
```diff
@@ -12,11 +12,12 @@
     SAMPLE("Sample"), CONTIG("Chromosome"), START("Start"), END("End"),
     NUM_PROBES("Num_Probes"), MEAN("Segment_Mean"), SD("Segment_Std"), CALL("Segment_Call"),
     NUM_TARGETS("Num_Targets"), NUM_SNPS("Num_SNPs"),
-    SEGMENT_MEAN_POSTERIOR_MEAN("Segment_Mean_Post_Mean"),
-    SEGMENT_MEAN_POSTERIOR_STANDARD_DEVIATION("Segment_Mean_Post_Std"),
-    MINOR_ALLELE_FRACTION_POSTERIOR_MEAN("MAF_Post_Mean"),
-    MINOR_ALLELE_FRACTION_POSTERIOR_STANDARD_DEVIATION("MAF_Post_Std"),
-
+    SEGMENT_MEAN_POSTERIOR_MODE("Segment_Mean_Post_Mode"),
```
LeeTL1220: Should we call this ``Center`` instead of ``Mode``?

samuelklee: hmm, in this case, ACNV specifically sets `PosteriorSummary.center()` to be the mode, so i think it's worth keeping this extra information in the column header.  in any case, if we change this in the future, the code will have to change in two places regardless---the place where we set the center value, and the place where we document what that center is---so why not let the buck stop at the column header?

however, at the same time, we should keep `PosteriorSummary` flexible enough to store whatever we want in the `center` field.

PR #337, opened 2016-01-30
Exposed a lot more CBS parameters for PerformSegmentation CLI.

### LINKED ISSUE #325: Expose sdundo and the rest of the CBS R parameters that are not exposed
``sdundo`` most important.

LeeTL1220:
Closes #325 

Exposes most of the remaining parameters.

Testing only checks that values in the output actually change.  Exact changes would be very onerous to test.  ``trim`` could not be tested this way.  However, output of the R script indicated that alternate values of ``trim`` were being passed in.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/PerformSegmentation.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/PerformSegmentation.java
@@ -92,7 +92,7 @@ public final class PerformSegmentation extends CommandLineProgram {
     protected File weightFile = null;
 
     @Argument(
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/hellbender/tools/exome/PerformSegmentation.java
```diff
@@ -57,6 +91,94 @@
     )
     protected File weightFile = null;
 
+    @Argument(
+            doc = "Please see https://www.bioconductor.org/packages/release/bioc/manuals/DNAcopy/man/DNAcopy.pdf",
```
achevali: Any reason this doesn't have (Advanced) in front of it?

PR #342, opened 2016-01-31
CLI for allele fraction plotting

davidbenjamin:


src/test/java/org/broadinstitute/hellbender/tools/exome/PlotSegmentedCopyRatioIntegrationTest.java
```diff
@@ -53,8 +51,6 @@ public void testUnLoggedCommandLineSexChrs() throws IOException {
         final String sampleName = "HCC1143";
         File tDir = IOUtils.tempDir("Test", "Plotting");
         System.out.println(tDir);
```
LeeTL1220: There is never a reason to have a system.out.println.  Use logging, instead (examples are found in other CLIs)

src/test/java/org/broadinstitute/hellbender/utils/plotter/AlleleFractionSegmentedPlotterTest.java
```diff
@@ -8,7 +8,7 @@
 import java.io.File;
 import java.io.IOException;
 
-public class AlleleFractionPlotterTest extends BaseTest {
+public class AlleleFractionSegmentedPlotterTest extends BaseTest {
     @Test
     public void testAlleleFractionPlotting() throws IOException {
```
LeeTL1220: I do not like tests that throw IOExceptions.  I would argue that you should catch the IOException in the test and use:

```java
try {
    //.....snip....
} catch (final IOException ioe) {
    Assert.fail("Could not run test due to problems with a file", ioe);
}
```

src/test/java/org/broadinstitute/hellbender/tools/exome/PlotSegmentedCopyRatioIntegrationTest.java
```diff
@@ -21,8 +21,6 @@ public void testUnLoggedCommandLine() throws IOException {
...
```
LeeTL1220: Can you use logging instead?

LeeTL1220:
Minor stuff... back to you @davidbenjamin

src/test/java/org/broadinstitute/hellbender/tools/exome/PlotSegmentedCopyRatioIntegrationTest.java
samuelklee: also note that i removed these in #330, so if you decide to log, then rebase accordingly.  although it looks to me like this was just debugging stuff that got left behind?

davidbenjamin: Debugging stuff that I wasn't sure whether to keep.  removed in this commit.

src/test/java/org/broadinstitute/hellbender/utils/plotter/AlleleFractionSegmentedPlotterTest.java
davidbenjamin: Tried to do this and IntelliJ realized that `runCommandLine` never throws an `IOException`, so the `throw` was silly to begin with.  Removed -- and now the code looks like all the other integration tests in the repo which seem to throw only when there's an expected exception.

davidbenjamin:
Back to @LeeTL1220

Issue #245, opened 2015-12-11
ACNV evaluation


samuelklee:
- [x] get /dsde/working/slee/acs-eval/scripts/acs.sh running on gsa queue instead of using ParallelShell jobrunner
- [x] try running normal-normal pairs
- [ ] run GATK CNV, get normal/tumor HetPulldowns, and run GATK ACNV:
 - [x] STAD samples
 - [ ] amaro's CLL samples
 - [x] ignaty et al.'s samples
    - [x] LUAD (46/622, a few repeat cases w/ different normals---note queue script currently overwrites pulldown and result in such cases)
    - [x] THCA (52/511, a few repeat cases w/ different normals---note queue script currently overwrites pulldown and result in such cases)
    - [x] PRAD (50/51, one sample failing to complete)
- [x] CRSP validation (on purity series only) with revised segments
 - [x] change output of ACSModeledSegment segment files to be compatible with ModeledSegment files (unlog segment mean---i.e., transform log_2 segment-mean posterior samples appropriately---and change column headers in SegmentTableColumns, may need to use blank values for calls, depending on what the CRSP validation needs?) -- PR #327.
 - [x] check thresholds/evaluate procedure for similar-segment merging? (minor refactoring may be necessary)
 - [x] Run purity samples through GATK CNV and ACNV (incl. conversion to GATK CNV seg file)
 - [x] Run samples through CRSP validation script
- [ ] concordance with amaro's samples
 - [ ] ...
- [x] concordance with ignaty et al.'s samples
 - [x] scripts for fetching annotations and making concordance tables, plots, histograms
 - [x] concordance on MAF < 0.46 cut: As expected, MAE was only lower for ACS because of balanced segments (ACS pegs to 0.5, not sure if HAPSEG does as well).  After cut, ACNV is more concordant with HAPSEG than is ACS on both MAE and RMSE. Interestingly, this is primarily due to ACNV finding more segments (which might arise from oversegmentation or from true events) that are almost all discordant with HAPSEG balanced segments (and hence, look like outliers, leading to high RMSE); however, for segments where all three are roughly concordant, ACNV's estimates of the MAF tend to be slightly more concordant with those from ACS than with those from HAPSEG.
[...]

PR #351, opened 2016-02-05
Changed ACNV segment union to only use CR segments not called neutral.

### LINKED ISSUE #331: Add option to run ACNV in SNP-only mode.

### LINKED ISSUE #333: Segment-union step could filter on CR calls.
That is, only take the most obvious breakpoints to avoid oversegmentation from the noisy CR data.  This would be a stopgap until we have probabilistic breakpoints.

samuelklee:
By default.  Also added flag to allow use of all CR segments and changed some default parameter values to those used in evaluation.  Closes #331, closes #333.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/AllelicCNV.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/AllelicCNV.java
@@ -10,6 +10,7 @@ import org.broadinstitute.hellbender.utils.SimpleInterval;
 
 import java.io.File;
 import java.util.List;
+import java.util.stream.Collectors;
 
 /**
  * Detects copy-number events using allelic-count data and GATK CNV output.
@@ -20,7 +21,7 @@ import java.util.List;
         summary = "Detect copy-number events in a tumor sample using allelic-count data and GATK CNV output. " +
                 "Allelic-count data (reference/alternate counts from the GetHetCoverage tool) is segmented using " +
                 "circular binary segmentation; the result is combined with the target coverages " +
-                "and segments found by the GATK CNV tool. Bayesian parameter estimation of models for the " +
+                "and called segments found by the GATK CNV tool. Bayesian parameter estimation of models for the " +
                 "copy ratios and minor allele fractions in each segment is performed using Markov chain Monte Carlo.",
         oneLineSummary = "Detect copy-number events using allelic-count data and GATK CNV output.",
         programGroup = CopyNumberProgramGroup.class
@@ -56,6 +57,9 @@ public class AllelicCNV extends SparkCommandLineProgram {
     protected static final String INTERVAL_THRESHOLD_ALLELE_FRACTION_LONG_NAME = "intervalThresholdAlleleFraction";
     protected static final String INTERVAL_THRESHOLD_ALLELE_FRACTION_SHORT_NAME = "simThAF";
 
+    protected static final String USE_ALL_COPY_RATIO_SEGMENTS_LONG_NAME = "useAllCopyRatioSegments";
+    protected static final String USE_ALL_COPY_RATIO_SEGMENTS_SHORT_NAME = "useAllCRSeg";
+
     @Argument(
             doc = "Input file for tumor-sample ref/alt read counts at normal-sample heterozygous-SNP sites (output of GetHetCoverage tool).",
             fullName = ExomeStandardArgumentDefinitions.TUMOR_ALLELIC_COUNTS_FILE_LONG_NAME,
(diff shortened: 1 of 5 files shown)


samuelklee:
merging CR neutral segments seems to help with oversegmentation, although some issues remain (see comment in #245).

note that default behavior if a CNV seg file with no calls is provided is to ignore all CR segments; the added flag must be used to enable their use.  in the future, we can change this so that the CNV seg file is not required, but i decided to keep the CNV -> ACNV workflow intact for now.

src/main/java/org/broadinstitute/hellbender/tools/exome/AllelicCNV.java
```diff
@@ -135,7 +139,7 @@
             shortName = NUM_BURN_IN_ALLELE_FRACTION_SHORT_NAME,
             optional = true
     )
-    protected int numBurnInAlleleFraction = 50;
+    protected int numBurnInAlleleFraction = 100;
```
LeeTL1220: Why did these parameters need to be increased?

src/main/java/org/broadinstitute/hellbender/tools/exome/AllelicCNV.java
```diff
@@ -151,7 +155,17 @@
...
+
+    @Argument(
+            doc = "Enable use of all copy-ratio--segment breakpoints. " +
+                    "(Default behavior uses only breakpoints from segments not called copy neutral, " +
+                    "if calls are available in output of GATK CNV provided, and none otherwise.)",
+            fullName = USE_ALL_COPY_RATIO_SEGMENTS_LONG_NAME,
+            shortName = USE_ALL_COPY_RATIO_SEGMENTS_SHORT_NAME,
+            optional = true
+    )
+    protected boolean useAllCopyRatioSegments = false;
```
LeeTL1220: Is the called segment file configured to be the input in the workflow Queue scripts?  Not in the version I have....

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentMergeUtils.java
```diff
@@ -68,6 +68,42 @@ public static SimpleInterval mergeSegments(final SimpleInterval segment1, final
...
+                    (segment1.getCall().equals(ReCapSegCaller.NEUTRAL_CALL) ||
+                            segment1.getCall().equals(ModeledSegment.NO_CALL)) &&
+                    (segment2.getCall().equals(ReCapSegCaller.NEUTRAL_CALL) ||
+                            segment2.getCall().equals(ModeledSegment.NO_CALL))) {
+
+                final SimpleInterval mergedInterval =
+                        SegmentMergeUtils.mergeSegments(segment1.getSimpleInterval(), segment2.getSimpleInterval());
+                //use dummy values for targetCount and segmentMean; these will not be used
+                mergedSegments.set(index, new ModeledSegment(mergedInterval, ModeledSegment.NO_CALL, 0, 0));
+                mergedSegments.remove(index + 1);
```
LeeTL1220: Are you setting the probe count here to zero?  This would be pretty easy to keep in sync.

src/main/java/org/broadinstitute/hellbender/tools/exome/AllelicCNV.java
samuelklee: no real reason, i just decided to switch the defaults to the values i've been using for evaluation.  probably a bit overkill, but runtime is still reasonable (with merging CR neutral, typically ~5-15 minutes for LUAD, max ~30 minutes).

src/main/java/org/broadinstitute/hellbender/tools/exome/AllelicCNV.java
samuelklee: see https://github.com/broadinstitute/recapseg-hb-eval/pull/49 mentioned above.

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentMergeUtils.java
samuelklee: I think this is actually one use case we are trying to avoid.  I only need to know the call to merge the neutral ModeledSegments here, so I have to set both the target counts and the segment means to dummy values (you can see that I convert to SimpleIntervals in the return statement, so these dummy values get tossed).

You'll note that it's also awkward to carry around the CBS segment mean here as well.  I think that once we add proper model fitting to the CNV workflow (so that ACNVModeledSegment can more sensibly inherit from CNVModeledSegment) things will clean up.

samuelklee:
responded to comments, let me know when i have the OK to merge!

PR #352, opened 2016-02-05
OjAlgoAdapter

apete:
I wrote an ojAlgo adapter class that could be useful to you.

Also modified OjAlgoSingularValueDecomposer to make use of that adapter.

I believe you also need a reverse-adapter. The SVD outputs 3 potentially large matrices. Each of those matrices are copied - twice.


LeeTL1220:
@apete Hey, many thanks!  We appreciate it.  

- I want to make sure that you know about the license in GATK protected.  It is non-standard.  Can you please confirm that you understand?

I'll review shortly if you understand the license.  Thanks again!  

I understand about the adapter for RealMatrix and that we are making copies.

@akiezun

apete:
I think I get the main points, but tire easily when reading that kind of language.

I have no plans to use GATK in any way, and the small piece of code I wrote is a gift if you want it.

LeeTL1220:
@apete We really appreciate the contribution.

PR #354, opened 2016-02-09
Moved similar-segment merging control flow and file output into AllelicCNV.

samuelklee:


LeeTL1220:
Look good.  I'd like to put a little less logic in the CLI, but we can reserve that discussion for when we are doing the other refactorings.

PR #361, opened 2016-02-11
Closes #353 and rewrote tests to fail with the old code.

### LINKED ISSUE #353: CombineReadCounts fails when targets' chromosome order is 1, 2 . . . 23
`HashedListTargetCollection` is traversed in the lexicographical order of contigs because its `sortedIntervals` field is initialized via `IntervalUtils.LEXICOGRAPHICAL_ORDER_COMPARATOR`.  This is the order that `CombineReadCounts` uses when it calls `targets.targets()`.  Therefore, if the targets are in the order 1, 2, 3 . . . 23, it goes through chromosome 1, then does `seekTarget` to chromosomes 11, 12, 13, 14 . . . 19.  Next `seekTarget` looks for the first target in chromosome 2, which fails because it jumped past 2, 3, 4 . . . 9 in order to reach chromosome 11 after chromosome 1.

To reproduce this error quickly, run `CombineReadCountsIntegrationTest` but replace `TEST_CHR = new String[] { "1", "2", "3"}` with `TEST_CHR = new String[] { "1", "2", "11"}`.  Note also that it passes for {"1", "11", "2"}.

The root of the issue is that there are two inconsistent target orderings, the lexicographically sorted order of `HashedListTargetCollection` and the order of the original targets file.

davidbenjamin:
@vruano do you have time to review this?

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/CombineReadCounts.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/CombineReadCounts.java
@@ -76,7 +76,7 @@ import java.util.stream.IntStream;
  * </p>
  * <h2>Input and Output File Format</h2>
  * <p>
- *    Input read count file format follows the one produced by {@link ExomeReadCounts}.
+ *    Input read count file format follows the one produced by {@link CalculateTargetBaseCallCoverage}.
  *    Namely a tab separated value table file where there is a least one column, <b>NAME</b>;
  *    indicating the target name, an arbitrary number of read count columns (e.g. one per sample)
  *    and the target coordinates (<b>CONTIG</b>, <b>START</b> and <b>STOP</b>) which are optional if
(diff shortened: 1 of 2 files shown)


vruano:
Yup.

vruano:
Finish with reviews. Check my code proposal. If you think it makes things easier to understand take it on board otherwise you can commit the code as it is. Up to you.

davidbenjamin:
@vruano I like your simplification.  The previous code works even if the input files don't have the same targets, as long as they contain the *specified* targets in the same order.  The proposed code requires all targets, even those not used, to be shared by all inputs.  Are you okay with this?

vruano:
Yes I am, is there any reason why this would not be the case?

davidbenjamin:
@vruano changes based on your suggestion.

davidbenjamin:
@vruano It's a lot prettier now.

vruano:
Well done rebase, squash and merge when test pass again.

PR #362, opened 2016-02-12
Issue 346 acs conversion

### LINKED ISSUE #133: Make segmenter sample name parameter optional
Check that contig start stop name are in the colnames or fail
If only one remaining colname use that as sample name if none specified
If multiple colnames remaining fail if no sample name specified

### LINKED ISSUE #346: Allow ACNV to output an allelic seg file that mimics the legacy format of AllelicCapSeg
Allelic CapSeg format:
```
Chromosome | Start.bp | End.bp | n_probes | length | n_hets | f | tau | sigma.tau | mu.minor | sigma.minor | mu.major | sigma.major | SegLabelCNLOH	
---------- | -------- | ------ | -------- | ------ | ------ | - | --- | --------- | -------- | ----------- | -------- | ----------- | -------------
1 | 2500 | 124997500 | 24999 | 124995000 | 5000 | 0.5 | 2 | 4e-05 | 1 | 0.02 | 1 | 0.02 | 2	
```

ACNV format:
```
Sample  Chromosome      Start   End     Num_Targets     Num_SNPs        Segment_Mean_Post_Mode  Segment_Mean_Post_Lo    Segment_Mean_Post_Hi    MAF_Post_Mode   MAF_Post_Lo     MAF_Post_Hi

```

### LINKED ISSUE #358: Simple MAF == 0.5 caller in ACS output
This caller can be implemented as a threshold or otherwise clunky manner.

In the *.acs.seg file, move the MAF center value to 0.5 if the upper value (of the credible interval) is sufficiently close to 0.5.  Exact procedure is TBD, but simple and easy-to-implement is fine.

LeeTL1220:
- Removes sample name parameter from several tools.
- Produces files compatible with GATK CNV and CGA ACS.
- Simple Balanced Segment Caller for the ACS conversion to peg to MAF 0.5 when likely.

Closes #133 
Closes #346 
Closes #358


src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
```diff
@@ -213,6 +217,30 @@ public static SimpleInterval trimInterval(final Locatable interval,
                 dataLine -> toModeledSegment(dataLine, false));
     }
 
+    /** Read only the sample names from a segment file.
+     *
+     * @param segmentFile Never {@code null}
+     * @return a list of the unique sample names in the seg file.
+     */
+    public static List<String> readSampleNamesFromSegmentFile(final File segmentFile) {
```
samuelklee: This should use a one-liner to read in the list of all names:
`readSegmentFile(segmentFile, new String[]{SegmentTableColumns.SAMPLE.toString()}, SegmentUtils::toSampleName)`
You can then convert this list to a Set of unique names and back to a list, if desired.

When I previously refactored this class, I extracted all of the boilerplate TableReader code, checking of mandatory columns, and exception throwing to the private readSegmentFile method.  This got rid of a lot of copy/pasted code that was common to each readXFile method (where X = Segment, ModeledSegment, etc.)---all of these methods could be turned into one-liners that simply pass the mandatory columns and a lambda for translating the dataLine to X.  Same goes for the writeSegmentFile method.

However, I see that @vruano is organizing IO of germline segments in a different manner.  So we should definitely standardize when refactoring all the segment stuff.  @davidbenjamin

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
```diff
@@ -414,33 +488,44 @@ public int getSite() {
...
-                dataLine.getLong(SegmentTableColumns.NUM_PROBES.toString()),
-                takeLog2 ? ParamUtils.log2(mean) : mean);
     }
 
-    private static <T extends Locatable> List<T> readSegmentFile(final File segmentsFile,
-                                                                 final String[] mandatoryColumns,
-                                                                 final Function<DataLine, T> dataLineToSegmentFunction) {
+    public static <T extends Locatable> List<T> readSegmentFile(final File segmentsFile,
+                                                                final String[] mandatoryColumns,
+                                                                final Function<DataLine, T> dataLineToSegmentFunction) {
```
samuelklee: Can add the

`Utils.nonNull(segmentsFile);
 Utils.regularReadableUserFile(segmentsFile);`

back here.

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
```diff
@@ -302,7 +351,32 @@ public static void writeACNVModeledSegmentFile(final File outFile,
...
+                                        .replaceAll("\\s+", "")));
+    }
+
+    /**
+     * Read ACNV file into ACNV modeled segments.
+     *
+     * @param acnvSegFile File in the ACNV format (*-final.seg).  Must be readable file and never {@code null}
+     * @return Never {@code null}.  Can be empty list if no entries exist in the input file.
+     */
+    public static List<ACNVModeledSegment> readACNVModeledSegmentFile(final File acnvSegFile) {
```
samuelklee: Should also be a one-liner:
`return readSegmentFile(acnvSegFile, SegmentTableColumns.ACNV_MODELED_SEGMENT_COLUMN_NAME_ARRAY, SegmentUtils::toACNVModeledSegment);`

Also it seems cleaner to me to have all of the parameters for these readXFile methods be named "segmentsFile" or something standard.

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
```diff
@@ -302,7 +351,32 @@ public static void writeACNVModeledSegmentFile(final File outFile,
...
                                         .replaceAll("\\s+", ""))
                                 .append(String.format("%6.4f", segment.getMinorAlleleFractionPosteriorSummary().upper())
-                                        .replaceAll("\\s+","")));
+                                        .replaceAll("\\s+", "")));
+    }
+
+    /**
+     * Read ACNV file into ACNV modeled segments.
+     *
+     * @param acnvSegFile File in the ACNV format (*-final.seg).  Must be readable file and never {@code null}
```
samuelklee: Actually, all sim-*.seg files can be read (but I guess we might decide to suppress output of intermediate files in the near future).

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
samuelklee: It also looks like the methods in this class got a bit shuffled out of order.  Could you put all of the public read methods together, all of the public write methods together, etc.?

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
```diff
@@ -414,33 +488,44 @@ public int getSite() {
...
 
     }
 
-    private static ModeledSegment toModeledSegment(final DataLine dataLine, final boolean takeLog2) {
-        final double mean = dataLine.getDouble(SegmentTableColumns.MEAN.toString());
+    /** Simple conversion to an interval given standard segment columns
+     *
+     * @param dataLine line in a file
+     * @param contigColumn name of the contig column
+     * @param startColumn
```
samuelklee: javadoc missing for these two parameters

src/main/java/org/broadinstitute/hellbender/tools/exome/AllelicCNV.java
```diff
@@ -258,13 +259,19 @@ protected void runPipeline(final JavaSparkContext ctx) {
...
         modeller.writeACNVModeledSegmentFile(finalModeledSegmentsFile);
+
         //write file for GATK CNV formatted seg file
         final File finalModeledSegmentsFileAsGatkCNV = new File(outputPrefix + "-" + FINAL_SEG_FILE_TAG + "." + GATK_SEG_FILE_TAG + ".seg");
         modeller.writeModeledSegmentFile(finalModeledSegmentsFileAsGatkCNV);
 
+        // Write file for ACS- compatible output to help Broad CGA
+        final File finalACSModeledSegmentsFile = new File(outputPrefix + "-" + FINAL_SEG_FILE_TAG + "." + CGA_ACS_SEG_FILE_TAG + ".seg");
+        modeller.writeACNVModeledSegmentFileAsAllelicCapSegFile(finalACSModeledSegmentsFile);
+
```
samuelklee: Sample name is still present here (and the test class) and in ExomeStandardArgumentDefinitions...did you want to completely remove it?  Shouldn't affect the eval scripts as the output prefix is used instead.

src/main/java/org/broadinstitute/hellbender/tools/exome/ACNVModeller.java
```diff
@@ -121,7 +123,13 @@ public void writeACNVModeledSegmentFile(final File outFile) {
      * @param outFile   output file
      */
     public void writeModeledSegmentFile(final File outFile) {
-        SegmentUtils.writeModeledSegmentFile(outFile, segments, segmentedModel.getGenome().getSampleName());
+        SegmentUtils.writeModeledSegmentFile(outFile,
+                ACNVModeledSegmentConverter.convertACNVModeledSegmentsToModeledSegments(segments, segmentedModel.getGenome()),
+                segmentedModel.getGenome().getSampleName());
+    }
+
+    public void writeACNVModeledSegmentFileAsAllelicCapSegFile(final File outFile) {
```
samuelklee: This can just be named writeAllelicCapSegFile (since the previous method is just named writeModeledSegmentFile, not writeACNVModeledSegmentFileAsModeledSegmentFile).  The verbose name in ACSModeledSegmentUtils can be kept if you like.

src/main/java/org/broadinstitute/hellbender/tools/exome/acnvconversion/ACNVModeledSegmentConverter.java
```diff
@@ -0,0 +1,35 @@
...
+import org.broadinstitute.hellbender.tools.exome.ModeledSegment;
+import org.broadinstitute.hellbender.utils.Utils;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Utilities for converting ACNV modeled segments.
+ */
+public class ACNVModeledSegmentConverter {
```
samuelklee: Maybe ACNVModeledSegmentConversionUtils instead.  Class can also be final.

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
```diff
@@ -451,6 +536,41 @@ private static ModeledSegment toModeledSegment(final DataLine dataLine, final bo
...
+                toInterval(dataLine),
+                segmentMeanPosteriorSummary,
+                minorAlleleFractionPosteriorSummary);
+    }
+
+    private static String toSampleName(final DataLine dataLine) {
+        return dataLine.get(SegmentTableColumns.SAMPLE.toString());
+    }
+
+
```
samuelklee: whitespace.  Also see comment in ACSModeledSegmentUtils about making writeSegmentFile public.

src/main/java/org/broadinstitute/hellbender/tools/exome/acsconversion/ACSModeledSegmentUtils.java
```diff
@@ -0,0 +1,139 @@
...
+    /**
+     * Writes a list of segments represented by {@link ACNVModeledSegment} to a file with header:
+     * <p>
+     *      Chromosome | Start.bp | End.bp | n_probes | length | n_hets | f | tau | sigma.tau | mu.minor | sigma.minor | mu.major | sigma.major | SegLabelCNLOH
+     * </p>
+     *
+     * This header is produced by the Broad Institute CGA tool, AllelicCapSeg.  Please note that SegLabelCNLOH is not populated properly.
+     */
+    public static void writeACNVModeledSegmentFileAsAllelicCapSegFile(final File outFile,
+                                                                      final List<ACNVModeledSegment> acnvModeledSegments,
```
samuelklee: If writeSegmentFile in SegmentUtils was made public, this code could be reduced to a one-liner (as in the methods calling the public readSegmentFile method above).

src/main/java/org/broadinstitute/hellbender/tools/exome/acsconversion/ACSModeledSegmentUtils.java
```diff
@@ -0,0 +1,139 @@
...
+                                                                      final Genome genome) {
+        Utils.nonNull(genome, "The genome cannot be null.");
+
+        // This was taken since it approximates 1 sigma from 95% confidence intervals.  Then it was made larger at user request.
+        final double divisor = 2.0;
```
samuelklee: extract as a constant with a more descriptive name

src/main/java/org/broadinstitute/hellbender/tools/exome/acsconversion/ACSModeledSegmentUtils.java
```diff
@@ -0,0 +1,139 @@
...
+     * @param genome Populated genome object.  Never {@code null}
+     * @return segment using the CGA ACS model
+     */
+    public static ACSModeledSegment convertACNVSegmentToACSSegment(final ACNVModeledSegment acnvModeledSegment, final double divisor, final Genome genome) {
+
+        final BalancedSegmentCaller balancedSegmentCaller = new SimpleBalancedSegmentCaller();
+        final boolean isSegmentBalanced = balancedSegmentCaller.isSegmentBalanced(acnvModeledSegment);
+
+        final double f = (isSegmentBalanced? 0.5 : acnvModeledSegment.getMinorAlleleFractionPosteriorSummary().center());
+        final double tau = Math.pow(2,acnvModeledSegment.getSegmentMeanPosteriorSummary().center()) * 2.0;
```
samuelklee: whitespace 2,acnvModeledSegment...

src/main/java/org/broadinstitute/hellbender/tools/exome/acsconversion/ACSModeledSegmentUtils.java
```diff
@@ -0,0 +1,139 @@
...
+     * @param divisor used for estimating the sigma for the segments
+     * @param genome Populated genome object.  Never {@code null}
+     * @return segment using the CGA ACS model
+     */
+    public static ACSModeledSegment convertACNVSegmentToACSSegment(final ACNVModeledSegment acnvModeledSegment, final double divisor, final Genome genome) {
+
+        final BalancedSegmentCaller balancedSegmentCaller = new SimpleBalancedSegmentCaller();
+        final boolean isSegmentBalanced = balancedSegmentCaller.isSegmentBalanced(acnvModeledSegment);
+
+        final double f = (isSegmentBalanced? 0.5 : acnvModeledSegment.getMinorAlleleFractionPosteriorSummary().center());
```
samuelklee: whitespace isSegmentBalanced?

src/main/java/org/broadinstitute/hellbender/tools/exome/acsconversion/ACSModeledSegment.java
```diff
@@ -0,0 +1,69 @@
...
+    final double tau;
+    final double sigmaTau;
+    final double muMinor;
+    final double sigmaMinor;
+    final double muMajor;
+    final double sigmaMajor;
+    final int segLabelCNLOH;
+    final int hetCount;
+
+    public ACSModeledSegment(final SimpleInterval interval, final String call, final long targetCount, final double segmentMeanInLogCR, final int hetCount, final double f, final double sigmaTau, final double muMinor, final double sigmaMinor, final double muMajor, final double sigmaMajor, final int segLabelCNLOH) {
```
samuelklee: segmentMeanInLog2CR to be absolutely clear.  Actually, please do a find+replace on "InLogCR", it appears in a few other places.

src/main/java/org/broadinstitute/hellbender/tools/exome/acsconversion/SimpleBalancedSegmentCaller.java
```diff
@@ -0,0 +1,19 @@
...
+    public SimpleBalancedSegmentCaller() {}
+
+    /**
+     * Extremely simple method to classify an ACNVModeledSegment as balanced (i.e. MAF is exactly 0.5) or not.
+     *
+     * @param segment Never {@code null}
+     * @return whether this segment should be considered balanced
+     */
+    public boolean isSegmentBalanced(final ACNVModeledSegment segment) {
+        return segment.getMinorAlleleFractionPosteriorSummary().upper() > 0.49;
```
samuelklee: I think we may want to bring some measure of the width (upper - center, for example) into play here, but let's see how this works out for CGA first.

src/test/java/org/broadinstitute/hellbender/tools/exome/samplenamefinder/SampleNameFinderTest.java
```diff
@@ -0,0 +1,22 @@
...
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import java.io.File;
+import java.util.List;
+
+public class SampleNameFinderTest extends BaseTest {
+    static final File INPUT_DIR = new File("src/test/resources/org/broadinstitute/hellbender/utils/segmenter/input/");
+    static final File INPUT_FILE = new File(INPUT_DIR, "HCC1143_reduced_log.tsv");
+
```
samuelklee: maybe add some tests of exceptions for files with no names, tests that unique names are found when multiple names are present, etc.?

samuelklee:
back to @LeeTL1220  mostly minor, but some comments about using readSegmentFile and writeSegmentFile methods in SegmentUtils and avoiding duplication of code.

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
LeeTL1220: I cannot use readSegmentFile, because I have to output a ``List<T extends Locatable>`` , which String is not.

I could create a more generic readSegmentFile, but I would rather punt until we refactor.  Is that okay?

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
LeeTL1220: Comment 1: Done.

Comment 2: I think the shuffling is only because I had to change a few private's to public, so I moved them appropriately.  Do you see counter-example?

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
samuelklee: Ah right, forgot to mention that you have to change the signature of readSegmentFile to just return List<T>.  I think this is OK.

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/tools/exome/AllelicCNV.java
LeeTL1220: I removed it, but I think that means I need to adjust our workflow scripts.

LeeTL1220: Done (in this repo)

src/main/java/org/broadinstitute/hellbender/tools/exome/ACNVModeller.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/tools/exome/acnvconversion/ACNVModeledSegmentConverter.java
LeeTL1220: No strong opinion.

Done

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
LeeTL1220: whitespace done.

src/main/java/org/broadinstitute/hellbender/tools/exome/acsconversion/ACSModeledSegmentUtils.java
LeeTL1220: You can't really, because segments have to have a sample name and ACS does not want a sample name.

No action.  Punting until refactoring.

samuelklee: hmm, you're right.  you can see that the sampleName is kind of hackily inserted into the DataLine in writeSegmentFile already, so once we clean that up such one-liners will be possible.

src/main/java/org/broadinstitute/hellbender/tools/exome/acsconversion/ACSModeledSegmentUtils.java
LeeTL1220: Want to keep parameter incase we specify later.  Otherwise, done

src/main/java/org/broadinstitute/hellbender/tools/exome/acsconversion/ACSModeledSegmentUtils.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/tools/exome/acsconversion/ACSModeledSegmentUtils.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
samuelklee: both toInterval and readSegmentFile public methods are under PRIVATE METHODS FOR SEGMENT UNION, readACNVModeledSegmentFile is at the end of PUBLIC METHODS FOR READING/WRITING SEGMENT FILES after all of the write methods, instead of before them

src/main/java/org/broadinstitute/hellbender/tools/exome/acsconversion/ACSModeledSegment.java
LeeTL1220: Did the ctl+shift+R.

Done

src/main/java/org/broadinstitute/hellbender/tools/exome/acsconversion/SimpleBalancedSegmentCaller.java
LeeTL1220: Of course, but this was the "simple version"...

No action

LeeTL1220:
@samuelklee Please read my responses, because there were some "no action" comments.

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
samuelklee: You can also remove those lines from the methods that call this if you like.

samuelklee:
I added a few responses while you were addressing the previous comments, can you take a look at those too?

LeeTL1220:
and unit tests failed... need to see why

LeeTL1220:
Unit test should be fixed now.

src/main/java/org/broadinstitute/hellbender/tools/exome/SegmentUtils.java
LeeTL1220: Would rather not, since future methods may rely on this one.

src/main/java/org/broadinstitute/hellbender/tools/exome/acsconversion/ACSModeledSegmentUtils.java
LeeTL1220: Exactly

LeeTL1220:
@samuelklee I responded.  Tell me if you are okay.  I did not make any further changes (other than to fix the unit test)

samuelklee:
there was a comment about shuffling methods into the appropriate sections in SegmentUtils that you might have missed.  otherwise ok!

LeeTL1220:
I found what you meant about shuffled methods (was not public vsprivate, it was more about merging versus IO), @samuelklee ...  If unit tests pass, I am merging.

PR #363, opened 2016-02-12
Command line tool to normalize read counts by samples' average coverage.

### LINKED ISSUE #357: Normalize by samples' average coverage in germline
The germline pipeline calculates average read depths per target per sample and gives this directly to `CalculateCoverageComponents` (which does PCA) without further processing other than filtering.  If samples are sequenced to different average depths they are not comparable and we get garbage results.

The XHMM paper assumes that all samples are sequenced to the same depth -- they had a very strict QC filter that got rid of samples with coverage deviating by a few percent.  This might seem defensible based on the idea that sample depth affects other laboratory conditions and therefore samples differing in this respect should not be put in the same PoN.  However, we put multiple samples on a flow cell, so sample depth could easily vary dramatically without affecting any other conditions, eg loading concentration.

To illustrate the significance of this issue, here are average depths (averaged over targets, not weighted by target size) of the 8 LMM samples we have so far: 587.6815, 664.3898, 548.8378, 512.2434, 458.8918, 641.2119, 402.2741, 746.3032.

davidbenjamin:
Closes #357.


davidbenjamin:
Assigning @LeeTL1220 since @vruano already has two of my PRs.  This addresses an issue (#357) that is blocking the germline evaluation.

src/main/java/org/broadinstitute/hellbender/tools/exome/NormalizeBySampleDepth.java
```diff
@@ -0,0 +1,124 @@
...
+        try {
+            inputCounts = ReadCountCollectionUtils.parse(inputFile);
+        } catch (final IOException e) {
+            throw new UserException.CouldNotReadInputFile(inputFile, "cannot reach or read the input target coverage file");
+        }
+
+        final RealMatrix counts = inputCounts.counts();
+
+        //kludgy way to deep copy input targets list
+        final List<Target> targets = inputCounts.targets().stream()
```
LeeTL1220: ``new ArrayList<>(...)`` makes a copy.  This would be reasonable if you had something that might extend Target, but you want to enforce the copy being a Target.

src/main/java/org/broadinstitute/hellbender/tools/exome/NormalizeBySampleDepth.java
```diff
@@ -0,0 +1,124 @@
...
+        try {
+            ReadCountCollectionUtils.write(outputFile, normalizedCounts);
+        } catch (final IOException e) {
+            throw new UserException.CouldNotCreateOutputFile(outputFile, "cannot  write to the given output file");
+        }
+
+        return "SUCCESS";
+    }
+
+    private static int length(final Target target) { return target.getEnd() - target.getStart() + 1; }
```
LeeTL1220: This should be moved to a utility class or as a static method on Target.

src/main/java/org/broadinstitute/hellbender/tools/exome/NormalizeBySampleDepth.java
```diff
@@ -0,0 +1,124 @@
...
+            fullName = WEIGHTED_AVERAGE_LONG_NAME,
+            shortName = WEIGHTED_AVERAGE_SHORT_NAME,
+            optional = true
+    )
+    protected boolean weightByTargetSize = false;
+
+    @Override
+    public Object doWork() {
+        final ReadCountCollection inputCounts;
+        try {
```
LeeTL1220: The bulk of doWork belongs in another class.  We should get in the habit of separating the CLI (presentation) layer of code from backend classes.  I can explain why this is important in person.  

The general structure should be:
```
@Override
public Object doWork() {
  // read file
  // call backend object to the processing on a ReadCountCollection, which returns the counts or some class of results.
  // write to file
}
```

src/test/java/org/broadinstitute/hellbender/tools/exome/NormalizeBySampleDepthIntegrationTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+import java.io.File;
+import java.io.IOException;
+
+/**
+ * Integration tests for {@link CombineReadCounts}.
+ *
+ * @author David Benjamin &lt;davidben@broadinstitute.org&gt;
+ */
+public class NormalizeBySampleDepthIntegrationTest extends CommandLineProgramTest {
```
LeeTL1220: When you separate out the backend class, consider writing automated test(s) for it.

src/main/java/org/broadinstitute/hellbender/tools/exome/NormalizeBySampleDepth.java
```diff
@@ -0,0 +1,124 @@
...
+    @Argument(
+            doc = "Output coverage file",
+            fullName = StandardArgumentDefinitions.OUTPUT_LONG_NAME,
+            shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME,
+            optional = false
+    )
+    protected File outputFile;
+
+    @Argument(
+            doc = "Weighted average with weights proportional to target sizes",
```
LeeTL1220: ``Weight average with weights proportional to target sizes``

LeeTL1220:
@davidbenjamin Back at you...

src/main/java/org/broadinstitute/hellbender/tools/exome/NormalizeBySampleDepth.java
davidbenjamin: Fixed.

src/main/java/org/broadinstitute/hellbender/tools/exome/NormalizeBySampleDepth.java
davidbenjamin: Fixed.

src/main/java/org/broadinstitute/hellbender/tools/exome/NormalizeBySampleDepth.java
davidbenjamin: Agreed and fixed.

src/main/java/org/broadinstitute/hellbender/tools/exome/NormalizeBySampleDepth.java
davidbenjamin: Fixed.  Still need to write tests for the backend.

src/test/java/org/broadinstitute/hellbender/tools/exome/NormalizeBySampleDepthIntegrationTest.java
davidbenjamin: Done.

src/main/java/org/broadinstitute/hellbender/tools/exome/NormalizeBySampleDepth.java
davidbenjamin: Done.

davidbenjamin:
@LeeTL1220 . . . and back to you.

Issue #355, opened 2016-02-10
Het Pulldown in acnv finds sites on nonhomolgous Y


dlivitz:
I've been running acnv in firecloud on the colocated tcga bams. 

I noticed it called some hetsites on Y - I included the path to my run below. 

I know an easy fix would be to remove the Y chromosome, but I'm worried that other noisy sites are included. 

Let me know if more information will be helpful! 

all run files (what cromwell copied out): 
/xchip/cga_home/dlivitz/dropbox/acnv_y_hets/
normal hets: 
/xchip/cga_home/dlivitz/dropbox/acnv_y_hets/TCGA-GC-A4ZW-01A-11D_TCGA-GC-A4ZW-10A-01D.normal.hets.tsv

LeeTL1220:
@samuelklee @davidbenjamin I'm not sure whether we should be worried about this.  If we want to address this, then this issue should be re-labeled as a bug.  Can we decide tomorrow (Feb 10, 2016)?

LeeTL1220:
```
$ egrep "^Y" /xchip/cga_home/dlivitz/dropbox/acnv_y_hets/TCGA-GC-A4ZW-01A-11D_TCGA-GC-A4ZW-10A-01D.normal.hets.tsv
Y       13496371        28      20
Y       13500604        29      25
Y       13500769        14      5
Y       13524507        13      12
Y       13524664        23      23
Y       13524719        10      3


```

samuelklee:
These don't directly affect the ACNV results because X and Y are dropped (as a result of being dropped by CNV).  Whether or not we want such sites to be emitted by HetPulldown is a separate issue.  As for whether such "noisy" sites are being emitted elsewhere, I'd first want to understand where these reads are coming from...I'll look into it tomorrow!

akiezun:
Did you exclude the PAR ?
https://en.m.wikipedia.org/wiki/Pseudoautosomal_region

On Tuesday, February 9, 2016, samuelklee <notifications@github.com> wrote:

> These don't directly affect the ACNV results because X and Y are dropped
> (as a result of being dropped by CNV). Whether or not we want such sites to
> be emitted by HetPulldown is a separate issue. As for whether such "noisy"
> sites are being emitted elsewhere, I'd first want to understand where these
> reads are coming from...I'll look into it tomorrow!
>
> —
> Reply to this email directly or view it on GitHub
[...]

PR #364, opened 2016-02-12
Added base/mapping quality filters for HetPulldown

samuelklee:
Also removed hetAlleleFraction parameter as it was a holdover from previous versions of HetPulldown where allelic skew was given as input.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/GetHetCoverage.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/GetHetCoverage.java
@@ -28,6 +28,12 @@ public final class GetHetCoverage extends CommandLineProgram {
     protected static final String PVALUE_THRESHOLD_FULL_NAME = "pvalueThreshold";
     protected static final String PVALUE_THRESHOLD_SHORT_NAME = "p";
 
+    protected static final String MINIMUM_MAPPING_QUALITY_SHORT_NAME = "minMQ";
+    protected static final String MINIMUM_MAPPING_QUALITY_FULL_NAME = "minimumMappingQuality";
+
+    protected static final String MINIMUM_BASE_QUALITY_SHORT_NAME = "minBQ";
+    protected static final String MINIMUM_BASE_QUALITY_FULL_NAME = "minimumBaseQuality";
+
     @Argument(
             doc = "BAM file for normal sample.",
             fullName = ExomeStandardArgumentDefinitions.NORMAL_BAM_FILE_LONG_NAME,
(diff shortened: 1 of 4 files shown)


samuelklee:
Evaluation is not quite complete, but it looks like results slightly improve or otherwise do not change drastically.  In any case, it seems that some sort of filtering should be done in this step, so I am of the opinion that we should merge even if concordance with HAPSEG decreases.

This was prompted by #355 but does not completely resolve that issue.  Looks like two hets are consistently pulled down on Y across a few normals...will continue investigating.

LeeTL1220:
Despite the fact that we still see hets on Y, I am assuming that I should still review this PR.

src/main/java/org/broadinstitute/hellbender/tools/exome/GetHetCoverage.java
```diff
@@ -72,14 +69,6 @@
     protected File tumorHetOutputFile;
 
     @Argument(
-            doc = "Heterozygous allele fraction.",
```
LeeTL1220: Does this mean that we need to change the workflow scripts?

src/main/java/org/broadinstitute/hellbender/tools/exome/HetPulldownCalculator.java
```diff
@@ -31,8 +31,12 @@
     private final File refFile;
     private final IntervalList snpIntervals;
 
+    private static final int NUMBER_OF_LOG_UPDATES = 20;
+    private static final double HET_ALLELE_FRACTION = 0.5;
+
     /** Set quality and read-depth thresholds for pulldown, interval threshold for indexing for SamLocusIterator. */
-    private static final int MIN_QUALITY = 0;
+    private static final int MIN_BASE_QUALITY = 20;
```
LeeTL1220: Are you sure that you want these hardcoded and not an advanced parameter with a default value?

src/main/java/org/broadinstitute/hellbender/tools/exome/HetPulldownCalculator.java
```diff
@@ -31,8 +31,12 @@
     private final File refFile;
     private final IntervalList snpIntervals;
 
+    private static final int NUMBER_OF_LOG_UPDATES = 20;
```
LeeTL1220: What is a log update?  Can you add a comment?

LeeTL1220:
Minor comments... Back to you @samuelklee

src/main/java/org/broadinstitute/hellbender/tools/exome/GetHetCoverage.java
samuelklee: yup, should be a minor change.  will issue a PR for it.

src/main/java/org/broadinstitute/hellbender/tools/exome/HetPulldownCalculator.java
samuelklee: done

src/main/java/org/broadinstitute/hellbender/tools/exome/HetPulldownCalculator.java
samuelklee: done, exposed as minMQ and minBQ

samuelklee:
Thanks, @LeeTL1220, I've addressed the PR comments.  Also, evaluation finished on the PRAD samples and it looks like MAE and RMSE either slightly improve or don't change much across various cuts of the samples.  So even though we lose ~5% of the old hets I'm comfortable with this change.

Issue #295, opened 2016-01-15
Avoid counting twice overlapping base calls from the reads that come from the same target.


vruano:
The CalculateTargetCoverage tool can either count reads or base calls (then averaged across position in the target). 

We need to add the option to avoid double counting base calls that come from the same molecule; that is when to read-mates overlap, the overlapped bases should only be counted once.

davidbenjamin:
Closed by PR #365.

PR #365, opened 2016-02-12
Adds the ability to count fragments rather than reads to the Calculate…

vruano:
…TargetBaseCallCoverage tool.

When average-depth is selected this means that the bases of read-pair that overlap are just counted once.
Also, instead of count reads one can now count fragments (read-pairs or a singleton read if the mate is unmapped or mapped somewhere else).

Addresses issue #295


vruano:
@davidbenjamin I divided the single commit between the actual code that does the job and the testing code. That way it looks less intimidating in terms of number of lines.

davidbenjamin:
Back to @vruano

vruano:
I addressed your comments, please re-review.

vruano:
@davidbenjamin

PR #376, opened 2016-02-26
Adds an argument to FilterTargets so that the user can explicitly ind…

vruano:
…icate exclusion intervals.

All targets that overlap any of the intervals indicated (even if it is just by 1bp) will be excluded.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/FilterTargets.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/FilterTargets.java
@@ -1,5 +1,6 @@
 package org.broadinstitute.hellbender.tools.exome;
 
+import htsjdk.samtools.util.Locatable;
 import org.apache.commons.io.output.NullWriter;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
@@ -10,15 +11,14 @@ import org.broadinstitute.hellbender.exceptions.GATKException;
 import org.broadinstitute.hellbender.exceptions.UserException;
 import org.broadinstitute.hellbender.utils.SimpleInterval;
 import org.broadinstitute.hellbender.utils.Utils;
+import org.broadinstitute.hellbender.utils.collections.IntervalsSkipList;
 import org.broadinstitute.hellbender.utils.tsv.DataLine;
 import org.broadinstitute.hellbender.utils.tsv.TableColumnCollection;
 import org.broadinstitute.hellbender.utils.tsv.TableWriter;
 
-import java.io.File;
-import java.io.FileWriter;
-import java.io.IOException;
-import java.io.Writer;
+import java.io.*;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
 import java.util.function.Predicate;
 
@@ -48,7 +48,7 @@ import java.util.function.Predicate;
         oneLineSummary = "filters targets on coverage and annotations",
         programGroup = CopyNumberProgramGroup.class
 )
-public class FilterTargets extends CommandLineProgram {
+public final class FilterTargets extends CommandLineProgram {
 
     public static String REJECT_FILE_TARGET_COLUMN_NAME = "TARGET";
     public static String REJECT_FILE_FILTER_COLUMN_NAME = "FILTER";
@@ -59,7 +59,8 @@ public class FilterTargets extends CommandLineProgram {
      */
     public enum TargetFilter {
         ExtremeTargetSize, ExtremeGCContent, ExtremeRepeatContent,
-        ExtremeCoverageMean, ExtremeCoverageVariance, ExtremeCoverageInterquartileRange;
+        ExtremeCoverageMean, ExtremeCoverageVariance, ExtremeCoverageInterquartileRange,
+        ExcludedInterval
     }
 
     public static final int DEFAULT_MINIMUM_TARGET_SIZE = 0;
@@ -96,6 +97,9 @@ public class FilterTargets extends CommandLineProgram {
 
     public static final String REJECT_OUTPUT_FILE_FULL_NAME = "rejectedOutputFile";
     public static final String REJECT_OUTPUT_FILE_SHORT_NAME = "rejected";
+
+    public static final String EXCLUDED_INTERVALS_FULL_NAME = "excludedIntervals";
+    public static final String EXCLUDED_INTERVALS_SHORT_NAME = "excl";
 
     @Argument(
             doc = "Minimum target size; targets spanning fewer base-pairs are filtered out",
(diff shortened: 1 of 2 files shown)


vruano:
@davidbenjamin please review.

src/main/java/org/broadinstitute/hellbender/tools/exome/FilterTargets.java
```diff
@@ -94,6 +94,9 @@
     public static final String REJECT_OUTPUT_FILE_FULL_NAME = "rejectedOutputFile";
     public static final String REJECT_OUTPUT_FILE_SHORT_NAME = "rejected";
 
+    public static final String EXCLUDED_INTERVAL_FULL_NAME = "excludeInterval";
```
davidbenjamin: Slight inconsistency in names -- there is the constant `EXCLUDED_INTERVAL`, the string literal `"excludeInterval"` and the variable `excludeIntervals`.  My vote for all is `excludedIntervals` and I will comment wherever relevant.

davidbenjamin: `EXCLUDED_INTERVALS_FULL_NAME`

src/main/java/org/broadinstitute/hellbender/tools/exome/FilterTargets.java
```diff
@@ -94,6 +94,9 @@
...
+    public static final String EXCLUDED_INTERVAL_SHORT_NAME = "excl";
```
davidbenjamin: `EXCLUDED_INTERVALS_SHORT_NAME`

src/main/java/org/broadinstitute/hellbender/tools/exome/FilterTargets.java
```diff
@@ -166,6 +169,13 @@
     )
     protected double maximumCoverageVariance = DEFAULT_MAXIMUM_COVERAGE_VARIANCE;
 
+    @Argument(
+            doc = "Targets to explicitly exclude",
+            fullName = EXCLUDED_INTERVAL_FULL_NAME,
+            shortName = EXCLUDED_INTERVAL_SHORT_NAME,
+            optional = true
+    )
+    protected List<String> excludeIntervals;
```
davidbenjamin: `excludedIntervals`

src/main/java/org/broadinstitute/hellbender/tools/exome/FilterTargets.java
```diff
@@ -234,6 +244,9 @@ protected Object doWork() {
                 minimumCoverageVariance > Double.NEGATIVE_INFINITY) {
             result.add(new ExtremeCoverageVariance());
         }
+        if (!excludeIntervals.isEmpty()) {
```
davidbenjamin: `excludedIntervals`

src/main/java/org/broadinstitute/hellbender/tools/exome/FilterTargets.java
```diff
@@ -317,9 +330,159 @@ public String reasonToFilter(final Target target) {
     }
 
     /**
+     * Class that represent a interval of targets to exclude.
```
davidbenjamin: represents an interval

src/main/java/org/broadinstitute/hellbender/tools/exome/FilterTargets.java
```diff
@@ -317,9 +330,159 @@ public String reasonToFilter(final Target target) {
...
+     *
+     * It contains a copy of the interval specification (e.g. "chr1", "chr1:1+", "chr2:100,000,000-200,000,000").
+     *
+     * And the location where it was specified so that we can provide more informative error and target rejection messages.
+     */
+    private final class ExcludeInterval implements Locatable {
```
davidbenjamin: `ExcludedInterval`

src/main/java/org/broadinstitute/hellbender/tools/exome/FilterTargets.java
```diff
@@ -317,9 +330,159 @@ public String reasonToFilter(final Target target) {
...
+        /**
+         * Creates a new exclusion interval.
+         * @param specString the interval specification as accepted by
+         * {@link SimpleInterval#SimpleInterval(String) SimpleInterval(String)}.
+         * @param specLocation where the user indicated the excluded interval.
+         * @throws UserException.BadInput if {@code specString} is not a valid interval specification.
+         * @throws IllegalArgumentException if either {@code specString} or {@code specLocation} is
+         * {@code null}.
+         */
+        public ExcludeInterval(final String specString, final String specLocation) {
```
davidbenjamin: `excludedInterval`

src/main/java/org/broadinstitute/hellbender/tools/exome/FilterTargets.java
```diff
@@ -317,9 +330,159 @@ public String reasonToFilter(final Target target) {
...
+        }
+
+        public String reasonToFilter(final Target target) {
+            return (interval.overlaps(target)) ?
+                    String.format("overlaps with excluded interval (%s) provided by the user (%s)", specString, specLocation) : null;
+        }
+    }
+
+    /**
+     * Evaluates targets versus the excluded intervals list
```
davidbenjamin: "against" instead of "versus"

src/main/java/org/broadinstitute/hellbender/tools/exome/FilterTargets.java
```diff
@@ -317,9 +330,159 @@ public String reasonToFilter(final Target target) {
...
+            }
+            excludedIntervals = new IntervalsSkipList<>(exclusionObjects);
+        }
+
+        /**
+         * Reads the exclusion specs enclosed in a list file.
+         * @param file the file.
+         * @return never {@code null}, perhaps empty.
+         */
+        private Collection<? extends ExcludeInterval> readExcludeIntervalsFromFile(final File file) {
```
davidbenjamin: `readExcludedIntervalsFromFile`

src/main/java/org/broadinstitute/hellbender/tools/exome/FilterTargets.java
davidbenjamin: If I understand correctly, this could be a `SimpleInterval` but you want to store e.g. what file each interval came from for error/output.  Personally, I don't believe that this is worth it.  It complicates the code and the most common use case will be a single exclude file, in which case finding the source is trivial.  If an advanced user wants to exclude several interval files I would rather keep the code simple and let the user grep his targets if he needs to troubleshoot.  What are your thoughts on this?

davidbenjamin:
@vruano Done with review.  The main thing is that I argue for using `SimpleInterval` instead of a richer `ExcludeInterval` class.  Feel free to argue back; I'll probably learn something.

src/main/java/org/broadinstitute/hellbender/tools/exome/FilterTargets.java
vruano: Is done and it works. The code added complexity is minimal. Sure not vital but it does not harm IMO. 
If you insist I will remove; let me know if that is the case.

davidbenjamin: I don't insist.  My goal in code review is to point out other possibilities (even pure opinion) for you to choose from *if you want* i.e. to be an extra set of eyes but not in the sense of "Big Brother is watching."

vruano:
While a user can specify a whole chromosome just with its name, say 'Y', the corresponding SimpleInterval would read 'Y:1-2147483648'; I wanted to keep track of the original spec string and then I added the location to it too.

vruano:
Back to you.

vruano:
Rebased and resolved conflicts.

davidbenjamin:
Looks good.  I will squash and merge.

PR #377, opened 2016-02-28
Added interquartile range to target stats and target filtration.  Closes #375.

### LINKED ISSUE #375: interquartile range for target stats and filtering
We filter targets based on mean coverage (raw, proportional, and tangent normalized at various stages of the pipeline) and on the standard deviation of coverage.  The idea is that some targets are unpredictable and can't be tamed by tangent normalization.  However, standard deviation is very sensitive to outliers.  In particular, if a true event gets into the PoN (perhaps unlikely) or into the germline cohort (this is how the germline pipeline is *designed* to be run) it could skew the standard deviation, leading to the filtering out of its targets.

To avoid this, we should filter based on the interquartile range or some other measure that is robust to outliers. 

I have measured this in R by plotting interquartile range against standard deviation.  The relation is predictably linear with a very small number of outliers where the standard deviation is much larger than expected.  Among these are some of the events in our truth data.  So, at least for the LMM data, there is real evidence that this proposal will improve sensitivity.

davidbenjamin:


davidbenjamin:
@vruano please review.  Your PR on `FilterTargets` was first, so any rebasing hassle in on me.

vruano:
Looks good. Don't worry I will merge mine after, conflicts should be minimal.

PR #379, opened 2016-02-29
Added validation stringency to GetHetCoverage, some refactoring.

samuelklee:


src/main/java/org/broadinstitute/hellbender/tools/exome/GetHetCoverage.java
```diff
@@ -98,16 +99,19 @@
     )
     protected int minimumBaseQuality = 20;
 
+    @Argument(
+            doc = "Validation stringency for all BAM files read by this program.  Setting stringency to SILENT " +
+            "can improve performance when processing a BAM file in which variable-length data (read, qualities, tags) " +
+            "do not otherwise need to be decoded.",
+            common=true)
```
LeeTL1220: What is ``common=true``?

src/main/java/org/broadinstitute/hellbender/tools/exome/HetPulldownCalculator.java
```diff
@@ -113,7 +117,7 @@ public HetPulldownCalculator(final File refFile, final File snpFile,
      * </p>
      * @param baseCounts        map of base-pair counts
      * @param totalBaseCount    total base-pair counts (excluding N, etc.)
-     * @param pvalThreshold     p-value threshold for two-sided binomial test
+     * @param pvalThreshold     p-value threshold for two-sided binomial test (should be in [0, 1], but no check is performed)
```
LeeTL1220: Why not check here?

src/main/java/org/broadinstitute/hellbender/tools/exome/HetPulldownCalculator.java
```diff
@@ -231,7 +236,7 @@ private Pulldown getHetPulldown(final File bamFile, final IntervalList snpInterv
             }
             logger.info("Examined " + totalNumberOfSNPs + " sites.");
             return hetPulldown;
-        } catch (final IOException e) {
+        } catch (final Exception e) {
```
LeeTL1220: This is considered bad form to catch the Exception class...  Can you catch the specific types that are actually thrown?

LeeTL1220:
Comments are minor.  Back to you, @samuelklee

src/main/java/org/broadinstitute/hellbender/tools/exome/GetHetCoverage.java
samuelklee: i just copied the VALIDATION_STRINGENCY argument parameters almost word-for-word from the analogous code in public.  according to the Argument documentation, common=true if "this an Option common to all command line programs.  If it is then it will only be displayed in usage info when H or STDHELP is used to display usage."

once LocusWalker/pileup is implemented, i think we'll be able to remove this duplicated code.

src/main/java/org/broadinstitute/hellbender/tools/exome/HetPulldownCalculator.java
samuelklee: it seemed a bit redundant to check for every het, since it's checked in the true public entry point in the getNormal method.  i'll change this method to VisibleForTesting instead.

src/main/java/org/broadinstitute/hellbender/tools/exome/HetPulldownCalculator.java
samuelklee: done.

samuelklee:
thanks, back to you @LeeTL1220

src/main/java/org/broadinstitute/hellbender/tools/exome/GetHetCoverage.java
```diff
@@ -98,16 +99,19 @@
...
+    protected ValidationStringency VALIDATION_STRINGENCY = ValidationStringency.DEFAULT_STRINGENCY;
```
cmnbroad: FWIW, the gatk tools (the Picard tools are currently an exception) use SILENT stringency as the default (actually, they use the constant defined [here](https://github.com/broadinstitute/gatk/blob/master/src/main/java/org/broadinstitute/hellbender/utils/read/ReadConstants.java#L14).

src/main/java/org/broadinstitute/hellbender/tools/exome/GetHetCoverage.java
```diff
@@ -98,16 +99,19 @@
...
     protected Object doWork() {
-        if (pvalThreshold < 0 || pvalThreshold > 1) {
-            throw new UserException.BadArgumentValue(PVALUE_THRESHOLD_FULL_NAME,
-                    Double.toString(pvalThreshold),
-                    "p-value threshold should be in the [0, 1] range.");
-        }
-
-        final HetPulldownCalculator hetPulldown = new HetPulldownCalculator(REF_ARGUMENTS.getReferenceFile(), snpFile,
-                minimumMappingQuality, minimumBaseQuality);
+        SamReaderFactory.setDefaultValidationStringency(VALIDATION_STRINGENCY);
```
cmnbroad: This is kind of a big hammer to use since it sets global static state. See comments below.

src/main/java/org/broadinstitute/hellbender/tools/exome/HetPulldownCalculator.java
```diff
@@ -52,10 +56,13 @@
      */
     public HetPulldownCalculator(final File refFile, final File snpFile,
                                  final int minMappingQuality, final int minBaseQuality) {
+        ParamUtils.isPositiveOrZero(minMappingQuality, "Minimum mapping quality must be nonnegative.");
+        ParamUtils.isPositiveOrZero(minBaseQuality, "Minimum base quality must be nonnegative.");
         this.refFile = refFile;
         this.snpIntervals = IntervalList.fromFile(snpFile);
         this.minMappingQuality = minMappingQuality;
         this.minBaseQuality = minBaseQuality;
+        this.samReaderFactory = SamReaderFactory.makeDefault();     //will use default validation stringency
```
cmnbroad: Rather than caching the reader factory derived from setting the global state, see the suggested code fragment below.

src/main/java/org/broadinstitute/hellbender/tools/exome/HetPulldownCalculator.java
```diff
@@ -176,7 +185,7 @@ public Pulldown getTumor(final File tumorBAMFile, final IntervalList normalHetIn
      */
     private Pulldown getHetPulldown(final File bamFile, final IntervalList snpIntervals, final SampleType sampleType,
                                     final double pvalThreshold) {
-        try (final SamReader bamReader = SamReaderFactory.makeDefault().open(bamFile);
+        try (final SamReader bamReader = samReaderFactory.open(bamFile);
```
cmnbroad: The canonical form for this would be something like:

bamReader = SamReaderFactory.makeDefault().setValidationStringency(VALIDATION_STRINGENCY).setReference(this.refFile).open(bamFile)

This doesn't rely on any cached global state, and also passes in the reference file, which allows it to work on CRAM files as well.

cmnbroad:
We've been trying to unify some of the patterns used by GATK for various things like validation stringency and bam file handling, so I added a couple of comments on the patterns we're using. Feel free to ignore if they're not appropriate here.

samuelklee:
thanks for the comments, @cmnbroad!  i've added the changes you suggested.

src/main/java/org/broadinstitute/hellbender/tools/exome/GetHetCoverage.java
samuelklee: thanks for pointing this out.  i had actually originally implemented something more along the lines of what you suggested below, but then changed it after seeing how things were done in PicardCommandLineProgram---didn't appreciate that things were being done differently for the gatk tools!

cmnbroad: Yeah, the PicardCommandLineProgram-derived tools are a bit behind the
times, though we'll update those at some point as well. This looks much
better now though!

On Tue, Mar 1, 2016 at 8:44 AM, samuelklee <notifications@github.com> wrote:

> In
> src/main/java/org/broadinstitute/hellbender/tools/exome/GetHetCoverage.java
> <https://github.com/broadinstitute/gatk-protected/pull/379#discussion_r54567512>
> :
>
> > @@ -98,16 +99,19 @@
> >      )
> >      protected int minimumBaseQuality = 20;
> >
> > +    @Argument(
> > +            doc = "Validation stringency for all BAM files read by this program.  Setting stringency to SILENT " +
> > +            "can improve performance when processing a BAM file in which variable-length data (read, qualities, tags) " +
> > +            "do not otherwise need to be decoded.",
> > +            common=true)
> > +    protected ValidationStringency VALIDATION_STRINGENCY = ValidationStringency.DEFAULT_STRINGENCY;
>
> thanks for pointing this out. i had actually originally implemented
> something more along the lines of what you suggested below, but then
> changed it after seeing how things were done in
> PicardCommandLineProgram---didn't appreciate that things were being done
> differently for the gatk tools!
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/broadinstitute/gatk-protected/pull/379/files#r54567512>
> .
>



-- 
Chris Norman
Software Engineer/Temp
Broad Institute of MIT and Harvard
email: cnorman@broadinstitute.org <bhill@broadinstitute.org>

PR #388, opened 2016-03-01
Db boy scout rule

davidbenjamin:


samuelklee:
looks OK to squash and merge (can optimize imports, too), just a couple of extremely minor things.  thanks for going through and cleaning things up!

lbergelson:
> [If hiking along a roadway, stay in a single file line and hike on the left side of the
roadway, facing traffic](http://www.troop0775.com/Rules%20for%20Safe%20Hiking.pd)

samuelklee:
"DO NOT HITCHHIKE. This is dangerous and illegal. It is also cheating, it called hiking,
not riding."

PR #389, opened 2016-03-01
Command line tool to convert read counts into target-specific Z scores

davidbenjamin:
@vruano Could you review this?  It passes locally but Travis is taking forever.


vruano:
Back to @davidbenjamin

PR #392, opened 2016-03-02
Added a section about making the shadowJar. Also, removed the redunda…

mbabadi:
…nt copy of the R install script.


LeeTL1220:
@mbabadi Looks good -- merge away!

PR #405, opened 2016-03-07
Added normal-only mode to GetHetCoverage.

samuelklee:
To be used for allelic-PoN exploratory work where the tumor pulldown is not needed.


samuelklee:
here's a quick one for you, @LeeTL1220, thanks!

LeeTL1220:
@samuelklee  Missed this one.  I swear I reviewed it, but it looks good to me as is.

PR #408, opened 2016-03-11
Removed duplicate reader method

mbabadi:
Just practicing Git :) I noticed a duplicate method definition in TableUtils and removed it.


mbabadi:
Ouch, my bad! :-)

PR #410, opened 2016-03-13
More principled germline HMM transition matrix.  Closes #372.

### LINKED ISSUE #372: Improve germline CNV HMM arguments
XHMM was written in one-off style for a specific bait set and contains parameters that are defined in a bait-specific way.  For example `meanTargetsPerSegment` and `meanTargetsPerSegment` depend on the density of targets used.  This is unnecessary because any such prior expectations about the quantity and size of CNVs are biological properties independent of the technology.

davidbenjamin:
Tested this change in the eval repo and it beats the old version.  I'm making this change even while trying to get an evaluated product out as fast as possible because the search over parameter space is too slow.

It's not just that XHMM has three transition matrix parameters and this brings it down to two.  Rather, it's that XHMM has three parameters that *change depending on the targets* and therefore can't really be defended and furthermore are highly prone to overfitting, whereas the two new parameters are biological constants.  Thus a time-consuming 3-D grid search guaranteed to overfit the model is replaced by *nothing*.

@vruano can you take this one?


vruano:
What is the gain by catching the matrices based on distance? Apart from the MAX_INT, is there a considerable number of cache hits?

davidbenjamin:
@vruano there are few cache hits (although it would be an excellent approximation to round distances to the nearest thousand or hundred bases in which case the caching speedup would be enormous) in terms of repeated distances.  However, there are 9 (3 from states x 3 to states) cache hits for each transition -- one matrix covers all transitions and you don't want to recalculate it.  Maybe "cache" is a misleading word for this behavior.  What do you think?

vruano:
Done with the review. Nice change in parametrization, it makes lots of sense to drop the number-target + av. intra-event target distance thing.

About the cache, I don't know it would make much of a difference...thinking more about it since we are using the repetitively per each sample then I guess there can be a nominal gain. Do you notice the difference if you deactivate it, (clear it between samples)? Seldomly consecutive targets pairs may share the exact distance in bp, although I might be wrong here. 

Since you are always using the log value of those transition, I would cache the log values so that you save a repetitive log operation; since we are caching them we can go all the way. 
...

vruano:
Back to you @davidbenjamin

davidbenjamin:
@vruano I made the changes.  The effect of caching between samples is quite small.  Evidently the transition probabilities are not the rate-limiting step.  Still, I think it's worth keeping because it's asymptotically better and more elegant.

vruano:
Looks fine although I would never had add 'Log' particle to the class names. I would had renamed ```get(A, B)``` to ```logProbability(A, B)``` or ```getLogProbability(A, B)```.... Maybe tomorrow someone likes the probs not to be logged and you just need to provide a convenient method beside ```getLogProbability``` called ```getProbability```.

Up to you, you can merge this if you want.

davidbenjamin:
@vruano I trust your judgment and will name things as you suggested before emerging.

PR #412, opened 2016-03-16
Update build.gradle for compatibility with gradle 2.11

mbabadi:
1.2.3 introduces compatibility with Gradle 2.11+. Gradle 2.11+ builds fail with shadowJar 1.2.2.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
 
     dependencies {
         classpath 'org.kt3k.gradle.plugin:coveralls-gradle-plugin:2.3.1'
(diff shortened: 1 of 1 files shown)


LeeTL1220:
@mbabadi I am assuming that I should not review since the unit tests are failing.

LeeTL1220:
@mbabadi Not sure if the unit test error is due to changes you made.  Might want to check with @lbergelson  ...

samuelklee:
that SparkSVD bug has been cropping up for a while now (note that master is currently failing because of it...)

samuelklee:
actually, sorry---it's a different test from the one that was previously failing.

PR #416, opened 2016-03-24
Docs for generative coverage model

davidbenjamin:
@samuelklee @mbabadi @LeeTL1220 here it is.  How about I present it at Wednesday's CNV meeting before code review?  It's not like I would have any time to implement it before then.


samuelklee:
thanks for working this out!  would be glad to hear about it on wednesday.

LeeTL1220:
Yes, let's talk about it on Wednesday.

davidbenjamin:
@samuelklee @LeeTL1220 @mbabadi Who's in the mood for some math?  BTW, Bishop Chapter 12 on probabilistic PCA and factor analysis is very similar.

davidbenjamin:
@samuelklee Forgive the frailty of my memory if we did not give this to you at standup yesterday.

samuelklee:
yup, but i think we agreed that @mbabadi should feel free to also leave comments!

docs/CNVs/CNV-methods.tex
```diff
@@ -482,6 +497,134 @@ \subsection{Using Panel of Normals for Allelic Fraction Model} \label{allelic-Po
...
+P(\vc, \vz | \vc, \vW, \vm, \vPsi, d, \vn) = \prod_s {\rm N}(\vz_s | \vzero, \vI) {\rm N}(\vm_s | \vW \vz_s + \vm, \vPsi + \vSigma_s)
+\end{equation}
+%
+
+For fixed $\vc$ we may obtain maximum likelihood estimates of $\vW$, $\vSigma$, and $\vm$ by an iterative EM approach in which we alternate between computing the Gaussian posteriors of each $\vz_s$ (E step) and maximizing the log-likelihood with respect to $\vW$, $\vSigma$, and $\vm$ (M step).  This is similar to the factor analysis model discussed in Chapter 12 of Bishop.  We note that a second level of iteration, on $\vc$, allows us to correct for CNV events that may be present in the panel of normals: alternate between 1) learning $\vW$, $\vSigma$, and $\vm$ and 2) calling $\vc_s$ for each sample in the panel.
+
+The E step follows from substituting $\vPsi \rightarrow \vPsi_s \equiv  \vPsi + \vSigma_s$ in Bishop's Equations 12.66-67 to get
+%
+\begin{align}
+\vG_s =& \left( \vI + \vW^T \vPsi_s W \right)^{-1} \\
```
samuelklee: \vW

docs/CNVs/CNV-methods.tex
```diff
@@ -482,6 +497,134 @@ \subsection{Using Panel of Normals for Allelic Fraction Model} \label{allelic-Po
...
+E \left[ \vz_s \right] =& \vG_s W^T \vPsi_s (\vm_s - \vm) \\
```
samuelklee: same

docs/CNVs/CNV-methods.tex
```diff
@@ -482,6 +497,134 @@ \subsection{Using Panel of Normals for Allelic Fraction Model} \label{allelic-Po
...
+
+The M step equation for the row $\vW_t$ is also routine and gives
+%
+\begin{equation}
+\vW_t^T = \left( \sum_s \vPsi_{st}^{-1}  E[\vz_{s} \vz_{s}^T]  \right)^{-1} \sum_s \vPsi_{st}^{-1} ( \vm_{st} - \vm_t) E[\vz_{s}]
+\end{equation}
+
+We may automatically select the number of latent features to use via an automatic relevance determination (ARD) prior on $\vW$.  That is, we add a term
+\begin{equation}
+P(\vW) = \prod_k \left( \frac{\alpha_k}{2 \pi} \right)^{T/2} \exp \left( -\frac{1}{2} \alpha_k \sum_t W_{tk}^2  \right)
```
samuelklee: same here and below (since elsewhere you use \bf for matrix elements)

docs/CNVs/CNV-methods.tex
```diff
@@ -482,6 +497,134 @@ \subsection{Using Panel of Normals for Allelic Fraction Model} \label{allelic-Po
...
+%
+\begin{equation}
+n_{st} \sim {\rm Poisson}(d_s c_{st} e^{b_{st}})
+\end{equation}
+%
+We can achieve many of the goals listed above by performing probabilistic PCA not on $\vn$ directly but rather on $\vb$.  This is one reason why we choose a Poisson parameter proportional to $e^b$ and not simply $b$ -- probabilistic PCA involves a normal distribution on $\vb_s$, whereas the Poisson parameter must be positive.  Furthermore, since PCA is a linear model with additive effects of different latent features, using $e^b$ represents multiplicative effects, which is more plausible.  Our model is
```
samuelklee: i think this is reasonable, but i think we could explore the possibility of using a Poisson parameter proportional to b and a conjugate prior on b, with the Laplace approximation coming in at the end as we discussed.  or perhaps some combination of the two to capture both multiplicative and additive effects?

docs/CNVs/CNV-methods.tex
```diff
@@ -482,6 +497,134 @@ \subsection{Using Panel of Normals for Allelic Fraction Model} \label{allelic-Po
...
+where $\vz_s \in \mathbb{R}^D$ is a low-dimensional latent vector of laboratory conditions, $\vW \in \mathbb{R}^{T \times D}$ is a linear map from latent space to target space, $\vm \in \mathbb{R}^T$ is the vector of mean biases, and $\vPsi \in \mathbb{R}^{T \times T}$ is a diagonal matrix of residual variance not explained by the latent features.  We approximate the Poisson as a Gaussian centered at its mode (as a function of $b_{st}$):
+%
+\begin{align}
+\prod_t {\rm Poisson}(n_{st} | d_s c_{st} \lambda_{st}) \approx {\rm N}(\vb_s | \vm_s, \vSigma_s), \, m_{st} = \ln(n_{st}/(c_{st} d_s)), \, \vSigma_{st} = 1/n_{st}
+\end{align}
+%
+to obtain the joint likelihood
+%
+\begin{equation}
+P(\vn_s, \vb, \vz | \vc, \vW, \vm, \vPsi, d ) = \prod_s {\rm N}(\vz_s | \vzero, \vI) {\rm N}(\vb_s | \vW \vz_s + \vm, \vPsi) {\rm N}(\vb_s | \vm_s, \vSigma_s)
```
samuelklee: should this be the likelihood for \vn?

docs/CNVs/CNV-methods.tex
```diff
@@ -482,6 +497,134 @@ \subsection{Using Panel of Normals for Allelic Fraction Model} \label{allelic-Po
...
+\end{equation}
+%
+
+We can integrate out $\vb$ via the identity $\int {\rm N}(x | \mu_1, \sigma_1^2) {\rm N}(x | \mu_2, \sigma_2^2) \, dx = {\rm N}(\mu_1 | \mu_2, \sigma_1^2 + \sigma_2^2)$ to obtain
+%
+\begin{equation}
+P(\vc, \vz | \vc, \vW, \vm, \vPsi, d, \vn) = \prod_s {\rm N}(\vz_s | \vzero, \vI) {\rm N}(\vm_s | \vW \vz_s + \vm, \vPsi + \vSigma_s)
```
samuelklee: also not sure i understand where \vc and \vn should go on the LHS?

docs/CNVs/CNV-methods.tex
samuelklee: also, in Bishop \vPsi^{-1} appears in these equations instead?

docs/CNVs/CNV-methods.tex
```diff
@@ -482,6 +497,134 @@ \subsection{Using Panel of Normals for Allelic Fraction Model} \label{allelic-Po
...
+\vG_s =& \left( \vI + \vW^T \vPsi_s W \right)^{-1} \\
+E \left[ \vz_s \right] =& \vG_s W^T \vPsi_s (\vm_s - \vm) \\
+E \left[ \vz_s \vz_s^T \right] =& \vG_s + E[\vz_s] E[\vz_s]^T
+\end{align}
+%
+
+For the M steps, we must maximize with respect to 
+%
+\begin{align}
+E \left[ \ln {\rm N}(\vm_s | \vW \vz_s + \vm, \vPsi + \vSigma_s) \right] =& -\frac{1}{2} \sum_{st} \left[ \ln \vPsi_{st}+ E[(\vW \vz_s + \vm - \vm_s)_t^2] (\vPsi_{st})^{-1} \right]
```
samuelklee: is there a sum over s missing on LHS?

docs/CNVs/CNV-methods.tex
```diff
@@ -482,6 +497,134 @@ \subsection{Using Panel of Normals for Allelic Fraction Model} \label{allelic-Po
...
+E \left[ \ln {\rm N}(\vm_s | \vW \vz_s + \vm, \vPsi + \vSigma_s) \right] =& -\frac{1}{2} \sum_{st} \left[ \ln \vPsi_{st}+ E[(\vW \vz_s + \vm - \vm_s)_t^2] (\vPsi_{st})^{-1} \right] 
+\end{align}
+which reduces to
+\begin{equation}
+ -\frac{1}{2} \sum_{st} \left[ \ln (\vPsi_t + \Sigma_{st}) + \left( \vW_t  E[\vz_s \vz_s^T] \vW_t^T + (\vm_t - \vm_{st})^2 +2 (\vm_t - \vm_{st}) \vW_t E[\vz_s] \right)\vPsi_{st}^{-1} \right]
+\end{equation}
+%
+
+
+The M step for $\vm$ is routine and comes out to:
```
samuelklee: is there a missing minus sign here?

docs/CNVs/CNV-methods.tex
```diff
@@ -482,6 +497,134 @@ \subsection{Using Panel of Normals for Allelic Fraction Model} \label{allelic-Po
...
+%
+\begin{equation}
+\vW_t^T = \left( \sum_s \vPsi_{st}^{-1}  E[\vz_{s} \vz_{s}^T]  \right)^{-1} \sum_s \vPsi_{st}^{-1} ( \vm_{st} - \vm_t) E[\vz_{s}]
+\end{equation}
+
+We may automatically select the number of latent features to use via an automatic relevance determination (ARD) prior on $\vW$.  That is, we add a term
+\begin{equation}
+P(\vW) = \prod_k \left( \frac{\alpha_k}{2 \pi} \right)^{T/2} \exp \left( -\frac{1}{2} \alpha_k \sum_t W_{tk}^2  \right)
+\end{equation}
+The E step is unaffected by this addition and the only change to the M step is that in Equation \ref{M_step_W} we now have
```
samuelklee: broken \ref (or maybe need to rebuild PDF)

samuelklee:
done with my review, just a few possible minor typos.  i think this looks great and am excited to see it implemented!

passing on to @mbabadi in case he wants to do a run through.

mbabadi:
Great, I'll do a run through tomorrow.

docs/CNVs/CNV-methods.tex
davidbenjamin: fixed both

docs/CNVs/CNV-methods.tex
davidbenjamin: fixed

docs/CNVs/CNV-methods.tex
davidbenjamin: Un-using \bf for matrix elements.

docs/CNVs/CNV-methods.tex
davidbenjamin: I have thought a lot about trying to make this work.  Zoubin Ghahramani has a paper on conjugate exponential generalizations to PPCA but it doesn't quite fit what we need.  If one of us is able to think of a model along these lines I would be very interested in replacing the Gaussian-only model.

docs/CNVs/CNV-methods.tex
davidbenjamin: fixed

docs/CNVs/CNV-methods.tex
davidbenjamin: fixed, I think

docs/CNVs/CNV-methods.tex
davidbenjamin: fixed

docs/CNVs/CNV-methods.tex
davidbenjamin: If I understand what you're referring to, it's implicit matrix multiplication.

docs/CNVs/CNV-methods.tex
davidbenjamin: fixed

docs/CNVs/CNV-methods.tex
samuelklee: sorry, what i meant is that shouldn't there be an overall minus sign in the expression for \vm?  that is, should it be \vm_{s} - \vW  E[\vz_{s}]?

davidbenjamin: fixed.

davidbenjamin:
@mbabadi Is this back to me?

docs/CNVs/CNV-methods.tex
```diff
@@ -482,6 +497,135 @@ \subsection{Using Panel of Normals for Allelic Fraction Model} \label{allelic-Po
...
+\begin{align}
+\vz_s \sim& {\rm N}(\vzero, \vI) \\
+\vb_s \sim& {\rm N}(\vW \vz_s + \vm, \vPsi),
+\end{align}
+%
+where $\vz_s \in \mathbb{R}^D$ is a low-dimensional latent vector of laboratory conditions, $\vW \in \mathbb{R}^{T \times D}$ is a linear map from latent space to target space, $\vm \in \mathbb{R}^T$ is the vector of mean biases, and $\vPsi \in \mathbb{R}^{T \times T}$ is a diagonal matrix of residual variance not explained by the latent features.  We approximate the Poisson as a Gaussian centered at its mode (as a function of $b_{st}$):
+%
+\begin{align}
+\prod_t {\rm Poisson}(n_{st} | d_s c_{st} \lambda_{st}) \approx {\rm N}(\vb_s | \vm_s, \vSigma_s), \, m_{st} = \ln(n_{st}/(c_{st} d_s)), \, \vSigma_{st} = 1/n_{st}
+\end{align}
```
mbabadi: Why \prod_t on the l.h.s?

docs/CNVs/CNV-methods.tex
```diff
@@ -482,6 +497,135 @@ \subsection{Using Panel of Normals for Allelic Fraction Model} \label{allelic-Po
...
+%
+to obtain the joint likelihood
+%
+\begin{equation}
+P(\vn, \vb, \vz | \vc, \vW, \vm, \vPsi, d ) = \prod_s {\rm N}(\vz_s | \vzero, \vI) {\rm N}(\vb_s | \vW \vz_s + \vm, \vPsi) {\rm N}(\vb_s | \vm_s, \vSigma_s)
```
mbabadi: Should d be bold?

docs/CNVs/CNV-methods.tex
```diff
@@ -482,6 +497,135 @@ \subsection{Using Panel of Normals for Allelic Fraction Model} \label{allelic-Po
...
+%
+\begin{equation}
+n_{st} \sim {\rm Poisson}(d_s c_{st} e^{b_{st}})
+\end{equation}
+%
+We can achieve many of the goals listed above by performing probabilistic PCA not on $\vn$ directly but rather on $\vb$.  This is one reason why we choose a Poisson parameter proportional to $e^b$ and not simply $b$ -- probabilistic PCA involves a normal distribution on $\vb_s$, whereas the Poisson parameter must be positive.  Furthermore, since PCA is a linear model with additive effects of different latent features, using $e^b$ represents multiplicative effects, which is more plausible.  Our model is
+%
+\begin{align}
+\vz_s \sim& {\rm N}(\vzero, \vI) \\
+\vb_s \sim& {\rm N}(\vW \vz_s + \vm, \vPsi),
```
mbabadi: I am confused -- by subscript s in z and b; do you imply that we have different latent random variables z_s and b_s for each sample? (I don't think so, in which case, doesn't it make sense to drop s?)

docs/CNVs/CNV-methods.tex
mbabadi: I can not think of any physical source for additive bias for the Poisson parameter: capture, mapping, and PCR biases are all multiplicative, right?

mbabadi: Naive question: what is the fundamental advantage of doing PCA over let's say, a naive MLE of the multiplicative bias and copy number at each target? we still get a super simple generative model (Eq. 42) without any latent variables. By doing PCA, we are tacitly assuming that the biases at different targets are strongly correlated and can be described via an affine mapping from a lower dimensional manifold. Is this a good assumption?  Are we doing PCA because we do not have enough data to infer the multiplicative bias at each target separately, or do we gain something more profound from PCA? educate me please, I'm sure these questions stem from my lack of insight.

docs/CNVs/CNV-methods.tex
```diff
@@ -482,6 +497,135 @@ \subsection{Using Panel of Normals for Allelic Fraction Model} \label{allelic-Po
...
+\end{align}
+%
+where $\vz_s \in \mathbb{R}^D$ is a low-dimensional latent vector of laboratory conditions, $\vW \in \mathbb{R}^{T \times D}$ is a linear map from latent space to target space, $\vm \in \mathbb{R}^T$ is the vector of mean biases, and $\vPsi \in \mathbb{R}^{T \times T}$ is a diagonal matrix of residual variance not explained by the latent features.  We approximate the Poisson as a Gaussian centered at its mode (as a function of $b_{st}$):
+%
+\begin{align}
+\prod_t {\rm Poisson}(n_{st} | d_s c_{st} \lambda_{st}) \approx {\rm N}(\vb_s | \vm_s, \vSigma_s), \, m_{st} = \ln(n_{st}/(c_{st} d_s)), \, \vSigma_{st} = 1/n_{st}
```
mbabadi: You have not defined \lambda_{st} \equiv e^{b_{st}} before.

docs/CNVs/CNV-methods.tex
```diff
@@ -482,6 +497,135 @@ \subsection{Using Panel of Normals for Allelic Fraction Model} \label{allelic-Po
...
+%
+
+For the M steps, we must maximize with respect to 
+%
+\begin{align}
+E \left[ \sum_s \ln {\rm N}(\vm_s | \vW \vz_s + \vm, \vPsi + \vSigma_s) \right] =& -\frac{1}{2} \sum_{st} \left[ \ln \Psi_{st}+ E[(\vW \vz_s + \vm - \vm_s)_t^2] (\Psi_{st})^{-1} \right] 
+\end{align}
+which reduces to
+\begin{equation}
+ -\frac{1}{2} \sum_{st} \left[ \ln (\vPsi_t + \Sigma_{st}) + \left( \vW_t  E[\vz_s \vz_s^T] \vW_t^T + (\vm_t - \vm_{st})^2 +2 (\vm_t - \vm_{st}) \vW_t E[\vz_s] \right)\Psi_{st}^{-1} \right]
```
mbabadi: (checked up to this point)

docs/CNVs/CNV-methods.tex
davidbenjamin: @mbabadi I'm strongly inclined to think everything is multiplicative, which is why I opted for `n ~ Poisson(. . . exp(b))`.  I wouldn't bet the farm on it, but it seems like a safe assumption.

davidbenjamin: @mbabadi The biases of different targets are very strongly correlated.  The latent factors are real physical properties.  For example, in the LMM gene panel data, which is illustrative because its high depth implies relatively little variance comes from Poisson noise, 85% of the variance is explained by the first 5 principal components.

To be clear about something, each target's *average* bias depends on the details of its particular bait and need not have much to do with any other target.  However, PCA captures the *fluctuations* of bias relative to the average, and these fluctuations depend on physical characteristics such as GC content that are shared among different targets.

davidbenjamin: That being said, the linear assumption of PCA / factor analysis is definitely not the whole story.  GC bias is a great example, which is why it should be modeled as an explicit nonlinear effect.  Ultimately the right thing to do might be to discover a latent feature vector for each target and sample-specific response curves (akin to the familiar GC bias curve) for each feature.

samuelklee: I'm not sure about this, but couldn't mapping bias contribute some sort of additive component in regions of high similarity?  These regions might have a noisy floor of stray reads that is proportional to the average depth.  Anyway, I think starting with multiplicative biases is fine, but we could expand to a more general model in the future if we can make it tractable!

mbabadi: @samuelklee, I don't think low complexity regions will get an additive bias due to mapping uncertainty: since the total number of reads to be mapped is "conserved", if region A is similar to region B, the reads will be aligned let's say with equal probability to to either region. This results in a heightened level of statistical fluctuation in the Poisson parameter, but this can be captured by \Psi in Eq. (44), right?

mbabadi: @davidbenjamin, doesn't it make sense to assume that the Poisson parameter is also linearly proportional to the target length? in principle, $\mathbf{m}$ will be able to represent this type of bias, but any reason why we do not to include this explicitly in Eq. (42)?

docs/CNVs/CNV-methods.tex
mbabadi: I second the minus sign error. @davidbenjamin, did you commit changes after @samuelklee's review?

docs/CNVs/CNV-methods.tex
```diff
@@ -482,6 +497,135 @@ \subsection{Using Panel of Normals for Allelic Fraction Model} \label{allelic-Po
...
+The M step for $\vm$ is routine and comes out to:
+\begin{align}
+\vm = \left( \sum_s \vPsi_{s}^{-1} \right)^{-1} \sum_s \left[ \vPsi_{s}^{-1}  (\vW  E[\vz_{s}] - \vm_{s} ) \right] 
+\end{align}
+
+The M step for $\vPsi_t$ is
+\begin{align}
+\sum_s \frac{1}{\vPsi_t + \Sigma_{st}} - \frac{\left( \vW_t  E[\vz_s \vz_s^T] \vW_t^T + (\vm_t - \vm_{st})^2 +2 (\vm_t - \vm_{st}) \vW_t E[\vz_s] \right)}{(\vPsi_t + \Sigma_{st})^2} = 0,
+\end{align}
+which is straightforward to solve numerically.  Alternatively, we could approximate $\vSigma_s \approx \bar{\vSigma} \equiv \frac{1}{S} \sum_s \vSigma_s$, in which case we obtain
```
mbabadi: Since we are working with raw read counts, \Sigma_s will be inversely proportional to the average read depth of the sample in question. If d_s changes significantly from sample to sample (does it?), averaging over s may not be a good idea. I suggest using Eq. (55) as a guess, followed by a few Brent iterations until convergence of Eq. (54).

docs/CNVs/CNV-methods.tex
```diff
@@ -482,6 +497,135 @@ \subsection{Using Panel of Normals for Allelic Fraction Model} \label{allelic-Po
...
+\end{align}
+If we posterior is tightly peaked about the MLE estimate $\vW$, this has the solution
+\begin{equation}
+\alpha_k = \frac{T}{\sum_t W_{tk}^2}
+\end{equation}
+The meaning of these parameters is that $\alpha_k \rightarrow \infty$ sets the $k$-th column of $\vW$ to zero, turning off the $k$-th latent feature.  Thus we can initially choose a liberal estimate of $K$ and the model will automatically become more parsimonious.
+
+We can apply the parameters learned from the panel of normals to single-sample calling, which requires the likelihood as a function of the copy numbers $\vc$.  Applying the same E step as above, the likelihood is
+\begin{equation}
+P({\rm data} | \vc) \propto \prod_t \exp \left[ -\frac{1}{2} \left(  (\vm_t - \vm_{st})^2 +2 (\vm_t - \vm_{st}) \vW_{t} E[\vz_{s}] \right)\Psi_{st}^{-1} \right],
```
mbabadi: Do you suggest intertwining PON creation and CNV calling into a single workflow? i.e. estimate PON (W, \Sigma, m) by via EM (z latent, c_st a fixed hyper parameter), followed by estimating c_st by keeping the PON fixed (via HMM, CBS, maximum likelihood on Eq. 63, etc), and continue until convergence? I can think of two caveats:

(1) Perhaps it is going to be very time consuming compared to our current one-pass scheme (PON creation -> CNV calling)
(2) c_st is not treated quite as a latent variable, but as a hyper parameter determined by an auxiliary segmentation step. Do we have a guarantee that the combined iterations of PCA-EM and CNV calling actually converges? in other words, do we have a guarantee that P(n, z, c | ...) increases after each pass (as in the usual EM)? perhaps the answer is positive (e.g. if the CNV calling increases Eq. 63), but I can not quite see it.

docs/CNVs/CNV-methods.tex
davidbenjamin: Committed but perhaps I didn't push the changes yet.  You're both right.

docs/CNVs/CNV-methods.tex
davidbenjamin: About the multiplicative vs additive mapping effect, let's think. . . Suppose there is some latent factor that varies from sample to sample and target to target.  If this factor is badness of the aligner  used (the target-specific part being difficulty of alignment) and the response to this factor is an error rate proportional to the badness then I think we get @samuelklee's additive effect -- the amount of reads dumped *by* a target is proportional to its coverage, hence multiplicative, but the amount dumped *on* a target is proportional to the overall depth.  I could also imagine the scenario @mbabadi is talking about, and I agree about the conservation law in that case.

Regardless, I think we all agree that this is for the future.

davidbenjamin: @mbabadi About the target length, wouldn't explicitly un-absorbing the length from the target mean be mathematically equivalent -- just a rescaling of the mean?  Another thing is that the general trend of coverage versus length is not linear, because the issue of designing a bait is length-dependent.

docs/CNVs/CNV-methods.tex
davidbenjamin: fixed

docs/CNVs/CNV-methods.tex
davidbenjamin: I think it should be kept -- each sample has its own position in latent space, if that's what you're referring to.

docs/CNVs/CNV-methods.tex
davidbenjamin: Whoops on the lambda; fixed.  As for the normal approximation to the Poisson, it's actually a lot better than that.  As a function of n, you need n >> 1, but n is a constant.  We need the approximation to have the right dependence on b, for fixed n, and the normal approximation is quite good for that purpose even down to n = 1.  (n = 0 is pathological and inelegant but not hard to deal with if you have an extra prior on b that vanishes as |b| becomes large).

docs/CNVs/CNV-methods.tex
davidbenjamin: I had in mind that this approximation would be an initial proof-of-concept implementation and, as you said, a quick initialization.  I put in an estimate of the run time and the exact version will be slower but reasonable even on a single core.

docs/CNVs/CNV-methods.tex
davidbenjamin: 1)  One doesn't have to mix PoN creation with CNV calling -- you could just assume c_st = 2 as we do now and accept that germline events will get in the PoN.  My hunch, which I can't prove until the code actually runs, is that a single iteration of germline calling (which is quite fast) on the 0th-order PoN will yield most of the potential benefits.
2) (*I hope I'm getting this right*) If we consider it as a joint model of c, W, m, Psi, and z (after marginalizing b), then the M step maximizes W, m, and Psi and the E step gets the moments of the posterior on z.  Since the Viterbi algorithm is max-likelihood, it seems to me that using it to make calls on c is on equal footing as part of the M step provided that the HMM likelihood uses the proper prescription where likelihood -> exp( E[log likelihood] wrt posterior on z).  In the latest commit it seems to me that Eq 64 is doing exactly that.

davidbenjamin:
@mbabadi Back to you.

docs/CNVs/CNV-methods.tex
mbabadi: I think I'm being a bit picky here -- but we are just declaring the priors on $\vz$ and $\bz$ here, which is the same for all samples. I'd reserve the notation $\vs_s$ and $\bz_s$ for posterior expectations (as in Bishop).

docs/CNVs/CNV-methods.tex
mbabadi: Sure, but my rough idea is the following: IF target lengths have a relatively broad distribution and IF the coverage is a monotonically increasing function of the target length, we would be spending one precious principal component just to represent this well-defined variation :D so why not get rid of it from the outset? but if the relation between target coverage and target length is not linear, then there is no point doing this (why wouldn't it be anyway? if target length >> short read length ~ 100bp, wouldn't the number of reads that align inside the target interval be proportional to its length?)

docs/CNVs/CNV-methods.tex
```diff
@@ -482,6 +498,149 @@ \subsection{Using Panel of Normals for Allelic Fraction Model} \label{allelic-Po
...
...culating all $\beta_{st}$.  In the approximate version we only need the \textit{sum} $\sum_s \beta_{st}$ , which is cheap to compute because, for example, $\sum_s \vW_t  E[\vz_s \vz_s^T] \vW_t^T = \vW_t  \left( \sum_s E[\vz_s \vz_s^T] \right) \vW_t^T$.  Calculating this for all $t$ while caching the value of $ \sum_s E[\vz_s \vz_s^T] $ costs $O(D^2 T)$.  The M step for $\vW$ costs $O(D^2 S T)$ or $O(D S T)$ depending on the approximation.    The dropped prefactors for all these are small -- 1 or 2, usually -- so taking everything together the costs of one full EM iteration is roughly $10 D^2 T S$ without the approximation and $10 D T S$ with it.  Assuming $T = 2 \times 10^5$, $D = 10$, $S = 500$ these come out to $10^{10}$ and $10^{11}$ flops, respectively.  On a single 1 GHz core ($10^9$ flops per second) this comes out to roughly 10 or 100 seconds.  Very conservatively, the EM algorithm will converge well within iterations, so the worst-case scenario is about two hours for the unapproximated version.  Given that a panel of normals only has to be learned once, no approximations will be needed in production.  However, we may find the approximate version very useful for development.
```
mbabadi: Nice discussion!

mbabadi:
Back to you @davidbenjamin.

docs/CNVs/CNV-methods.tex
davidbenjamin: Okay, fair enough.

docs/CNVs/CNV-methods.tex
davidbenjamin: I don't think we'd be spending a principal component.  Suppose the target-length dependence were exactly linear i.e. we had a perfect analytic model.  In that case it would renormalize the target means (before exponentiation) without affecting the principal components at all, since there are no fluctuations in this effect.  And maybe there is some latent feature that yields variability relating to target length, but we certainly don't know the structure of this relationship a priori.

To use the workhorse GC bias example, it's not actually GC bias per se that needs to be modeled explicitly.  If GC bias were significant but sample-independent it would get absorbed into the target means and a linear model would handle it exactly.  The reason GC needs to be modeled outside of PCA is because each sample has a different non-linear GC response function.  

Now this raises the question of whether maybe every sample has its own target-length bias curve just as it has its own GC bias curve.  If so, we could treat target length like GC.  Regardless, though, I don't think that modeling a sample-independent effect will spare any principal components or yield a more expressive model.

davidbenjamin:
@ mbabadi I made the change you asked for about the sample subscript.  I still disagree about the target length dependence (see above), at least insofar as the means are concerned, but I do see how one might consider treating it as an a priori known covariate for inter-sample variability just like GC.

mbabadi:
@davidbenjamin, @samuelklee: I did a prototype implementation in Python and I have bad news: it does not converge, and I think I know what the issue is.

In the M step, the equations for $m$, $W$, and $\Psi$ are coupled to each other and one can not simply solve for $m$ with the old $W$, proceed to find the new $W$, and finally the new $\Psi$; rather, these equations must be solved self-consistently. In Bishop §12.2.4, this problem is evaded by plugging the ML of x (here m) -- $\bar{x}$ -- from the outset in the M step equations. One is then left with just $\Psi$ and $W$. The equation for the MLE of $W$ turns out to be independent of $\Psi$, so it can be solved independently. Then one plugs the MLE of $W$ in the likelihood function and gets the MLE equation for $\Psi$, which gives (Eq. 12.70 -- note "W_new").

Using the approximation proposed by @davidbenjamin, i.e. replacing $\Psi_s$ with $\sum_s \Psi_s / S$, the MLE equations for $W$ and $\Psi$ will be decoupled and that's a progress in the right direction, however, we still have dependence on $m$. I think $m$ can be solved for exactly (from Eq. 47) as in Bishop before doing EM, so we don't need to estimate it. Intuitively, I expect $m_exact = \sum m_st / S$.

mbabadi:
@davidbenjamin, I think it is a good idea to dispense with all bold face symbols altogether and use explicit summation convention everywhere. Stuff like Eq. 55 look cryptic to me and take a while to "parse". At the moment, we have a mix of boldface matrices, boldface vectors, boldface matrices with either s or t specified, boldface vectors with specified indices (which should not be boldface), transpositions of these objects, etc.

davidbenjamin:
> I think it is a good idea to dispense with all bold face symbols altogether and use explicit summation
> convention everywhere. Stuff like Eq. 55 look cryptic to me and take a while to "parse".

@mbabadi Thanks for this comment.  I will change the notation.

davidbenjamin:
> I did a prototype implementation in Python and I have bad news.

@mbabadi Thank you for prototyping and finding this problem!  Catching mistakes is not bad news, just an invitation to think harder.

> I think $m$ can be solved for exactly in our case as well (by maximizing Eq. 47 before marginalizing $z$) as in Bishop, so we don't need to estimate it at every step. Intuitively, I expect $m_exact = \sum  m_st / S$.

I tried an approach analogous to Bishop's in which I marginalized $z$ (instead of averaging wrt its E step posterior) and then derived a max likelihood equation for $m$.  If we use the approximation $\Psi_s \rightarrow \sum_s \Psi_s / S$ it does indeed reduce to $m= \sum m_s / S$.  It's thornier and probably intractable without that approximation.  The intuitive reason that the simple result isn't exactly right is that low-coverage samples ought to be weighted less in the average.

However, this got me thinking and as of now I am more optimistic about the approximation than before.  So much so that I wouldn't worry about about using it, at least for $m$, to justify the use of the easy explicit formula.  Here's why: Basically, the approximation says that the sum of 1) Poisson variance and 2) variance due to noise we can't tangent-normalize away is comparable for all samples.  As we discussed earlier, this approximation is trivially good for the (probably common) case in which all PoN samples have comparable coverage.  But note also that it's good whenever contribution 2) dominates contribution 1), since 2) comes the the parameter $\Psi$ which is common to all samples.  If we can show that we only really care about situations where 2) is dominant we can use the approximation with confidence.

If 1) is small for all PoN samples (due to high coverage) but 2) is even smaller, the approximation isn't justified, but *who cares* -- if all the PoN samples have high coverage, we should be thrilled with our good fortune and not really worry about assigning more weight to a mega high coverage samples than to a merely super high coverage sample.  We should also be thrilled with our good fortune that latent space miraculously seems to produce no noise.  Basically, this is a scenario in which we have won the lottery and using the approximation won't change the fact that we won the lottery.

If 1) is large and 2) is small we have a case where noise due to low coverage -- which of course we have no hope of normalizing away -- dominates the inherent noise.  This is a case where the lab has invented a wonderful low-noise protocol but our terrible PoN has low depth.  In this case a good model isn't all that necessary, because the noise from latent space is small, and wouldn't help because the depth is simply too low.

In any case, I personally think the most likely scenario is neither of these but a straightforward case where 2) is bigger than 1) by virtue of the high depth of most exome sequencing.

With the permission of @mbabadi @samuelklee and @LeeTL1220 I would like to switch the proposal to use the simple non-iterative formula $m= \sum m_s / S$ for the mean.

mbabadi:
UpdatE: I got the code to work in the (semi-)approximate mode -- the persistent sign error in Eq. (53) was the culprit, please fix it @davidbenjamin. By semi-approximate, I mean using Eq. (56) rather than (55), but not doing $s$-averaging elsewhere. Even though we still get coupled equations for the M-step, we can still increase the log likelihood via sequential maximization along $m$, $W$, and $\Psi$, with replacement. This still guarantees the convergence of the EM algorithm.

We still have a remaining issue: if we use the exact Eq. (55), we often face the problem that it has no solutions (i.e. the optimum $\Psi$ lies in the boundary); but putting $\Psi$ on the boundary is dangerous, and I believe is a spurious solutions we'd like to avoid. I think we can fix this issue by regularizing $\Psi$ with a Gaussian prior.

mbabadi:
@davidbenjamin, @samuelklee -- what is your take on estimating $d_s$ -- the average sample-dependent raw coverage? I am not happy with using $d_s = (1/T) \sum_t n_st$; rather, I think we must treat it as a latent variable on par with $c_{st}$. Do you agree?

mbabadi:
**Update on ARD:** I implemented ARD and it works well -- if and only if the number of samples is ~ an order of magnitude larger than the dimension of the latent space, in which case, we indeed recover the expected number of principal components +- 1 (the ARD coefficients either diverge, or remain within the same order of magnitude of $T$).

If the number of samples is comparable to the dimension of the latent space, all ARD coefficients flow to $\infty$. This behavior is expected since the Laplace approximation grossly overestimates $\alpha$ if likelihood(W) is not sharply peaked arounds its MLE value.

In summary, for PoNs that have ~ 200 samples in them and the latent space dimensions is ~ 20, we are safe. For PoNs with ~ 30 samples, however, ARD does not work. We may still try to make it work e.g. by replacing the Laplace approximation with something that makes sense in the finite N limit. I bet there's already literature on this. One heuristic idea to try is to re-exponentiate the expectation of the log likelihood from the M step, plug it in Bishop 12.61, and treat it as an approximation of exact marginal likelihood. Then 12.61 will be a Gaussian integral which can be calculated analytically. Any thoughts @davidbenjamin, @samuelklee?

davidbenjamin:
@mbabadi Good to know that it works so well under decent conditions.  As for small PoNs, I also think re-exponentiating the M step likelihood is the way to go.  It doesn't strike me as all that heuristic, since it's the exact MLE for alpha if you split the variables into 1) z and 2) everything else and then do block variational Bayes.  I'll put this in the docs.

davidbenjamin:
@mbabadi I made the changes you requested (let me know about the readability of the new notation), put in explicit steps for exact and approximate modes, put in the new updates of the ARD coefficients, and added a few discussions, including some new reasons to believe in the approximation.

mbabadi:
Nice job @davidbenjamin , just a few minor issues:

- In Eq. (56), please change bold face $\Sigma_{st}$ to normal weight
- Minus sign error in Eq. (72). This error propagates through the rest of the text (in the text after Eq. 64, the mean is - (A+Q)^{-1} v_t, the 7th item of the summary of the exact and approximate mode)
- In the summary of the approximate mode, Q still needs to carry a target label. The multiplication complexity is still O(D^2 T S). Also, please put the t subscript on Q in item 7 and 8.

mbabadi:
If we apply the approximation $\Psi_{st} \rightarrow \Psi_t$ to Eq. (66), we find:
$$ m_t = \frac{1}{S} \sum_s (m_{st} - W_{t\mu} E[z_{s\mu}]) $$
For large number of samples $S$, the LLN implies that the second term converges to $0$ with statistical errors that decay like $1/\sqrt{S}$. In practice, is there a reason to retain these "statistical errors", or shall we just use Eq. (61)? for small sample sizes $S \sim 50$, these statistical fluctuations are very appreciable.

davidbenjamin:
> In Eq. (56), please change bold face $\Sigma_{st}$ to normal weight
Fixed.
>  Minus sign error in Eq. (72). This error propagates through the rest of the text. . .
Fixed by redefining v --> -v in Equations 70 and 71 and in the summary.
> In the summary of the approximate mode, Q still needs to carry a target label.
Fixed.
> The multiplication complexity is still O(D^2 T S).
I think it's O(D^2 (T + S)) if you distribute the sum over samples.  See revised.

davidbenjamin:
> In practice, is there a reason to retain these "statistical errors", or shall we just use Eq. (61)? for small 
> sample sizes $S \sim 50$, these statistical fluctuations are very appreciable.

The first term $\sum_s m_{st}$ also has 1/sqrt(S) statistical errors, so I don't think we will improve accuracy by modifying the MLE result and discarding the second term.  I do feel though that this is another argument in favor of always using Eq. (61) since intuitively any error or uncertainty due to the E step will propagate into the M step and amplify the 1/sqrt(S) error.  

I think it's worth it to test both ways on real data before committing to use Equation 61 everywhere, but I suspect we will end up adopting it.

davidbenjamin:
@mbabadi back to you

mbabadi:
Thanks for your comments @davidbenjamin. Why don't we keep the notes as it is, and revise it accordingly once we lay hands on real data? :)

davidbenjamin:
@mbabadi sounds good to me.

Issue #415, opened 2016-03-24
SparkSingularValueDecomposerUnitTest.testBasicFunction failing in master


samuelklee:
Gradle suite > Gradle test > org.broadinstitute.hellbender.utils.svd.SparkSingularValueDecomposerUnitTest.testBasicFunction FAILED
    java.lang.OutOfMemoryError: Java heap space
        at org.apache.commons.math3.linear.BlockRealMatrix.createBlocksLayout(BlockRealMatrix.java:270)
        at org.apache.commons.math3.linear.BlockRealMatrix.<init>(BlockRealMatrix.java:106)
        at org.apache.commons.math3.linear.MatrixUtils.createRealMatrix(MatrixUtils.java:81)
        at org.apache.commons.math3.linear.MatrixUtils.createRealIdentityMatrix(MatrixUtils.java:176)
        at org.apache.commons.math3.linear.QRDecomposition$Solver.getInverse(QRDecomposition.java:473)
        at org.broadinstitute.hellbender.utils.svd.SVDTestUtils.assertUnitaryMatrix(SVDTestUtils.java:23)
        at org.broadinstitute.hellbender.utils.svd.SVDTestUtils.assertSVD(SVDTestUtils.java:72)
        at org.broadinstitute.hellbender.utils.svd.SparkSingularValueDecomposerUnitTest.testBasicFunction(SparkSingularValueDecomposerUnitTest.java:31)

vruano:
I guess the solution may consist in reduce the size of the problem to avoid the OoM error given that travis  VM resources are restricted.

samuelklee:
@lbergelson can you give us a hand with some travis issues?

if increasing resources is not an option then we'll go with @vruano's suggestion.  @LeeTL1220

lbergelson:
The test jvm size is set very low right now.  It can likely be safely increased.  Do you know how much heap space you need for this test?

Currently the following extremely conservative settings are set.  These were configured when you were running in a vm with 3gb of memory total and never relaxed when we moved to a large vm.  The current vm has 7.5 gb available. 
```
    // set heap size for the test JVM(s)
    minHeapSize = "1G"
    maxHeapSize = "1.5G"
    jvmArgs = ['-Xmx2G']
```
You could probably just remove the heap size arguments and set `jvmArgs = ['Xmx6G']`.  If that fixes the problem you're good.  If it doesn't you'll have to reduce the size of the test input.

PR #419, opened 2016-03-27
Adds a tool to transform Genome-Strip vcf calls into segments.

vruano:


vruano:
test are failing only due to a separate issue in master described in #415. 

@kareembroad please review.

vruano:
The second commit is just to fix another standing issue in master (testing), not really part of this pull-request but spotted and fixed while doing work in it.

vruano:
Example syntax:

java -jar hellbender-protected.jar ConvertGSVariantsToSegments -V gs-calls.vcf.gz -O my-output.tab -targets my-targets.tab

vruano:
Just rebased it, to see whether the testing problems go away.

vruano:
@kareembroad did you have a chance to review the PR?

vruano:
@kareembroad, beside the travis failure do you have any comments on the code; is there anything that you rather change.

vruano:
Ok I seem to been able to fix the Travis problem. The coverage is worse mainly because I put some of the test classes under the main src directory, once I move them to test, that should be fixed as well. @kareembroad, any comments on the code?

vruano:
Done some commit squashing and moved the integration test to the test source folder.

vruano:
No more changes unless there some review comments to address.
Is there any @kareembroad?

vruano:
@kareembroad. Can you take a look at the pull request today?

vruano:
@LeeTL1220 can you take a look into this pull-request? Second commit is a piggybacking fix for a existing problem in master test.

scripts/install_R_packages.R
```diff
@@ -13,4 +13,9 @@ dependencies = c("naturalsort")
 if (!all(dependencies %in% rownames(installed.packages()))) {
   install.packages(setdiff(dependencies, rownames(installed.packages())), repos="http://cran.cnr.Berkeley.edu")
 }
+
+hmmUrl = "http://cran.r-project.org/src/contrib/HMM_1.0.tar.gz"
```
LeeTL1220: This is for testing only, right?  We are not going to be dependent on another R package, right?

src/main/java/org/broadinstitute/hellbender/tools/exome/eval/ConvertGSVariantsToSegments.java
```diff
@@ -0,0 +1,203 @@
...
+import java.util.List;
+import java.util.stream.Stream;
+
+/**
+ * Tool to convert Genotype Strip Variant call file into {@link CopyNumberTriStateSegmentRecord} table file.
+ *
+ * @author Valentin Ruano-Rubio &lt;valentin@broadinstitute.org&gt;
+ */
+@CommandLineProgramProperties(
+        summary = "Convert GS VCF file contents into a CopyNumberTriState segment file",
```
LeeTL1220: Can you spell out Genome STRiP  and include a URL?
http://www.broadinstitute.org/software/genomestrip/ should work

src/main/java/org/broadinstitute/hellbender/tools/exome/eval/ConvertGSVariantsToSegments.java
```diff
@@ -0,0 +1,203 @@
...
+        oneLineSummary = "Convert GS VCF file into a segment file",
```
LeeTL1220: Only spell out genome strip for the one line summary.  No need for the URL

LeeTL1220:
@vruano Why is the R library needed?  Apologies if I missed something obvious.

scripts/install_R_packages.R
vruano: Yup, just for testing.

vruano:
To verify that our HMM engine is doing the right comparing results with another implementation, in this case one available in R.

vruano:
I have moved the additional R package installation to a separate script.

vruano:
Back to @LeeTL1220

.travis.yml
```diff
@@ -48,6 +48,7 @@ before_install:
 - sudo apt-get install -y --force-yes r-base-dev=3.1.3-1trusty
 - sudo apt-get install -y --force-yes r-base-core=3.1.3-1trusty
 - sudo Rscript scripts/install_R_packages.R
+- sudo Rscript scripts/install_R_packages_for_testing.R
```
LeeTL1220: There is documentation (on the forum and in README) that will need to be updated to match this step. 

Also, you should update the Dockerfile.

vruano:
I merged those scripts back again as discussed.

vruano:
Back to @LeeTL1220

PR #424, opened 2016-03-30
Plotting R scripts check for missing input and output

davidbenjamin:
and quit with exit code recognizable by GATK.


davidbenjamin:
@samuelklee I think it's good to go.

samuelklee:
looks good, i think!

PR #425, opened 2016-03-30
Changed ACNV to clean up sample names derived from file paths.

samuelklee:


samuelklee:
i think this should do the trick?

davidbenjamin:
@samuelklee @LeeTL1220 Ready to merge.  I will probably be on my commute when Travis passes -- anyone who wants it sooner has my permission to go ahead.

PR #428, opened 2016-03-31
Exposing ability to change minimum read count in het pulldown and set…

LeeTL1220:
…ting default to 15 reads, instead of ten.


src/main/java/org/broadinstitute/hellbender/tools/exome/GetHetCoverage.java
```diff
@@ -108,6 +111,14 @@
             common=true)
     protected ValidationStringency VALIDATION_STRINGENCY = ReadConstants.DEFAULT_READ_VALIDATION_STRINGENCY;
 
+    @Argument(
+            doc = "Minimum raw number of reads that must be present to even be considered as a het.",
```
samuelklee: maybe "Minimum raw number of reads that must be present at a site for it to be considered for het compatibility." instead?

samuelklee:
just a minor suggestion, back to you!

PR #437, opened 2016-04-05
several gradle changes

lbergelson:
Three simple commits:

1.  added a settings.gradle file
2.  updated gradle to 2.12 to match public, updated readme to recommend `./gradlew` instead of `gradle`
3.  removed some duplicated code and moved some into the shared shadowJar settings, combined two configurations blocks


build.gradle
```diff
@@ -2,26 +2,17 @@ buildscript {
...
-        classpath 'com.github.jengelman.gradle.plugins:shadow:1.2.2'
-    }
 }
 
 plugins {
-    id "de.undercouch.download" version "1.2"
+    id "java"
+    id "application"
+    id "jacoco"
+    id "com.github.johnrengelman.shadow" version "1.2.3"
```
LeeTL1220: Can you add some quick comments about what some of these are?

build.gradle
```diff
@@ -295,46 +295,19 @@ shadowJar {
...
-        exclude group: 'org.slf4j'
-        exclude module: 'jul-to-slf4j'
-        exclude module: 'javax.servlet'
-        exclude module: 'servlet-api'
-        exclude group: 'com.esotericsoftware.kryo'
-        exclude module: 'spark-mllib_2.10'
-        exclude group: 'org.scala-lang'
-        exclude module: 'kryo'
+shadowJar {
+    classifier = 'spark_standalone'
```
LeeTL1220: Does classifier ever change?

lbergelson: shadowJar and sparkJar have different classifiers, as far as I know they're totally arbitrary and used to make it more human readable output

LeeTL1220:
@lbergelson One minor comment and then feel free to merge

lbergelson:
@LeeTL1220 I've made an additional change since your review.  I had to update the .travis.yml as well.  It was using a different version of gradle for some commands which was causing tests to be run twice.  This should correct it.

PR #442, opened 2016-04-08
fix the incomplete manifest in sparkJar and shadowJar

lbergelson:
I messed up the shadow and spark jars with my previous commit.  I thought the settings on the jar task would propogate to the shadow tasks, but it turns out they didn't.  I'd checked to make sure that they still had a manifest, but it turned out they had an empty manifest.

This fixes it.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -221,7 +221,7 @@ logger.info("build for version:" + version);
 group = 'org.broadinstitute'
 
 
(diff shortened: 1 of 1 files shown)


LeeTL1220:
I'll take your word for it.

LeeTL1220:
Feel free to merge.

PR #447, opened 2016-04-15
Eliminated TargetCoverage in favor of ReadCountCollection.  Closes #397.

### LINKED ISSUE #397: Replace TargetCoverage with ReadCountCollection
Note that this should not change any output formats, which is a separate issue #201.

davidbenjamin:


davidbenjamin:
@LeeTL1220 As written this replaces the "name contig start stop" format with "NAME CONTIG START END" internally and in all output.  I forgot what the spec for this issue was and can easily change this.  Is it supposed to:
* do this
* switch to the lowercase CGA format everywhere
* do this but have some command line tools for format conversion
* none of the above
?

davidbenjamin:
@LeeTL1220 All output files are now in the lowercase header format (i.e. unchanged).  No changes to pipeline scripts will be necessary.  Can I assign you to review?

src/main/java/org/broadinstitute/hellbender/tools/exome/ACNVCopyRatioModeller.java
```diff
@@ -362,13 +362,13 @@ private CopyRatioDataCollection(final SegmentedModel segmentedModel) {
                 throw new IllegalArgumentException("Cannot construct CopyRatioDataCollection with no target-coverage data.");
             }
             //construct coverages and number of targets per segment from the segments and Genome in SegmentedModel
-            final TargetCollection<TargetCoverage> targetCoverages = segmentedModel.getGenome().getTargets();
+            final TargetCollection<ReadCountRecord.SingletonRecord> targetCoverages = segmentedModel.getGenome().getTargets();
```
LeeTL1220: Why is this not a ReadCountCollection?  Why use a ReadCountRecord.SingletonRecord?

src/main/java/org/broadinstitute/hellbender/tools/exome/Case2PoNTargetMapper.java
```diff
@@ -96,7 +96,6 @@ public RealMatrix fromPoNToCaseCounts(final RealMatrix ponCounts) {
      * @return never {@code null} a read-count collection with the row order according to the case read count target order.
      */
     public ReadCountCollection fromPoNtoCaseCountCollection(final RealMatrix counts, final List<String> countColumnNames) {
-        return new ReadCountCollection(SetUniqueList.setUniqueList(new ArrayList<>(outputTargets)),SetUniqueList.setUniqueList(new ArrayList<>(countColumnNames)),
-                fromPoNToCaseCounts(counts));
+        return new ReadCountCollection(outputTargets, countColumnNames, fromPoNToCaseCounts(counts));
```
LeeTL1220: The previous version was creating a copy of the lists.  Are you sure that is okay to move to reference?

src/main/java/org/broadinstitute/hellbender/tools/exome/CnvCaller.java
```diff
@@ -52,39 +55,39 @@
...
+     * Make calls for a list of segments based on the coverage data in a set of targets and their tangent-normalized coverage.
+     *  @param normalizedCoverage the collection representing targets and their tangent-normalized coverage
+     * @param segments segments, each of which holds a reference to these same normalizedCoverage
      * @param zThreshold number of standard deviations from mean required to call an amplification or deletion
      */
-    public static List<ModeledSegment> makeCalls(final TargetCollection<TargetCoverage> targets, final List<ModeledSegment> segments, final double zThreshold) {
+    public static List<ModeledSegment> makeCalls(final ReadCountCollection normalizedCoverage, final List<ModeledSegment> segments, final double zThreshold) {
         Utils.nonNull(segments, "Can't make calls on a null list of segments.");
-        Utils.nonNull(targets, "Can't make calls on a null list of targets.");
+        Utils.nonNull(normalizedCoverage, "Can't make calls on a null list of normalizedCoverage.");
```
LeeTL1220: Technically, normalizedCoverage is a tangentNormalizedCoverageProfile....

src/main/java/org/broadinstitute/hellbender/tools/exome/CnvCaller.java
```diff
@@ -52,39 +55,39 @@
...
+        final double []  nonExtremeCoverage = Arrays.stream(normalizedCoverage.counts().getColumn(0))
+                .filter(c -> Math.abs(c) < NON_NEUTRAL_THRESHOLD).toArray();
         final double neutralCoverage = new Median().evaluate(nonExtremeCoverage);
 
         logger.info(String.format("True copy neutral estimate (median): %.3f", neutralCoverage));
         logger.info(String.format("True copy neutral estimate (median) in CR: %.3f", Math.pow(2, neutralCoverage)));
 
-        // Get the standard deviation of targets belonging to high-confidence neutral segments
+        // Get the standard deviation of normalizedCoverage belonging to high-confidence neutral segments
+        final TargetCollection<ReadCountRecord.SingletonRecord> targetsWithCoverage =
```
LeeTL1220: Do we still need TargetCollection<...>?  Can we just use ReadCountCollection?

src/main/java/org/broadinstitute/hellbender/tools/exome/NormalizeSomaticReadCounts.java
```diff
@@ -242,7 +242,7 @@ private void writeTargetFactorNormalizedOutput(final ReadCountCollection targetF
      */
     private void writeTangentNormalizedOutput(final ReadCountCollection tangentNormalized) {
         try {
-            ReadCountCollectionUtils.writeAsTargetCoverage(outFile, tangentNormalized, "fileFormat = tsv",
+            ReadCountCollectionUtils.write(outFile, tangentNormalized, "fileFormat = tsv",
```
LeeTL1220: Are you *sure* this does not break something downstream?  Did you run the evaluation workflow?  Or make sure that the output is exactly the same as it was before?

src/main/java/org/broadinstitute/hellbender/tools/exome/ReadCountCollection.java
```diff
@@ -35,26 +34,12 @@
...
-     * </p>
+     * Read counts matrix with one row per target in {@link #targets} and columns corresponding to {@link #columnNames}.
      */
     private final RealMatrix counts;
 
     /**
      * Creates a new read-counts collection.
-     *
-     * <p>
-     *     All values in the input are copied, so these can be changed after invoking the constructor without
```
LeeTL1220: Why was this comment deleted?  Are you sure it is okay to eliminate the copy?

src/main/java/org/broadinstitute/hellbender/tools/exome/ReadCountCollection.java
```diff
@@ -72,30 +57,18 @@
      *     <li>{@code targets} contains any {@code null}},</li>
      *     <li>{@code columnNames} contains any {@code null}},</li>
      *     <li>{@code counts} contains any {@code null}},</li>
+     *     <li>{@code targets} contains any duplicates,</li>
+     *     <li>{@code columnNames} contains any duplicates,</li>
      *     <li>{@code targets} length does not match {@code counts} length or</li>
      *     <li>{@code counts} elements length does not match {@code columnNames} length</li>
      * </ul>
      */
-    public ReadCountCollection(final SetUniqueList<Target> targets, final SetUniqueList<String> columnNames, final RealMatrix counts) {
```
LeeTL1220: Why can you eliminate all of these checks?

src/main/java/org/broadinstitute/hellbender/tools/exome/ReadCountCollection.java
```diff
@@ -72,30 +57,18 @@
...
-        this.targets = Collections.unmodifiableList(new ArrayList<>(targets));
-        this.columnNames = Collections.unmodifiableList(new ArrayList<>(columnNames));
-        this.counts = counts.copy();
+    public ReadCountCollection(final List<Target> targets, final List<String> columnNames, final RealMatrix counts) {
+        this(targets, columnNames, counts, true);
     }
 
     /**
-     * Creates a new collection without verifying field values and without copying inputs.
+     * Creates a new collection or with or without verifying field values and without copying inputs.
```
LeeTL1220: I think it is copying the inputs, right?

src/main/java/org/broadinstitute/hellbender/tools/exome/ReadCountCollection.java
```diff
@@ -104,10 +77,32 @@ public ReadCountCollection(final SetUniqueList<Target> targets, final SetUniqueL
      * @param columnNames column name list, not a {@code null}, does not contain any {@code null}, does not contain repeats.
      * @param counts count matrix, not a {@code null}, has as many rows as {@code targets} elements and as many columns as {@code columnNames} elements.
      */
-    private ReadCountCollection(final List<Target> targets, final List<String> columnNames, final RealMatrix counts) {
-        this.targets = targets;
-        this.columnNames = columnNames;
-        this.counts = counts;
+    private ReadCountCollection(final List<Target> targets, final List<String> columnNames, final RealMatrix counts, final boolean verifyInput) {
```
LeeTL1220: Why would we ever not want to verify the input?

src/main/java/org/broadinstitute/hellbender/tools/exome/ReadCountRecord.java
```diff
@@ -121,6 +121,10 @@ public double getDouble(final int index) {
         return counts[index];
     }
 
+    public SingletonRecord asSingleton() {
+        return new SingletonRecord(this);
+    }
+
     @Override
     @Deprecated
     public String getChr() {
```
LeeTL1220: Can we delete this method?

src/main/java/org/broadinstitute/hellbender/tools/exome/TangentNormalizer.java
```diff
@@ -200,7 +200,7 @@ public static TangentNormalizationResult tangentNormalizeNormalsInPoN(final PoN
         final List<String> sampleNamesCopy = new ArrayList<>(pon.getSampleNames());
 
         logger.info("Loading normalized counts...");
-        final ReadCountCollection coverageProfile = new ReadCountCollection(SetUniqueList.setUniqueList(pon.getTargets()), SetUniqueList.setUniqueList(sampleNamesCopy), pon.getNormalizedCounts());
+        final ReadCountCollection coverageProfile = new ReadCountCollection(pon.getTargets(), sampleNamesCopy, pon.getNormalizedCounts());
```
LeeTL1220: Why are you no longer enforcing the SetUniqueList?

src/main/java/org/broadinstitute/hellbender/tools/exome/TargetCoverageStatsReader.java
```diff
@@ -61,7 +61,7 @@ public TargetCoverageStatsReader(final File outputFile) throws IOException {
                             dataLine.getInt(endColumnIndex)));
 
         } else {
-            final int nameColumnIndex = columns.indexOf(TargetTableColumns.NAME.name());
```
LeeTL1220: Do we need this class at all?

src/main/java/org/broadinstitute/hellbender/tools/exome/TargetTableColumns.java
```diff
@@ -13,10 +13,10 @@
  * @author Valentin Ruano-Rubio &lt;valentin@broadinstitute.org&gt;
  */
 public enum TargetTableColumns {
-    NAME("NAME", true),
-    CONTIG("CONTIG", true),
-    START("START", true),
-    END("END", true),
+    NAME("name", true),
```
LeeTL1220: Are you sure this is okay?  I tried this before and a lot of things broke...

src/test/java/org/broadinstitute/hellbender/tools/exome/ACNVCopyRatioModellerUnitTest.java
```diff
@@ -101,13 +105,13 @@ private static double relativeError(final double x, final double xTrue) {
      * </p>
      */
     @Test
-    public void testRunMCMCOnCopyRatioSegmentedModel() {
+    public void testRunMCMCOnCopyRatioSegmentedModel() throws IOException {
```
LeeTL1220: A test that throws an IOException?  That seems a bit abnormal...

src/test/java/org/broadinstitute/hellbender/tools/exome/ACNVModellerUnitTest.java
```diff
@@ -44,18 +45,18 @@
      * remerge segments broken by these breakpoints and reproduce the original list of true segments.
      */
     @Test
-    public void testMergeSimilarSegmentsCopyRatio() {
+    public void testMergeSimilarSegmentsCopyRatio() throws IOException {
```
LeeTL1220: Why are you throwing IOException from an automated test?

src/test/java/org/broadinstitute/hellbender/tools/exome/AllelicCountUnitTest.java
```diff
@@ -48,16 +48,4 @@ public void testEstimateAltAlleleFraction() {
         Assert.assertEquals(new AllelicCount(interval, 10, 20).estimateAltAlleleFraction(), 0.6666666, tolerance);
         Assert.assertEquals(new AllelicCount(interval, 10, 90).estimateAltAlleleFraction(), 0.9, tolerance);
     }
-
-    @Test
```
LeeTL1220: Does a test need to be done to replace this one or is truly extraneous?

src/test/java/org/broadinstitute/hellbender/tools/exome/NormalizeSomaticReadCountsIntegrationTest.java
```diff
@@ -98,7 +96,7 @@ public void testBadTargetFile() throws IOException {
         runCommandLine(arguments);
     }
 
-    @Test(expectedExceptions = UserException.BadInput.class)
+    @Test
```
LeeTL1220: Why does this no longer throw an BadInput exception?

src/test/java/org/broadinstitute/hellbender/tools/exome/NormalizeSomaticReadCountsIntegrationTest.java
```diff
@@ -207,7 +205,7 @@ public void testFullInputRunWithTargetFileWithBadNameOneSample() throws IOExcept
...
```
LeeTL1220: How can this be correct to eliminate the expected exception?

src/test/java/org/broadinstitute/hellbender/tools/exome/NormalizeSomaticReadCountsIntegrationTest.java
```diff
@@ -262,46 +260,24 @@ public void testFullReadCountsInputRunOneSample() throws IOException {
...
 
         // Make sure that we can read in the tangent normalized targets as a collection of TargetCoverage
-        final TargetCollection<TargetCoverage> targets = TargetCoverageUtils.readModeledTargetFileIntoTargetCollection(new File(tangentNormalizationOutput.getAbsolutePath()));
-        Assert.assertEquals(targets.targetCount(), 657);
-        Assert.assertEquals(targets.targets().get(2).getName(), "target_179700_CRYBB1");
+        Assert.assertEquals(tangentNormalized.targets().size(), 657);
+        Assert.assertEquals(tangentNormalized.targets().get(2).getName(), "target_179700_CRYBB1");
     }
 
-    @Test(expectedExceptions = UserException.BadInput.class)
```
LeeTL1220: Why can we get rid of this unit test?

LeeTL1220: I think that downstream tools will not support more than one sample in the file.

src/test/java/org/broadinstitute/hellbender/tools/exome/NormalizeSomaticReadCountsIntegrationTest.java
```diff
@@ -374,8 +348,8 @@ private RealMatrix readBetaHats(File betaHatsOutput, ReadCountCollection input)
         return new Array2DRowRealMatrix(betaHats,false);
     }
 
-    @Test(expectedExceptions = UserException.BadInput.class)
```
LeeTL1220: Are you sure we can get rid of this expected exception?

src/test/java/org/broadinstitute/hellbender/tools/exome/ReadCountCollectionUtilsUnitTest.java
```diff
@@ -193,7 +193,7 @@ public void testReadIntervalsOnlyFile() throws IOException {
         targetWriter.close();
         final File testFile = createTempFile();
         final PrintWriter writer = new PrintWriter(testFile);
-        writer.println("CONTIG\tSAMPLE3\tEND\tSAMPLE2\tSTART");
+        writer.println("contig\tSAMPLE3\tstop\tSAMPLE2\tstart");
```
LeeTL1220: This owuld imply that the workflow will fail.

src/test/java/org/broadinstitute/hellbender/tools/exome/SegmentUtilsUnitTest.java
```diff
@@ -39,28 +43,16 @@ public void testReadAndWriteCalledIntervals() {
     }
 
     @Test
-    public void testMeanTargetCoverage() {
```
LeeTL1220: Why can we delete this test?

src/test/java/org/broadinstitute/hellbender/tools/genome/SparkGenomeReadCountsIntegrationTest.java
```diff
@@ -26,7 +23,7 @@
     private static final File REFERENCE_FILE = new File("src/test/resources/hg19mini.fasta");
 
     @Test
-    public void testSparkGenomeReadCounts() {
+    public void testSparkGenomeReadCounts() throws IOException {
```
LeeTL1220: Why are we adding so many ``throws IOException``?

LeeTL1220:
I have lots of comments.  Some things I did not understand, so I may just need an explanation.  

I really do not think that you can mess with the column headers (even to change the lower case upper case) and still have the tools work in the workflow.  That may have changed at some point, but the real test is whether we can run a workflow.  If you already ran the workflow and everything was fine, then you can disregard those comments.

src/main/java/org/broadinstitute/hellbender/tools/exome/ACNVCopyRatioModeller.java
davidbenjamin: `Genome` stores its targets in a`TargetCollection`, so this is what you get when query the `SegmentedModel` for its targets.  Although this particular code would probably be fine if `Genome` stored `Targets` in a `ReadCountCollection`, there are other places where the ability of a `TargetCollection` to give all targets in a given interval is needed.

It might make sense to add this capability to `ReadCountCollection`, but I think that's beyond the scope of this ticket.

src/main/java/org/broadinstitute/hellbender/tools/exome/Case2PoNTargetMapper.java
davidbenjamin: The `ReadCountCollection` constructor now makes a defensive copy.  Previously the constructor took a `SetUniqueList` and literally every time it was invoked the calling code had to construct a `SetUniqueList`.  Moving the copying into the constructor eliminated all of this boilerplate.

src/main/java/org/broadinstitute/hellbender/tools/exome/CnvCaller.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/exome/CnvCaller.java
davidbenjamin: See above -- basically, `TargetCollection` has a method to get all targets overlapping a given `Locatable`, which `ReadCountCollection` lacks.  It probably makes sense to refactor `ReadCountCollection` in this way, but I would prefer to handle that separately.  In fact, I would enjoy doing this, but there's something to be said for smaller PRs.

src/main/java/org/broadinstitute/hellbender/tools/exome/NormalizeSomaticReadCounts.java
davidbenjamin: I shall do my due diligence and report back. . .

src/main/java/org/broadinstitute/hellbender/tools/exome/ReadCountCollection.java
davidbenjamin: The public constructor always makes copies.  I have fixed the javadoc.

src/main/java/org/broadinstitute/hellbender/tools/exome/ReadCountCollection.java
davidbenjamin: It copies if `verifyInput == true`, which is always the case when the entry point is the public constructor.

src/main/java/org/broadinstitute/hellbender/tools/exome/ReadCountCollection.java
davidbenjamin: This is the same behavior as before: the private non-verifying constructor is useful when invoked by `subsetTargets`, `subsetColumns`, and `arrangeTargets`.

src/main/java/org/broadinstitute/hellbender/tools/exome/ReadCountRecord.java
davidbenjamin: It's necessary to implement `Feature`.

src/main/java/org/broadinstitute/hellbender/tools/exome/TangentNormalizer.java
davidbenjamin: The previous code was kind of clumsy in that there was no inherent reason for `SetUniqueList`.  It was *only* used as an indirect way to check for duplicates.  I have replaced this behavior with a more natural and easier-to-debug approach of throwing an error in the constructor if there are duplicates.

src/main/java/org/broadinstitute/hellbender/tools/exome/TargetCoverageStatsReader.java
davidbenjamin: This class is not related to the deceased `TargetCoverage`.  It's part of the germline pipeline for filtering targets, which itself will probably be on the chopping block soon, but not yet.

src/main/java/org/broadinstitute/hellbender/tools/exome/TargetTableColumns.java
davidbenjamin: I could tell you why this will work, or I could run the darn workflow and tell you what happens.  Hmmm, I wonder what the responsible thing to do would be?

src/test/java/org/broadinstitute/hellbender/tools/exome/ACNVCopyRatioModellerUnitTest.java
davidbenjamin: If the test doesn't `throw` you have to wrap `ReadCountCollectionUtils.parse` in a `try. . . catch`, which strikes me as an annoying thing to do in testing code.  If you have a different opinion on the lesser of these two evils I will adopt it.

src/test/java/org/broadinstitute/hellbender/tools/exome/ACNVModellerUnitTest.java
davidbenjamin: Same as above -- eliminates the need for a `try. . . catch`.

src/test/java/org/broadinstitute/hellbender/tools/exome/AllelicCountUnitTest.java
davidbenjamin: It's extraneous.  The "backend" `estimateMinorAlleleFraction` is already tested, and the need to convert to an intermediate `TargetCoverage` no longer exists.

src/test/java/org/broadinstitute/hellbender/tools/exome/NormalizeSomaticReadCountsIntegrationTest.java
davidbenjamin: This one took me a while to figure out: the name of the test is misleading.  It wasn't supposed to fail due to the extra target (targets not in the PoN are ignored.  Note that the next test, `testFullInputWithExtraTargetOneSample` does not have an expected exception.  It failed because it had more than one sample, because converting to `TargetCoverage` requires a single sample.  But now that output remains a `ReadCountCollection` there is no problem with multiple samples.

src/test/java/org/broadinstitute/hellbender/tools/exome/NormalizeSomaticReadCountsIntegrationTest.java
davidbenjamin: I replaced this test with one that is supposed to ruin without an exception since now there's no reason to forbid multiple samples.

davidbenjamin: I will check downstream CLIs.

src/test/java/org/broadinstitute/hellbender/tools/exome/NormalizeSomaticReadCountsIntegrationTest.java
davidbenjamin: Same thing: despite the name, it was only meant to fail due to the multiple samples, not the `CoordOnly`-ness.

src/test/java/org/broadinstitute/hellbender/tools/exome/SegmentUtilsUnitTest.java
davidbenjamin: Because the only uses of `meanTargetCoverage()` were in `testMeanTargetCoverage()`.

src/test/java/org/broadinstitute/hellbender/tools/genome/SparkGenomeReadCountsIntegrationTest.java
davidbenjamin: See above -- I don't like putting `try . . . catch` blocks in test code.

davidbenjamin:
@LeeTL1220 So the main messages I get are 1) run the workflow!!!! and 2) try to give `ReadCountCollection` some of the capabilities of `TargetCollection`, and 3) check that downstream CLIs properly check for multiple samples.  I think 2) is worthwhile but not in this PR.  The rest I either changed or explained.  If you don't buy any explanation I will revisit.  Now I must run the workflow and hope nothing blows up.

LeeTL1220:
@davidbenjamin It is going to blow up.

davidbenjamin:
@LeeTL1220 You were right, of course.  It blew up, but it seems like into not so many pieces.  What I overlooked was that `CalculateTargetCoverage` in gatk-public uses uppercase column headers.  I am testing the following pragmatic solution: copy it into gatk-protected under the name `CalculateTargetCoverageCopy` and change to lowercase.  I have modified the pipeline to invoke the copy but otherwise left it unchanged.  If this works I would then submit a PR to gatk-public to change `CalcualteTargetCoverage` and then delete the copy in gatk-protected.

I will know by the morning if this works.

davidbenjamin:
@LeeTL1220 It works!  Everything is in /dsde/working/davidben/formats.  Every output file is as it should be.  The orientation is as follows:

* Full run of both PoN creation and ACNV.
* 10 LUAD tumor bams w/ 10 paired normals -- LUAD_tumors.txt and LUAD_normals.txt, respectively.
* 49 agilent bams -- agilent_pon_bams.list -- for the PoN.
* Create PoN pipeline results in pon/ directory.
* Case sample pipeline results in cases/ directory.
* running db_format branch of gatk-protected.
* running db_format branch of recapseg-hb-eval, which is identical to master but using the copy of `CalculateTargetCoverage`.
* invoking `./test.sh` produces all output.

I believe it's back to you now.  Let me know if any of my answers to your comments were in fact nothing more than evasions.

LeeTL1220:
@davidbenjamin You have failing unit tests and coverage decreased by a lot....  please address

davidbenjamin:
@LeeTL1220 Okay, for some reason I don't undersand commenting out the override annotation for `onTraversalEnd()` in the copied class fixed the problem.  BTW, it compiled fine without this hack locally and on gsa5.  As for the coverage, as discussed offline I believe the drop in coverage is solely due to refactoring some highly-covered classes out of existence.  This PR does not add new functionality and I can't think of anything that is no longer covered by tests.  Therefore, I think it's back to you for real.

src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetBaseCallCoverage.java
```diff
@@ -44,7 +44,7 @@
  * </h4>
  * <h4>Input and Output</h4>
  * <p>
- * The user must indicate the target table file using the argument {@link TargetArgumentCollection#targetsFile targetsFile}.
+ * The user must indicate the target table file using the argument {@link TargetArgumentCollection#targetsFile normalizedCoveragedFile}.
```
samuelklee: any chance we could be consistent with use of normalizedCoverage vs. tangentNormalizedCoverage throughout?

src/main/java/org/broadinstitute/hellbender/tools/exome/Genome.java
```diff
@@ -12,37 +15,46 @@
...
+     * @param normalizedCoverageFile    linear target-coverage file
      * @param snpFile       SNP-allele-count file
      * @param sampleName    name of the sample
      */
-    public Genome(final File targetFile, final File snpFile, final String sampleName) {
-        this(TargetCoverageUtils.readTargetsWithCoverage(targetFile), new AllelicCountCollection(snpFile).getCounts(),
-                sampleName);
+    public Genome(final File normalizedCoverageFile, final File snpFile, final String sampleName) {
+        try {
+            final ReadCountCollection poo = ReadCountCollectionUtils.parse(normalizedCoverageFile);
```
samuelklee: why the dummy variable?

src/main/java/org/broadinstitute/hellbender/tools/exome/Genome.java
```diff
@@ -12,37 +15,46 @@
...
+
+    //TODO: sampleName can be inferred from the ReadCounts
     /**
      * Constructs a genome from lists containing linear target-coverage and SNP-allele-count data.
      * @param targets       list of linear target coverages, cannot be {@code null}
      * @param snps          list of SNP allele counts, cannot be {@code null}
      * @param sampleName    name of the sample, cannot be {@code null}
      */
-    public Genome(final List<TargetCoverage> targets, final List<AllelicCount> snps, final String sampleName) {
+    public Genome(ReadCountCollection targets, final List<AllelicCount> snps, final String sampleName) {
```
samuelklee: final ReadCountCollection targets

src/test/java/org/broadinstitute/hellbender/tools/exome/allelefraction/AlleleFractionSimulatedData.java
```diff
@@ -22,6 +23,11 @@
  * @author David Benjamin
  */
 public final class AlleleFractionSimulatedData {
+    public static final ReadCountCollection TRIVIAL_TARGETS = new ReadCountCollection(
```
samuelklee: this seems a bit unsatisfying to me...is there no way to have an empty ReadCountCollection?

src/main/java/org/broadinstitute/hellbender/tools/exome/ACNVCopyRatioModeller.java
samuelklee: ReadCountRecord.SingletonRecord is verbose and yet is not very descriptive/evocative...could this be renamed?

src/main/java/org/broadinstitute/hellbender/tools/exome/CnvCaller.java
```diff
@@ -27,6 +28,8 @@
  *
  * Note:  Assumes that the incoming target file is in log_2(CR)
  * Note:  Assumes that the incoming segment means are CR.
+ * Note: Assumes that the input @{link ReadCountCollection} contains only one sample -- calls will be made for only one
```
samuelklee: using one column of a matrix to represent a real vector quantity seems like a code smell to me.  the fact that "count" is in the name of the class (implying integer quantities) is also telling.  i think that having the appropriate abstractions is worth an extra class or two.

src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetBaseCallCoverage.java
davidbenjamin: Thanks for joining the conversation.  That sounds worthwhile to me.

src/main/java/org/broadinstitute/hellbender/tools/exome/Genome.java
davidbenjamin: Good catch.  And now you know what I name my variables when I'm debugging.

src/test/java/org/broadinstitute/hellbender/tools/exome/allelefraction/AlleleFractionSimulatedData.java
davidbenjamin: I agree.  @LeeTL1220 if you approve I could allow `ReadCountCollection` to be empty pretty trivially.

src/main/java/org/broadinstitute/hellbender/tools/exome/CnvCaller.java
davidbenjamin: I agree it's a code smell but I don't think the smell is that the underlying value is a vector but rather that the code takes in this inherently matrix quality (the `ReadCountCollection` of multiple samples) and refuses to work properly if there is more than one sample.  One overall smell in our codebase involving the redundancy between `TargetCoverage` and `ReadCountCollection` was that some tools *in the same pipeline* worked fine with multiple samples and some didn't.

In my opinion the solution would be to rewrite this tool and others to call multiple samples at once, just as the switch to `ReadCountCollection` now trivially makes `NormalizeSomaticReadCounts` work for multiple samples.

As for the name, I totally agree.  What do you think about renaming it the integer/double-agnostic `CoverageCollection`?

davidbenjamin: @samuelklee @LeeTL1220 come to think of it, it might make sense to have an abstract parent class for tools that do something for each sample/column of a read column collection, so that child classes would just have to implement `processColumn(final int column)` or something like that.

samuelklee: hmm, i'm still not quite sure about this---why do we need any of the tools to work with multiple samples (beside the obvious ones that deal with the PoN that require it or may see a performance boost)?

for example, what is gained by allowing multiple samples to be passed to the caller in a single file/object if they are each going to be called individually anyway?  shouldn't we just rely on the workflow scripts to take care of this?  especially considering that the natural output of nearly all of the steps will be a single file for each sample---e.g., plotting and segmentation (because of our reliance on the R script)---i'm not sure what is gained by artificially aggregating the inputs into a single file.  however, maybe i'm missing something about how multiple samples were treated in recapseg/previous iterations?

davidbenjamin: You have a point. . . I'm not really sure what my opinion is now.  I wouldn't mind if others provided one for me!

src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetBaseCallCoverage.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/exome/Genome.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/exome/ACNVCopyRatioModeller.java
davidbenjamin: @samuelklee What would you think about the more verbose but also more evocative `SingleSampleRecord`?

samuelklee: that works for me... @LeeTL1220 do you have any strong opinions?

davidbenjamin:
@LeeTL1220 Did everything we discussed in this repo, need to merge accompanying PR in gatk-public and re-run the pipeline.

davidbenjamin:
@droazen the build is failing due to `UnsatisfiedLinkError`s in `PairHMM`.  Do you know anything about this?

droazen:
@davidbenjamin The problem is we got overly excited about a multi-threaded version of the PairHMM, and turned it on before we had a graceful fallback in place in the case where the correct version of the required libraries are not available. We are going to revert this in public until such a fallback has been implemented, and publish a new snapshot.

davidbenjamin:
@droazen Thanks for the explanation.  I would have gotten overly excited too.

droazen:
@davidbenjamin I've released a new snapshot that reverts the PairHMM changes -- you should rebase this branch onto master once https://github.com/broadinstitute/gatk-protected/pull/509 is merged.

droazen:
Strike that -- since tests won't pass using the latest gatk-public without the changes here, you should take the one-line change in https://github.com/broadinstitute/gatk-protected/pull/509 and incorporate it directly into this branch.

davidbenjamin:
@LeeTL1220 I did everything we discussed, revved gatk-public, and reran the pipeline on the server as above.  Everything is fine.  Ready for your final review.

droazen:
Would be good to get this in soon, as it's blocking several other branches that require the new gatk-public snapshot.

src/main/java/org/broadinstitute/hellbender/tools/exome/ACNVCopyRatioModeller.java
LeeTL1220: I like ``SingleSampleRecord``

src/test/java/org/broadinstitute/hellbender/tools/exome/allelefraction/AlleleFractionSimulatedData.java
LeeTL1220: I would like an empty ``ReadCountCollection``

LeeTL1220:
@davidbenjamin Only one small change that came from @samuelklee

samuelklee:
Right, I think renaming ReadCountRecord.SingletonRecord and getting rid of the TRIVIAL_TARGETS if possible are the only issues left.

At some point I think it would be good to make sure we treat coverages and allelic counts identically and consistently.  For example, we use List<Target>, TargetCollection<ReadCountRecord.SingletonRecord>, ReadCountCollection, etc. and List<AllelicCount>, TargetCollection<AllelicCount>, AllelicCountCollection, etc.  If we do eventually give ReadCountCollection some of the capabilities of TargetCollection, as mentioned above, we should probably do the same for AllelicCountCollection (or have them both inherit them).

samuelklee:
Actually, I think I might still have some reservations about the vector vs. matrix issue too...something to keep in mind for future refactorings.

PR #453, opened 2016-04-25
Set segment-mean slice-sampling bounds.

### LINKED ISSUE #438: Crash in AllelicCNV due to infinite segment means
``java.lang.IllegalArgumentException: Segment Mean must be finite.``

Happens when trying to convert ACNV sim final seg files into ACS seg file.  Only happens to some samples.

samuelklee:
Closes #438.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/ACNVCopyRatioModeller.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/ACNVCopyRatioModeller.java
@@ -103,7 +103,6 @@ public final class ACNVCopyRatioModeller {
                 ll -= 0.5 * Math.log(newVariance) * numNotOutliers;
                 return ll;
             };
-
             return new SliceSampler(rng, logConditionalPDF, state.variance(),
                     VARIANCE_MIN, VARIANCE_MAX, varianceSliceSamplingWidth).sample();
         };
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/hellbender/tools/exome/ACNVCopyRatioModeller.java
```diff
@@ -126,22 +125,27 @@ public ACNVCopyRatioModeller(final SegmentedModel segmentedModel) {
...
+                            .mapToDouble(coverage -> -normalTerm(coverage, newMean, state.variance()))
                             .sum();
-
+                    //slice sample within range given by minimum and maximum non-outlier coverages
+                    final double minimumCoverageInSegment = Collections.min(nonOutlierCoveragesInSegment);
+                    final double maximumCoverageInSegment = Collections.max(nonOutlierCoveragesInSegment);
                     final SliceSampler sampler =
-                            new SliceSampler(rng, logConditionalPDF, state.getMeanInSegment(segment), meanSliceSamplingWidth);
+                            new SliceSampler(rng, logConditionalPDF, state.getMeanInSegment(segment),
+                                    minimumCoverageInSegment, maximumCoverageInSegment, meanSliceSamplingWidth);
```
LeeTL1220: It seems that all you did was take advantage of the min and max values of the SliceSampler.

PR #457, opened 2016-04-26
gradle extracts gatk-launch from the gatk jar

lbergelson:
adding a setupGatkLaunch task to gradle which extracts the gatk-launch script from the gatk dependency and places it in the root directory
updating gatk version

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,4 @@ foo.patch
 # odt temp file
 /docs/PoN/.~lock.TangentNormalizationInSparkNotes.odt#
 
+gatk-launch
--- a/build.gradle
+++ b/build.gradle
@@ -107,7 +107,7 @@ check.dependsOn installDist
 dependencies {
     compile files("${System.properties['java.home']}/../lib/tools.jar")
     compile 'org.apache.commons:commons-math3:3.5'
-    compile 'org.broadinstitute:gatk:broad-production-1-44-g230b706-SNAPSHOT'
+    compile 'org.broadinstitute:gatk:broad-production-1-48-g05b31ca-SNAPSHOT'
     compile 'com.opencsv:opencsv:3.4'
     compile 'org.ojalgo:ojalgo:39.0'
     compile ('org.apache.spark:spark-mllib_2.10:1.5.0') {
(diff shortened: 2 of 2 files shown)


build.gradle
```diff
@@ -173,6 +173,20 @@ configurations {
...
 
+def findJarByName(Configuration config, String name) {
+    config.find { File file -> file.name.startsWith(name)}
+}
+
+task setupGatkLaunch(type: Copy) {
+    def gatkJar = { findJarByName(configurations.compile, 'gatk') }
+    from {
+        zipTree( gatkJar ).matching{ include 'gatk-launch'}
+    }
```
droazen: Will this silently continue if there's no `gatk-launch` inside the jar?

.gitignore
```diff
@@ -39,3 +39,5 @@ foo.patch
 # odt temp file
 /docs/PoN/.~lock.TangentNormalizationInSparkNotes.odt#
 
+# gatk-launch
```
droazen: Remove uninformative comment? :)

build.gradle
```diff
@@ -173,6 +173,20 @@ configurations {
...
+    into "$projectDir"
+    outputs.file "$projectDir/gatk-launch"
```
droazen: Should the String `gatk-launch` be a named constant in case we rename it?

build.gradle
```diff
@@ -173,6 +173,20 @@ configurations {
     }
 }
 
+def findJarByName(Configuration config, String name) {
+    config.find { File file -> file.name.startsWith(name)}
+}
+
+task setupGatkLaunch(type: Copy) {
+    def gatkJar = { findJarByName(configurations.compile, 'gatk') }
```
droazen: Can you confirm that this won't match the `gatk-protected` jar itself?

droazen:
Review complete, back to @lbergelson, merge after addressing comments

build.gradle
lbergelson: I'm not sure how to check that actually...  I think it won't be a problem if it does because it won't have a gatk-launch.

lbergelson: well, we could check it by printlning I guess...

build.gradle
lbergelson: fixed

.gitignore
lbergelson: fixed

build.gradle
lbergelson: verified, only gatk public jar

build.gradle
lbergelson: fixed

PR #458, opened 2016-04-27
Fixed edge case for infinite segment-mean bug.

samuelklee:


src/main/java/org/broadinstitute/hellbender/tools/exome/ACNVCopyRatioModeller.java
```diff
@@ -141,8 +141,8 @@ public ACNVCopyRatioModeller(final SegmentedModel segmentedModel) {
                             .mapToDouble(coverage -> -normalTerm(coverage, newMean, state.variance()))
                             .sum();
                     //slice sample within range given by minimum and maximum non-outlier coverages
-                    final double minimumCoverageInSegment = Collections.min(nonOutlierCoveragesInSegment);
-                    final double maximumCoverageInSegment = Collections.max(nonOutlierCoveragesInSegment);
+                    final double minimumCoverageInSegment = Collections.min(allCoveragesInSegment);
```
LeeTL1220: Can you explain this?  As near as I can tell, this obviates the previous fix.

LeeTL1220:
@samuelklee Not sure I understand the change... seems to undo previous work.  Can you quickly explain?

samuelklee:
#453 set the bounds for slice sampling of the segment mean, but it used only non-outlier coverages to determine the min and max.  My guess is that we ran into trouble in a small segment where every target was an outlier, leading to an Collections.min/max throwing an error.  I simply changed the bounds to also include the outlier coverages.  (Note this choice then determines the meaning of the segment-mean parameter when there are no non-outlier coverages, which is otherwise not well defined...)

My guess is that a small segment in chr22 with many ~-31 log_2 coverages (in a deletion where the typical log_2 coverage was closer to ~-5) was the culprit in the failing sample.  So the question of what to do with these small coverages still remains.

LeeTL1220:
@samuelklee Possible solution:  adding one raw read to these empty targets...  Thoughts?  

Most hom deletions (with high purity and CCF) will have a smattering of reads, but some targets may have zero.  I.e.  log_2 coverage close to -5 for one or two reads, but -31 for zero reads.  De facto, we have essentially assumed that deletions will not have true zero read counts.  Or we truncate at something closer to log_2 coverage = -5 for zero reads.

samuelklee:
Another option would be to do a simple log_2(1 + x) transform instead.  However, both options still don't get around the artificially small variance.  I think this is something that the mean-dependent variance in the old model might have captured (purely by accident, since I don't think this relation is justified in general).

Another thought: we could also set the slice-sampling bounds to be given by the maximum of (minCoveragesInSegment, minAllowedSegmentMean) and the minimum of (maxCoveragesInSegment, maxAllowedSegmentMean).  This is effectively setting a bound on the uniform prior for the segment mean.  You'll note that the small segment in chr22 is now fitted with a log_2 segment mean of ~-31, since it is dominated by these outliers.  Although this is still a deletion, it's not at the same log_2 ~ -5 level as the surrounding regions.  The stricter prior could be used to prevent such small means.  However, we're still losing power by mislabeling these as -31 targets as "outliers"; they're outliers in the sense that they're not captured by the Gaussian, but they're clearly deletions.

However, at a practical level, the only reason we should be worried is if the downstream caller runs into trouble.  If I give it 2 segments with log_2 segment means of -31 and -5, is it going to split them into two different states or will it lump them together?

LeeTL1220:
@samuelklee The current caller looks at the target CR variance around the estimated copy neutral point.  So, it will not care about two segments at -31 and -5, since it will just call those as deletions.  I'd be surprised if a purity/ploidy model would care that much, since it should just look like noise around an absolute copy number of zero and very high rho (purity * CCF).

PR #462, opened 2016-04-27
add system properties and deflater to build

akiezun:
all the right properties from public were missing in protected. Also deflater binary was missing.

@droazen please have a look


akiezun:
@lbergelson please review

lbergelson:
@akiezun Looks good to me.

PR #481, opened 2016-04-30
eliminate eager string construction in assertion. Takes 3% of runtime of HC

akiezun:
The String was built regardless of whether the test fails or succeeds which was wasteful. Shows up on HC profile.

@lbergelson, ok?


src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/RefVsAnyResult.java
```diff
@@ -56,12 +54,16 @@ public RefVsAnyResult(final int likelihoodCapacity) {
     }
 
     void incrementRefAD(final int by){
-        Utils.validateArg(by >= 0, "expected a non-negative number but got " + by);
```
lbergelson: Bleh, what a pain.  I wish there was a way to have this lazily evaluated.

PR #483, opened 2016-05-02
Expanded documentation and removing more contigs when requested.

### LINKED ISSUE #471: SparkGenomeReadCount should have more intelligent blacklisting that can support other genome builds.
Currently, the tool can only exclude by X,Y, MT contigs.  This is okay for humans, but not necessarily useful for other reference genomes.

LeeTL1220:
Added more documentation for -L and making integration  test for basic check.

Closes #471

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/genome/SparkGenomeReadCounts.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/genome/SparkGenomeReadCounts.java
@@ -28,7 +28,8 @@ import java.util.stream.Collectors;
 @CommandLineProgramProperties(
         summary = "Calculate coverage on a WGS bam file using Spark.  This creates a set of pseudo-targets that span" +
                 "the entire genome.  Use the 'binsize' parameter to specify the size of each interval.  By default, any " +
-                "contigs X, Y, M, and MT are excluded.",
+                "contigs X, Y, M, and MT are excluded.\n" +
+        "Please see the " + SparkGenomeReadCounts.DROP_NON_AUTOSOMES_LONG_NAME + " option if using this tool on a non-human genome.",
         oneLineSummary = "Calculate coverage on a WGS bam file using Spark",
         programGroup = CopyNumberProgramGroup.class)
 public class SparkGenomeReadCounts extends GATKSparkTool {
--- a/src/main/java/org/broadinstitute/hellbender/tools/genome/SparkGenomeReadCounts.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/genome/SparkGenomeReadCounts.java
@@ -13,7 +13,6 @@ import org.broadinstitute.hellbender.engine.filters.ReadFilter;
 import org.broadinstitute.hellbender.engine.filters.ReadFilterLibrary;
 import org.broadinstitute.hellbender.engine.filters.WellformedReadFilter;
 import org.broadinstitute.hellbender.engine.spark.GATKSparkTool;
-import org.broadinstitute.hellbender.engine.spark.datasources.ReadsSparkSource;
 import org.broadinstitute.hellbender.exceptions.UserException;
 import org.broadinstitute.hellbender.tools.exome.SampleCollection;
 import org.broadinstitute.hellbender.tools.exome.Target;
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/hellbender/tools/genome/SparkGenomeReadCounts.java
```diff
@@ -42,7 +43,7 @@
     protected static final String DROP_NON_AUTOSOMES_SHORT_NAME = "noxy";
     protected static final String DROP_NON_AUTOSOMES_LONG_NAME = "noXYMT";
 
-    @Argument(doc = "Remove X, Y and MT regions",
+    @Argument(doc = "Remove X, Y, GL*, NC_*, and MT regions, regardless of intervals specified.  Use -L (or -XL) and disable this option for exact specification of intervals.  This option may be removed in the future.",
```
lbergelson: It's weird to have a boolean flag that defaults to true.   It would probably be better to invert this.   The change to the documentation looks good though.

src/main/java/org/broadinstitute/hellbender/tools/genome/SparkGenomeReadCounts.java
```diff
@@ -172,7 +174,10 @@ private void collectReads(final JavaSparkContext ctx) {
 
     protected static SAMSequenceDictionary dropContigsFromSequence(final SAMSequenceDictionary originalSequenceDictionary, final Set<String> nonAutosomalContigs) {
         final List<SAMSequenceRecord> initialSequences = new ArrayList<>(originalSequenceDictionary.getSequences());
-        final List<SAMSequenceRecord> finalSequences = initialSequences.stream().filter(s -> !nonAutosomalContigs.contains(s.getSequenceName())).collect(Collectors.toList());
+        final List<SAMSequenceRecord> finalSequences = initialSequences.stream()
+                .filter(s -> !nonAutosomalContigs.contains(s.getSequenceName()))
+                .filter(s -> !StringUtils.startsWith(s.getSequenceName(), "GL") && !StringUtils.startsWith(s.getSequenceName(), "NC_"))
```
lbergelson: I'm probably missing something here, but why not just String.startswith?

src/main/java/org/broadinstitute/hellbender/tools/genome/SparkGenomeReadCounts.java
```diff
@@ -71,7 +72,7 @@
     )
     protected File outputFile;
 
-    private void collectReads(final JavaSparkContext ctx) {
+    public void collectReads(final JavaSparkContext ctx) {
```
lbergelson: Did this need to become public?  I don't see an out of class use of it.

lbergelson:
@LeeTL1220 Tests are failing.  Not certain why.

lbergelson:
Minor comments.  Merge when ready.

src/main/java/org/broadinstitute/hellbender/tools/genome/SparkGenomeReadCounts.java
LeeTL1220: Accident.  Done.

src/main/java/org/broadinstitute/hellbender/tools/genome/SparkGenomeReadCounts.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/tools/genome/SparkGenomeReadCounts.java
LeeTL1220: Got in the habit of using StringUtils, for startsWith, from way back when

Done

LeeTL1220:
@lbergelson Waiting on outcome of (offline) discussion with you and @droazen before I merge.

LeeTL1220:
@lbergelson @droazen Automated test seems to be failing because ``TraversalParameters`` (in GATK public) is not serializable.  Or at least this is what the error message would have me believe.  Oh, and the class declaration...

``public class TraversalParameters { .... }``

LeeTL1220:
@lbergelson There is a bit more to look at, since we now support intervals in this tool.  I've responded to previous comments.

src/main/java/org/broadinstitute/hellbender/tools/genome/SparkGenomeReadCounts.java
```diff
@@ -71,12 +69,41 @@
...
+
+        if (keepNonAutosomes) {
+            return modifiedIntervals;
+        }
+
+        // Enforce the elimination of certain contigs when proper option is set.
+        logger.info("Dropping non-autosomes, as requested...");
+        return modifiedIntervals.stream()
+                .filter(s -> !NONAUTOSOMALCONTIGS.contains(s.getContig()))
+                .filter(s -> !(s.getContig().startsWith("GL")) && !(s.getContig().startsWith("NC_")))
```
lbergelson: It still seems a bit strange that it has an exclusion list rather than a human contigs inclusion list. Things like HLA contigs,  mitochondria, etc will pass this filter. Maybe doc should mention that it's assuming human hg19

lbergelson:
@LeeTL1220 One comment, but you know more about the usage than I do.  👍

src/main/java/org/broadinstitute/hellbender/tools/genome/SparkGenomeReadCounts.java
LeeTL1220: Mitochondria will not.

HLA will, but nobody has been filtering that out....

LeeTL1220: The docs are very explicit about what gets excluded.  

Added a little more docs.

PR #490, opened 2016-05-04
Added a method for HMMs to generate a random hidden sequence. Closes #456.

### LINKED ISSUE #456: Generate() method for HMM
For tests it would be very useful if the `HiddenMarkovModel` interface had a default method to generate observed data.  The signature would be something like `generate(final List<T> positions, final Function<S, D> emission)`, where `emission` is a function that randomly generates observed data given a hidden state.

davidbenjamin:
This is going to make writing tests for the germline (and possibly later somatic) segmentation easier.  As an example, this PR already uses the new method to simplify a lot of code in the germline integration tests.

@samuelklee would you mind taking a look at this?

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/utils/GATKProtectedMathUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/utils/GATKProtectedMathUtils.java
@@ -1,13 +1,17 @@
 package org.broadinstitute.hellbender.utils;
 
+import org.apache.commons.math3.distribution.EnumeratedDistribution;
 import org.apache.commons.math3.linear.RealMatrix;
+import org.apache.commons.math3.random.RandomGenerator;
 import org.apache.commons.math3.stat.descriptive.moment.Variance;
 import org.apache.commons.math3.stat.descriptive.rank.Percentile;
-import org.apache.commons.math3.util.FastMath;
+import org.apache.commons.math3.util.Pair;
 
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.List;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
 /**
@@ -15,7 +19,6 @@ import java.util.stream.IntStream;
  * Created by davidben on 1/22/16.
  */
 public class GATKProtectedMathUtils {
-
     /**
      * Computes $\log(\sum_i e^{a_i})$ trying to avoid underflow issues by using the log-sum-exp trick.
      *
@@ -149,5 +152,13 @@ public class GATKProtectedMathUtils {
     public static double stdDev(final double ... values) {
         Utils.nonNull(values);
         return Math.sqrt(new Variance().evaluate(values));
+    }
+
+    // given a list of options and a function for calculating their probabilities (these must sum to 1 over the whole list)
+    // randomly choose one option from the implied categorical distribution
+    public static <E> E randomSelect(final List<E> choices, final Function<E, Double> probabilityFunction, final RandomGenerator rng) {
+        final List<Pair<E, Double>> pmf = choices.stream()
+                .map(e -> new Pair<>(e, probabilityFunction.apply(e))).collect(Collectors.toList());
+        return new EnumeratedDistribution<>(rng, pmf).sample();
     }
 }
--- a/src/test/java/org/broadinstitute/hellbender/utils/GATKProtectedMathUtilsTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/utils/GATKProtectedMathUtilsTest.java
@@ -1,10 +1,17 @@
 package org.broadinstitute.hellbender.utils;
 
 import org.apache.commons.math3.linear.Array2DRowRealMatrix;
+import org.apache.commons.math3.random.RandomGenerator;
+import org.apache.commons.math3.random.RandomGeneratorFactory;
 import org.testng.Assert;
 import org.testng.annotations.Test;
 
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.List;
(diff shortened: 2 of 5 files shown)


src/main/java/org/broadinstitute/hellbender/utils/GATKProtectedMathUtils.java
```diff
@@ -136,4 +141,22 @@ public static double stdDev(final double ... values) {
         Utils.nonNull(values);
         return Math.sqrt(new Variance().evaluate(values));
     }
+
+    // given a list of options and a function for calculating their probabilities (these must sum to 1 over the whole list)
+    // randomly choose one option from the implied categorical distribution
+    public static <E> E randomSelect(final List<E> choices, final Function<E, Double> probabilityFunction) {
```
samuelklee: Since this is pretty standard, it seems like you should make use of EnumeratedDistribution in Apache Commons (or something along those lines) here.  You could then pass a RandomGenerator (which you could reset before each test/method that calls this, avoiding those problems on Travis that we ran into before), rather than using the private static Random (which you cannot reset in this code as written).

davidbenjamin: First off, I won't pretend that I wrote it as I did for any reason other than ignorance of the library version!

The implementation of `EnumeratedDistribution` is not lazy -- you have to pre-calculate the pmf -- whereas the way I wrote it you only have to calculate until you make a selection.  For cases where your first state has a large weight, such as germline segmentation where the neutral state is WLOG the first, this is a significant optimization.  Then again, we only plan on using this in tests, and also my code assumes a normalized pmf, so perhaps this optimization is unnecessary.  Furthermore, in germline segmentation there are so few states and the expensive part of the transition matrix is cached after one calculation, so the speedup is strictly less than 3x. 

All in all, I think I agree with your suggestion.

samuelklee: That's a fair point about the optimization.  If you want to keep your implementation, though, I still think passing the rng would be a good idea.

src/test/java/org/broadinstitute/hellbender/utils/GATKProtectedMathUtilsTest.java
```diff
@@ -113,4 +117,27 @@ public void testColumnStdDevs() {
...
+        final int NUM_SAMPLES = 1000;
+        final List<Integer> choices = Arrays.asList(-1,0,1);
+        final List<Integer> result = IntStream.range(0, NUM_SAMPLES)
+                .map(n -> GATKProtectedMathUtils.randomSelect(choices, j -> j*j/2.0))
+                .boxed()
+                .collect(Collectors.toList());
+        Assert.assertEquals(result.stream().filter(n -> n==0).count(), 0);
+        Assert.assertEquals(result.stream().filter(n -> n==1).count(), NUM_SAMPLES/2, 50);
+    }
+
```
samuelklee: white space

src/test/java/org/broadinstitute/hellbender/utils/hmm/HMMUnitTest.java
```diff
@@ -807,6 +807,53 @@ public void testViterbiWithNullDataPoints() {
...
+        final int chainLength = 100000;
+        final List<Integer> positions = IntStream.range(0, chainLength).boxed().collect(Collectors.toList());
+
+        final List<Integer> chain = model.generateHiddenStateChain(positions);
+
+        Assert.assertEquals(chain.size(), positions.size());
+        final int[] counts = new int[numStates];
+        chain.stream().forEach(n -> counts[n]++);
+
+        final int expectedCountsPerSTate = chainLength / numStates;
```
samuelklee: expectedCountsPerSTate -> expectedCountsPerState

src/test/java/org/broadinstitute/hellbender/utils/hmm/HMMUnitTest.java
```diff
@@ -807,6 +807,53 @@ public void testViterbiWithNullDataPoints() {
...
+    @Test
+    public void testGenerateHiddenStateChainOnUninformativeHMM() {
+        final int numStates = 3;
+        final UninformativeTestHMModel model = new UninformativeTestHMModel(numStates);
+
+        final int chainLength = 100000;
+        final List<Integer> positions = IntStream.range(0, chainLength).boxed().collect(Collectors.toList());
+
+        final List<Integer> chain = model.generateHiddenStateChain(positions);
+
```
samuelklee: this is super nitpicky, but the white space here is not quite the same as in the second test...

samuelklee:
i definitely think that the random seed should be reset at the beginning of each test, but otherwise looks good!  back to you, @davidbenjamin

src/test/java/org/broadinstitute/hellbender/utils/GATKProtectedMathUtilsTest.java
davidbenjamin: fixed

src/test/java/org/broadinstitute/hellbender/utils/hmm/HMMUnitTest.java
davidbenjamin: fixed

src/test/java/org/broadinstitute/hellbender/utils/hmm/HMMUnitTest.java
davidbenjamin: Now that you mention it, that looks horrible.  Fixed.

davidbenjamin:
@samuelklee back to you.  I went with everything you said; no point optimizing test code.

samuelklee:
looks good!

PR #496, opened 2016-05-05
adding a .gitattributes file for git-lfs

lbergelson:
this is necessary to make git-lfs work correctly


lbergelson:
@LeeTL1220 This should fix your problem.

@droazen Did this not get committed initially with the git-lfs files for some reason?

droazen:
Oh dear, my mistake -- merge when ready @lbergelson!

PR #503, opened 2016-05-12
Added option to disable or perform less MCMC refitting during similar-segment merging.

### LINKED ISSUE #499: POS -> POSITION change in pulldown column name breaks plotting.
This change was made in #444.  Also breaks evaluation scripts.  @LeeTL1220, is backwards compatibility with previous pulldowns desired?  See also #451.

samuelklee:
Exposes more control over MCMC refitting performed during similar-segment merging, which can be useful for development.

Similar-segment merging is currently performed by traversing the genome and merging adjacent segments that have credible intervals that overlap in CR and MAF by a specified amount.  The credible intervals of a merged segment are updated in a heuristic manner.  After traversal, an MCMC fit is performed.  This procedure is repeated until convergence.

Because each MCMC fit is expensive, this yields longer runtimes for samples that are initially very segmented (e.g., LUAD).  For development purposes, it is nice to have more control over this runtime (beyond just limiting the number of iterations of merging allowed, which is already possible).  With the new argument numIterationsSimSegPerFit, one can choose to only refit intermittently.  Evaluation shows that although less segmentation is smoothed out when a refit is not performed after every merge iteration, HAPSEG concordance is not drastically reduced.

This feature should still be useful even if we update the method of similar-segment merging in the future (hopefully to something less heuristic), since I expect it should remain a relatively expensive step.

Also added some basic validation of CLI arguments.

@achevali, note that this also closes #499.

### DIFF
--- a/src/test/java/org/broadinstitute/hellbender/tools/exome/ACNVModellerUnitTest.java
+++ b/src/test/java/org/broadinstitute/hellbender/tools/exome/ACNVModellerUnitTest.java
@@ -15,7 +15,8 @@ import java.util.Collections;
 import java.util.List;
 
 /**
- * Unit tests for {@link ACNVModeller}.
+ * Unit tests for {@link ACNVModeller}.  Note that behavior of the tests depends on the content of the input files
+ * and that changing them may break the tests.
  *
  * @author Samuel Lee &lt;slee@broadinstitute.org&gt;
  */
@@ -29,15 +30,20 @@ public final class ACNVModellerUnitTest extends BaseTest {
     private static final File TUMOR_ALLELIC_COUNTS_FILE = new File(TEST_SUB_DIR
             + "snps-for-acnv-modeller.tsv");
     private static final File SEGMENT_FILE =
-            new File(TEST_SUB_DIR + "segments-for-acnv-modeller.seg");
+            new File(TEST_SUB_DIR + "segments-for-acnv-modeller.seg");  //44 segments (some spurious)
     private static final File SEGMENTS_TRUTH_FILE = new File(TEST_SUB_DIR
-            + "segments-truth-for-acnv-modeller.seg");
+            + "segments-truth-for-acnv-modeller.seg");  //30 true segments
 
     private static final int NUM_SAMPLES = 100;
     private static final int NUM_BURN_IN = 50;
 
     private static final double INTERVAL_THRESHOLD = 1.; //merge when parameter modes are within 95% HPD interval
     private static final int MAX_SIMILAR_SEGMENT_MERGE_ITERATIONS = 25;
+
+    private static final List<SimpleInterval> SEGMENTS_TRUTH = SegmentUtils.readIntervalsFromSegmentFile(SEGMENTS_TRUTH_FILE);
+
+    //similar-segment merging without refitting may miss some merges, depending on the input test data; we allow up to 2 misses
+    private static final int DELTA_NUM_SEGMENTS_WITHOUT_REFITTING = 2;
 
     /**
      * Test of similar-segment merging using only copy-ratio data (simulated coverages and segments).
@@ -46,6 +52,7 @@ public final class ACNVModellerUnitTest extends BaseTest {
      */
     @Test
     public void testMergeSimilarSegmentsCopyRatio() throws IOException {
+        final boolean doRefit = true;
         final JavaSparkContext ctx = SparkContextFactory.getTestSparkContext();
         LoggingUtils.setLoggingLevel(Log.LogLevel.INFO);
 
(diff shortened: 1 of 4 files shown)


samuelklee:
@mbabadi, mind taking a look at this when you get a chance?

src/main/resources/org/broadinstitute/hellbender/utils/plotter/CNV_plotting_library.R
```diff
@@ -57,8 +57,8 @@ PlotAlleleFraction = function(snp_df, segments_df){
      contig = segments_df[s,"Chromosome"]
      segment_start = segments_df[s, "Start"]
      segment_end = segments_df[s, "End"]
-     snp_indices=which(snp_df[,"CONTIG"] == contig & snp_df[,"POS"] > segment_start & snp_df[,"POS"] < segment_end)
-     genomic_coordinates = chromosome_starts[contig] + snp_df[ snp_indices,"POS"]
+     snp_indices=which(snp_df[,"CONTIG"] == contig & snp_df[,"POSITION"] > segment_start & snp_df[,"POSITION"] < segment_end)
```
mbabadi: This is already in master.

src/test/java/org/broadinstitute/hellbender/tools/exome/ACNVModellerUnitTest.java
```diff
@@ -119,4 +125,49 @@ public void testMergeSimilarSegments() {
...
+        Assert.assertTrue(!modeller.isModelFit());
+
+        //write final model fit to file; this should force model refit
+        final File finalModeledSegmentsFile =
+                new File(tempDirFile.getAbsolutePath() + "/test-" + FINAL_SEG_FILE_TAG + ".seg");
+        modeller.writeACNVModeledSegmentFile(finalModeledSegmentsFile);
+        Assert.assertTrue(modeller.isModelFit());
+
+        //check final number of segments = 31 (ground truth has 30 segments, so we miss one merge by not performing intermediate refits)
+        final int expectedNumberOfSegments = 31;
```
mbabadi: Magic number and depends on the input data. Please assign a constant to it in the preamble or better yet, make a truth segmented model file and test equality against it (similar to what you do in `testMergeSimilarSegments`). Also, please consider renaming `testMergeSimilarSegments` to `testMergeSimilarSegmentsWithRefitting`.

src/test/java/org/broadinstitute/hellbender/tools/exome/AllelicCNVIntegrationTest.java
```diff
@@ -63,6 +101,64 @@ public void testACNVWithAllelicPON() {
...
+                "--verbosity", "INFO",
+        };
+        testACNV(arguments, outputPrefix);
+
+        //check that some similar segments were merged
+        final File initialSimilarSegmentsFile = new File(outputPrefix + "-" + AllelicCNV.INITIAL_SEG_FILE_TAG + ".seg");
+        final File finalSimilarSegmentsFile = new File(outputPrefix + "-" + AllelicCNV.FINAL_SEG_FILE_TAG + ".seg");
+        final List<ACNVModeledSegment> initialACNVModeledSegments = SegmentUtils.readACNVModeledSegmentFile(initialSimilarSegmentsFile);
+        final List<ACNVModeledSegment> finalACNVModeledSegments = SegmentUtils.readACNVModeledSegmentFile(finalSimilarSegmentsFile);
+        Assert.assertTrue(initialACNVModeledSegments.size() > finalACNVModeledSegments.size());
```
mbabadi: Please write a warning note in the header of `AllelicCNVIntegrationTest` mentioning that the behavior of the tests depend on the _content_ of the input files and changing them may break the tests.

mbabadi:
Back to you @samuelklee. The comments are mere suggestions. Feel free to rebase and merge.

src/test/java/org/broadinstitute/hellbender/tools/exome/ACNVModellerUnitTest.java
samuelklee: Good catch.  Now that I think about it, I'm not sure what is actually a satisfying test here.  Maybe just require that merging without refitting gets to within a certain threshold of the expected number of segments that merging with refitting finds?  In any case, I'll add comments and get rid of the magic number.

mbabadi: Well it's difficult to devise a bullet proof test here -- specially if we happen to change some details of the segment merging algorithm in the future (the behavior of the full algorithm with and without refitting is difficult to predict). I think it should be sufficient to warn about it in the header doc :)

PR #505, opened 2016-05-14
add mavenLocal to gradle

akiezun:
add mavenLocal to gradle to simplify 2-repo development.

with this change i just need to add 1 line when changing the snapshot build of gatk

I put local before central because it seems to me that local copies should override global ones. conflicts should be very rare
@lbergelson wdyt?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -78,7 +78,7 @@ repositories {
             dirs customJarPath
         }
     }
-
+    mavenLocal()
 }
 
 jacocoTestReport {


droazen:
@akiezun I'd feel safer if `mavenLocal()` were last here -- we want builds of the GATK to be repeatable in general across machines. Moving it to last would not harm your use case of using local snapshots of GATK public.

akiezun:
ok

akiezun:
back to @droazen - moved local to last

build.gradle
```diff
@@ -78,7 +78,7 @@ repositories {
             dirs customJarPath
         }
     }
-
+    mavenLocal()
```
akiezun: testing github/hipchat integration

droazen:
:+1: merging

PR #518, opened 2016-05-18
Removed sex chromosomes from pulldowns by default and added option to GetHetCoverage and GetBayesianHetCoverage to include them.

### LINKED ISSUE #355: Het Pulldown in acnv finds sites on nonhomolgous Y

samuelklee:
Also some minor cleanup.

This is a quick fix that simply filters on "X" and "Y" in the contig name.  Note that this is not analogous to how sex chromosomes are handled in coverage collection, where we only collect coverage at targets specified in a file---which does not include targets on X and Y.  So the analogous thing to do would be to exclude SNPs on X and Y from the common SNP list we use to collect het coverage.

Also note that I didn't include a proper integration test for GetHetCoverage/GetBayesianHetCoverage because hg19mini.fasta doesn't include X or Y.

Closes #355.


samuelklee:
@LeeTL1220 can you take a look at this?  I'm not actually convinced we should merge this (see comments above), but let me know what you think.

LeeTL1220:
@samuelklee As per our discussion, are we going to keep this one open?

samuelklee:
Let's close it without merging for now.

PR #539, opened 2016-06-06
Rev gatk dependency to snapshot of latest master (with dual mac/linux builds of the native PairHmm)

droazen:


droazen:
@lbergelson Can you review, and please advise on whether you think any changes are needed in protected's build system as a result of https://github.com/broadinstitute/gatk/pull/1846 and https://github.com/broadinstitute/gatk/pull/1863

lbergelson:
@droazen Looks fine to me.  I don't know if you want to remove the ml-lib dependency in gatk-protected now that it comes in transitively.  Being explicit should be fine.  We'll have to update or remove it if we bump it in gatk though.  

I would test running with spark submit on the cluster to be sure we haven't accidentally broken something.  The dependency changes don't always show up as problems in our test suite.

akiezun:
I'd remove the ml dependency from protected. The point of putting it in
public is to simplify the life of downstream projects. Can be another
ticket though because it may require some time to figure out how/what to
remove

droazen:
@lbergelson I tested on our on-prem cluster with `PrintReadsSpark` and it ran fine. `BaseRecalibratorSpark` was having out-of-memory issues, however, even with ridiculous amounts of memory per core (@akiezun you may want to check on this!)

droazen:
@akiezun I'll create a separate ticket to capture the task of removing cruft from protected's build file.

PR #543, opened 2016-06-07
less array allocation in PairHMMLikelihoodCalculationEngine

akiezun:
protected part of https://github.com/broadinstitute/gatk/pull/1822
reduces the amount of memory allocation by reusing the whole array and playing with indices rather than cutting slices of the array.

@lbergelson can you have a look?

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/PairHMMLikelihoodCalculationEngine.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/PairHMMLikelihoodCalculationEngine.java
@@ -3,6 +3,8 @@ package org.broadinstitute.hellbender.tools.walkers.haplotypecaller;
 import com.google.common.annotations.VisibleForTesting;
 import htsjdk.samtools.SAMUtils;
 import org.apache.commons.lang3.tuple.Pair;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.broadinstitute.hellbender.exceptions.GATKException;
 import org.broadinstitute.hellbender.utils.MathUtils;
 import org.broadinstitute.hellbender.utils.QualityUtils;
@@ -28,6 +30,8 @@ import java.util.stream.Collectors;
  * Classic likelihood computation: full pair-hmm all haplotypes vs all reads.
  */
 public final class PairHMMLikelihoodCalculationEngine implements ReadLikelihoodCalculationEngine {
+
+    private static final Logger logger = LogManager.getLogger(PairHMMLikelihoodCalculationEngine.class);
 
     private static final int MAX_STR_UNIT_LENGTH = 8;
     private static final int MAX_REPEAT_LENGTH   = 20;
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/PairHMMLikelihoodCalculationEngine.java
```diff
@@ -338,7 +342,7 @@ void applyPCRErrorModel( final byte[] readBases, final byte[] readInsQuals, fina
     @VisibleForTesting
     static Pair<byte[], Integer> findTandemRepeatUnits(final byte[] readBases, final int offset) {
         int maxBW = 0;
-        byte[] bestBWRepeatUnit = new byte[]{readBases[offset]};
+        byte[] bestBWRepeatUnit = {readBases[offset]};
```
lbergelson: This is just shorter syntax because it can infer the array type now?

akiezun: yes, just syntax

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/PairHMMLikelihoodCalculationEngine.java
```diff
@@ -347,19 +351,20 @@ void applyPCRErrorModel( final byte[] readBases, final byte[] readInsQuals, fina
             }
 
             // get backward repeat unit and # repeats
-            final byte[] backwardRepeatUnit = Arrays.copyOfRange(readBases, offset - str + 1, offset + 1);
-            maxBW = GATKVariantContextUtils.findNumberOfRepetitions(backwardRepeatUnit, Arrays.copyOfRange(readBases, 0, offset + 1), false);
+            maxBW = GATKVariantContextUtils.findNumberOfRepetitions(readBases, offset - str + 1,  str , readBases, 0, offset + 1, false);
```
lbergelson: This looks correct to me but this is a confusing method.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/PairHMMLikelihoodCalculationEngine.java
```diff
@@ -381,12 +385,12 @@ void applyPCRErrorModel( final byte[] readBases, final byte[] readInsQuals, fina
...
             } else {
                 // tandem repeat starting forward from current offset.
-                // It could be the case that best BW unit was differnet from FW unit, but that BW still contains FW unit.
+                // It could be the case that best BW unit was different from FW unit, but that BW still contains FW unit.
                 // For example, TTCTT(C) CCC - at (C) place, best BW unit is (TTC)2, best FW unit is (C)3.
                 // but correct representation at that place might be (C)4.
                 // Hence, if the FW and BW units don't match, check if BW unit can still be a part of FW unit and add
                 // representations to total
-                maxBW = GATKVariantContextUtils.findNumberOfRepetitions(bestFWRepeatUnit, Arrays.copyOfRange(readBases, 0, offset + 1), false);
+                maxBW = GATKVariantContextUtils.findNumberOfRepetitions(bestFWRepeatUnit, testString, false);
```
lbergelson: It seems like you're potentially doing an extra copy by pulling `testString` out of this block.  Why not initialize it here in the `else` block?

lbergelson:
@akiezun Looks good.  The new index twiddling method is less readable than the old one.  Oh well.  
One thing to potentially address.  Merge when ready.

PR #544, opened 2016-06-07
remove mllib from build script

### LINKED ISSUE #540: Remove explicit ml-lib dependency (and related dependencies) from build.gradle
As of https://github.com/broadinstitute/gatk-protected/pull/539, we get this dependency transitively from gatk public, so we should remove it from protected's build file, both for the sake of simplicity and to avoid versioning conflicts across public and protected.

akiezun:
gatk public provides it so i removed it from here. 

fixes #540 

@lbergelson have a look

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -123,16 +123,8 @@ check.dependsOn installDist
 
 dependencies {
     compile files("${System.properties['java.home']}/../lib/tools.jar")
-    compile 'org.apache.commons:commons-math3:3.5'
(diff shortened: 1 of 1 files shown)


lbergelson:
@akiezun Did you try launching any spark jobs on the cluster with this change?  Sometimes the dependencies failures don't show up in our tests suite because it's an interaction with the cluster environment.   👍 when you've tested that.

akiezun:
I did run MarkDups on the cloud cluster (run from protected)

PR #545, opened 2016-06-07
Emit pre/post/d/scaled_d qc files

achevali:


LeeTL1220:
@achevali Please add an automated test to check for existence of the three files and that the size is >0.  Also, you should check that ``postQC-preQC = dQC`` (or the other way around).

LeeTL1220:
``ParamUtils.readValuesFromFile(final File inputFile)``

```
final double postQC = ParamUtils.readValuesFromFile(.....post_Qc)
final double preQC = ParamUtils.readValuesFromFile(.....pre_Qc)
final double dQC = ParamUtils.readValuesFromFile(.....dQc)

// I may have post and pre swapped
Assert.assertEquals(postQC-preQC, dQC)
```

achevali:
Tests added, back to you.

PR #548, opened 2016-06-08
less memory allocation in RCM by reaching directly to bases and quals

akiezun:
I reduced the amount of memory allocated in HaplotypeCaller by by reaching directly to bases and quals in ReferenceConfidenceModel

@lbergelson can you review?


lbergelson:
@akiezun 👍

PR #552, opened 2016-06-10
Updated coverage docs

mbabadi:
The new notes includes several updates to the previous version, in particular, the introduction of a regularization scheme to ensure that the CNV events do not get picked up as "noise".


mbabadi:
@davidbenjamin, please kindly review. I am personally very intrigued by the results of the regularized model! :) @samuelklee, please also kindly comment.

PR #560, opened 2016-06-15
Pegging docker to alpha1.3

### LINKED ISSUE #386: Update Dockerfile to use tagged release rather than master

### LINKED ISSUE #501: Have NormalizeSomaticReadCounts check version number of the PoN
It should at least throw a warning if it detects that the version of the PoN is not the current one.  This should handle some user errors in advance...

LeeTL1220:
Closes #386 
Closes #501 

**Please note that this pull request assumes that the next release will be alpha-1.3 for the docker image.  It was tested on alpha-1.2**

Also, added warning if the pon version does not match the latest version in NormalizeReadCounts.

Adding warning if the PoN is not the current version in NormalizeSomaicReadCounts.

Setting to be the next version number of gatk-protected.

Adding plotting error dump to gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+/plotting_dump.rda
+
 # Mobile Tools for Java (J2ME)
 .mtj.tmp/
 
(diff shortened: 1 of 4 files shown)


lbergelson:
👍

PR #564, opened 2016-06-16
Fixing tau issue

### LINKED ISSUE #563: ACS conversion is erroneous for tau
The conversion is incorrect

LeeTL1220:
Closes #563


src/main/java/org/broadinstitute/hellbender/tools/exome/acsconversion/ACSModeledSegmentUtils.java
```diff
@@ -112,7 +112,7 @@ public static ACSModeledSegment convertACNVSegmentToACSSegment(final ACNVModeled
         final int targetCount = genome.getTargets().targetCount(acnvModeledSegment.getInterval());
         final int segLabelCNLOH = 2;
         final int hetCount = genome.getSNPs().targetCount(acnvModeledSegment.getInterval());
-        return new ACSModeledSegment(acnvModeledSegment.getInterval(), ModeledSegment.NO_CALL, targetCount, tau/2.0,
+        return new ACSModeledSegment(acnvModeledSegment.getInterval(), ModeledSegment.NO_CALL, targetCount, ParamUtils.log2(tau/2.0),
```
samuelklee: This is somewhat confusing.  I think the local variable should be `segmentMeanInLog2CR = acnvModeledSegment.getSegmentMeanPosteriorSummary().getCenter()` (since this is what the ACSModeledSegment constructor takes), rather than `tau`.  This obviates the need to perform and then immediately undo the transformation locally, which is what you currently do.

samuelklee:
I think some additional documentation in both `ACSModeledSegment` and `ACSModeledSegmentUtils` on the definition of `tau` would prevent future confusion.

samuelklee:
some minor comments, back to you @LeeTL1220

samuelklee:
Oh, also, are we sure that everything is kosher with sigmaTau?

LeeTL1220:
I confirmed a long time ago that sigmaTau is okay.  This was an issue with forgetting to specify the value in log2 space.

samuelklee:
OK, good to squash and merge!

LeeTL1220:
Answered issues.  Back to you @samuelklee

PR #566, opened 2016-06-17
Enabling the index.

### LINKED ISSUE #561: Use the index for SamLocusIterator in GetBayesianHetPulldown and test that this improves speed.
Needed for #395

LeeTL1220:
Closes #561 

Very simple change.  After running on WGS samples , speed did not seem to improve much.  Leaving in the change, since, according to the javadoc, the index should not hurt performance.


davidbenjamin:
That was an easy review.

PR #569, opened 2016-06-17
Fixed log10 issue in CopyRatioModeller and made unit test more stringent.

samuelklee:


samuelklee:
@LeeTL1220, if you could hold off on cutting a release until this is in, that'd be great!  This fixes a log10 vs ln bug that snuck into the copy-ratio modeller at some point (when it was originally ACNVCopyRatioModeller), which was throwing off the inference of the variance and outlier probability parameters.  Looking at the git history, I think this may have actually been present before the HAPSEG concordance evaluation was done, but I don't think it affected the results too much.

src/main/java/org/broadinstitute/hellbender/tools/exome/copyratio/CopyRatioSamplers.java
```diff
@@ -138,12 +139,11 @@ public OutlierIndicatorsSampler(final double outlierUniformLogLikelihood) {
...
                             notOutlierUnnormalizedLogProbabilityPrefactor
                                     - normalTerm(c.getCoverage(), state.segmentMean(segment), state.variance());
-                    //note: we are working in natural log space, but this differs from log10 by a multiplicative
-                    //constant which is absorbed into the normalization.  Thus normalizeFromLog10 works here.
+                    //note: we are working in natural log space, so we divide by ln(10) before using normalizeFromLog10
                     final double conditionalProbability =
                             MathUtils.normalizeFromLog10(new double[]{
-                                    outlierUnnormalizedLogProbability,
-                                    notOutlierUnnormalizedLogProbability})[0];
+                                    outlierUnnormalizedLogProbability / ln10,
```
davidbenjamin: Use `MathUtils::logToLog10`

samuelklee: ah, right!

davidbenjamin:
Just one thing and then you can merge.

samuelklee:
Not sure why coveralls is hanging.  Passed before rebase, so going to go ahead and merge.

PR #577, opened 2016-06-21
Changed representation of target/copy-ratio segments in GATK CNV output.

### LINKED ISSUE #485: Change representation of target/copy-ratio segments in GATK CNV output.
For legacy reasons, the copy-ratio segments reported by CNV are specified as "end of first target in segment--end of last target in segment," rather than the more natural "start of first target in segment--end of last target in segment."  One reason for this is because the R script responsible for segmentation acts on a list of genomic locations, rather than intervals, so we pass only the target ends during the target-segmentation step.

However, it's trivial to look up the target starts in the original list of targets and restore this information before writing the segment file.  In fact, this is already done by ACNV when taking in the segment files output by CNV---the reasoning here being that hets falling in the first target of each copy-ratio segment would otherwise not be grouped with the correct segment.

If no legacy pipelines/scripts will be affected, it makes more sense to perform this step in GATK CNV---specifically, in the PerformSegmentation tool.

samuelklee:
Also made tests for segmenter check equality of intervals, not just segment mean.

Closes #485.


samuelklee:
@LeeTL1220, could you review?

src/main/resources/org/broadinstitute/hellbender/utils/segmenter/CBS.R
```diff
@@ -95,6 +95,14 @@ segment_data = function(sample_name, tn_file, output_file, log_input, weights_fi
 	# Undo conversion of sample_name to R format (retain dashes, spaces, etc.)
 	segmented[,"Sample"] = sample_name
 
+	# Change segment start to start of first target in segment (rather than end of first target in segment)
```
LeeTL1220: @achevali What is your opinion?

LeeTL1220:
@samuelklee I reviewed and I appreciate the code cleanup.  Looks good to me.

PR #578, opened 2016-06-21
adding a gvcf to large files

lbergelson:


akiezun:
@lbergelson what needs to be done here?

lbergelson:
Oh, I made this to test an issue I was having with git lfs.  It was never meant to be merged.  Should have closed this before.

PR #597, opened 2016-07-01
Ported from fix for GATK3 issue #1357

SHuang-Broad:
@davidbenjamin, would you please review? Should be quick as this is almost the same code, except lines 252-256 in `HaplotypeCallerGenotypingEngine.java`, where the GATK4 is providing locations in a way not the same as in GATK3.

Other parts are in GATK, and has been dealt with (and in the new snapshot).

Thanks!

__EDIT__
Will remove the commented out code once reviewer says OK.


davidbenjamin:
@SHuang-Broad This is fine, and you can delete the commented out code in lines 252-256.

droazen:
Can you write a unit test for `HaplotypeCallerGenotypingEngine` that fails before the fix and passes after it?

build.gradle
```diff
@@ -123,7 +123,7 @@ check.dependsOn installDist
 
 dependencies {
     compile files("${System.properties['java.home']}/../lib/tools.jar")
-    compile 'org.broadinstitute:gatk:4.alpha.1-151-g76f78d2-20160624.165758-3'
+    compile 'org.broadinstitute:gatk:4.alpha.1-174-gbbd028b-20160701.172759-1'
```
droazen: This snapshot appears to be missing the native PairHMM libraries -- please don't merge until this is fixed.

droazen: Whoops, I take it back -- the snapshot is fine :)

SHuang-Broad:
@droazen. Let me think about it. 
The issue was that in face of high ploidy and high alt alleles, the number of genotypes may be higher than `Integer.MAX_VALUE`, giving exceptions. 
I can think about how to write a unit test for the newly added method `reduceNumberOfAlternativeAllelesBasedOnHaplotypesScores` to see if the "bad" alleles are indeed removed as desired. 
Is this what you are asking for?

droazen:
@SHuang-Broad Yes, just a targeted unit test of some kind that proves the fix works

SHuang-Broad:
Roger, @droazen.

SHuang-Broad:
@droazen , would you please take a look and see if this is desired (just the last two commits)?
Will back port to GATK3.
The unit test caught an logical error. Thanks!

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -40,13 +42,19 @@
 
     private final PloidyModel ploidyModel;
 
+    private final int MAX_GENOTYPE_COUNT_TO_ENUMERATE = configuration.genotypeArgs.MAX_GENOTYPE_COUNT;
```
lbergelson: This is a bit strange since it's relying on the constructor ordering in a bit of a strange way.  I think you'd be better off initializing this in the constructor where it's more obviously clear that it won't result in a null ptr dereference.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -40,13 +42,19 @@
...
+    private static final Map<Integer, Integer> practicaAlleleCountForPloidy = new HashMap<>();
```
lbergelson: I assume this is a a typo?  practica->practical

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -211,7 +221,32 @@ public CalledHaplotypes assignGenotypeLikelihoods(final List<Haplotype> haplotyp
...
+                                ploidy, originalAlleleCount, practicalAlleleCount));
+                        logger.warn(String.format("Alleles kept are:%s", practicalAlleleMapper.keySet()));
+                    }
+                }else{
+                    practicalAlleleMapper = alleleMapper;
+                }
+
+                ReadLikelihoods<Allele> readAlleleLikelihoods = readLikelihoods.marginalize(practicalAlleleMapper,
+                                                                                            new SimpleInterval(mergedVC).expandWithinContig(ALLELE_EXTENSION, header.getSequenceDictionary())
+                                                                                            );
```
lbergelson: If you want to split this up I would still keep the `);` on the second line I think.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -211,7 +221,32 @@ public CalledHaplotypes assignGenotypeLikelihoods(final List<Haplotype> haplotyp
                     logger.info("Genotyping event at " + loc + " with alleles = " + mergedVC.getAlleles());
                 }
 
-                ReadLikelihoods<Allele> readAlleleLikelihoods = readLikelihoods.marginalize(alleleMapper, new SimpleInterval(mergedVC).expandWithinContig(ALLELE_EXTENSION, header.getSequenceDictionary()));
+                // if the number of allele is so high that enumerating all possible genotypes is impractical,
+                // as determined by MAX_GENOTYPE_COUNT_TO_ENUMERATE,
+                // trim alleles that are not well supported by good-scored haplotypes,
+                // otherwise alias, i.e. no trimming if genotype count is small
+                final int originalAlleleCount = alleleMapper.size();
```
lbergelson: Could you extract this block of code as a function?

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
lbergelson: Also, this shouldn't be in all caps, since it isn't really a constant, it depends on the value of MAX_GENOTYPE_COUNT which also isn't really a constant since it's set on the commandline.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -211,7 +221,32 @@ public CalledHaplotypes assignGenotypeLikelihoods(final List<Haplotype> haplotyp
...
+                Integer practicalAlleleCount = practicaAlleleCountForPloidy.get(ploidy);
+                if(practicalAlleleCount==null) { // maximum allele count given this ploidy and MAX_GENOTYPE_COUNT_TO_ENUMERATE hasn't been computed
```
lbergelson: This might be clearer as 

```java
practicalAlleleCountForPloidy.putIfAbsent(GenotypeLikelihoodCalculators.computeMaxAcceptableAlleleCount(ploidy, MAX_GENOTYPE_COUNT_TO_ENUMERATE);
final int practicalAlleleCount = practicalAlleleCount.get(ploidy);
```

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -211,7 +221,32 @@ public CalledHaplotypes assignGenotypeLikelihoods(final List<Haplotype> haplotyp
...
+                    practicalAlleleCount = GenotypeLikelihoodCalculators.computeMaxAcceptableAlleleCount(ploidy, MAX_GENOTYPE_COUNT_TO_ENUMERATE);
+                    practicaAlleleCountForPloidy.put(ploidy, practicalAlleleCount);
+                }
+
+                Map<Allele, List<Haplotype>> practicalAlleleMapper = null;
+                if (practicalAlleleCount < originalAlleleCount) {
+                    practicalAlleleMapper = reduceNumberOfAlternativeAllelesBasedOnHaplotypesScores(alleleMapper, practicalAlleleCount);
+                    if( configuration.DEBUG && logger != null ) {
```
lbergelson: This is weird.  Is it possible to have a null logger here?  If this is an important  warning it seems like it should always be output, if it's not important than why not just set it to be an info  or debug level.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -211,7 +221,32 @@ public CalledHaplotypes assignGenotypeLikelihoods(final List<Haplotype> haplotyp
                     logger.info("Genotyping event at " + loc + " with alleles = " + mergedVC.getAlleles());
                 }
 
-                ReadLikelihoods<Allele> readAlleleLikelihoods = readLikelihoods.marginalize(alleleMapper, new SimpleInterval(mergedVC).expandWithinContig(ALLELE_EXTENSION, header.getSequenceDictionary()));
+                // if the number of allele is so high that enumerating all possible genotypes is impractical,
```
lbergelson: typo: allele -> alleles

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -211,7 +221,32 @@ public CalledHaplotypes assignGenotypeLikelihoods(final List<Haplotype> haplotyp
...
+
+                Map<Allele, List<Haplotype>> practicalAlleleMapper = null;
+                if (practicalAlleleCount < originalAlleleCount) {
+                    practicalAlleleMapper = reduceNumberOfAlternativeAllelesBasedOnHaplotypesScores(alleleMapper, practicalAlleleCount);
+                    if( configuration.DEBUG && logger != null ) {
+                        logger.warn(String.format("Removed alt alleles where ploidy is %d and original allele count is %d, whereas after trimming the allele count becomes %d",
+                                ploidy, originalAlleleCount, practicalAlleleCount));
+                        logger.warn(String.format("Alleles kept are:%s", practicalAlleleMapper.keySet()));
+                    }
+                }else{
```
lbergelson: spaces around the else generally

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -256,6 +295,79 @@ public CalledHaplotypes assignGenotypeLikelihoods(final List<Haplotype> haplotyp
...
+        // populate allelePriorityQ with the relevant information
+        for(final Allele allele : alleleMapper.keySet()){
+
+            if(allele.isReference()){ // collect scores information only on alt alleles; ref allele is never trimmed by this function
+                allelesToRetain.add(allele);
+                continue;
+            }
+
+            final List<Double> hapScores = alleleMapper.get(allele).stream().map(hap -> hap.getScore()).collect(Collectors.toList());
+            Collections.sort(hapScores);
```
lbergelson: do this as part of the stream. i.e `map(...).sorted().collect(...)`

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -256,6 +295,79 @@ public CalledHaplotypes assignGenotypeLikelihoods(final List<Haplotype> haplotyp
...
+    /**
+     * A utility class that provides ordering information, given best and second best haplotype scores.
+     * If there's a tie between the two alleles when comparing their best haplotype score, the second best haplotype score
+     * is used for breaking the tie. In the case that one allele doesn't have a second best allele, i.e. it has only one
+     * supportive haplotype, its second best score is set as null, and is always considered "worse" than another allele
+     * that has the same best haplotype score, but also has a second best haplotype score.
+     * TODO: in the extremely unlikely case that two alleles, having the same best haplotype score, neither have a second
+     * best haplotype score, the case is undecided.
+     */
+    private static final class AlleleScoredByHaplotypeScores {
```
lbergelson: Is there are reason that you only compare the first two scores?  Why not compare the entire chain until you find one?  

Is it possible to use some other attribute as the final tiebreaker so there is no indeterminate ordering?

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -256,6 +295,79 @@ public CalledHaplotypes assignGenotypeLikelihoods(final List<Haplotype> haplotyp
...
+
+        return alleleMapper.entrySet().stream().filter(p -> allelesToRetain.contains(p.getKey()))
+                .collect(Collectors.toMap(p->p.getKey(), p->p.getValue()));
+    }
+
+    /**
+     * A utility class that provides ordering information, given best and second best haplotype scores.
+     * If there's a tie between the two alleles when comparing their best haplotype score, the second best haplotype score
+     * is used for breaking the tie. In the case that one allele doesn't have a second best allele, i.e. it has only one
+     * supportive haplotype, its second best score is set as null, and is always considered "worse" than another allele
```
lbergelson: It looks like these comments about null are outdated since it will actually be negative infinity.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -256,6 +295,79 @@ public CalledHaplotypes assignGenotypeLikelihoods(final List<Haplotype> haplotyp
...
+     * all possible genotypes is impractical.
+     * Use the highest haplotype score of each allele to decide which ones to trim away, and in case of tie, use the second best.
+     * @param alleleMapper          original allele to haplotype map to be trimmed
+     * @param desiredNumOfAlleles   desired allele count (ref allele won't be removed)
+     * @return trimmed map of the desired size
+     */
+    @VisibleForTesting
+    static Map<Allele, List<Haplotype>> reduceNumberOfAlternativeAllelesBasedOnHaplotypesScores(final Map<Allele, List<Haplotype>> alleleMapper, final int desiredNumOfAlleles) {
+
+        final PriorityQueue<AlleleScoredByHaplotypeScores> altAlleleMaxPriorityQ = new PriorityQueue<>((sa1, sa2) -> sa2.compareTo(sa1));
```
lbergelson: I don't think you need to specify the comparator if it's just the natural ordering of the elements.

lbergelson: Oh, I see, you didn't make AlleleScoredByHaplotypeScores `Comparable`.  Either make it Comparable or just use the comparator directly here.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -256,6 +295,79 @@ public CalledHaplotypes assignGenotypeLikelihoods(final List<Haplotype> haplotyp
...
     /**
+     * Trim excessive alt alleles when the combination of ploidy and alleles is so large that exhaustively enumerating
+     * all possible genotypes is impractical.
+     * Use the highest haplotype score of each allele to decide which ones to trim away, and in case of tie, use the second best.
+     * @param alleleMapper          original allele to haplotype map to be trimmed
+     * @param desiredNumOfAlleles   desired allele count (ref allele won't be removed)
+     * @return trimmed map of the desired size
+     */
+    @VisibleForTesting
+    static Map<Allele, List<Haplotype>> reduceNumberOfAlternativeAllelesBasedOnHaplotypesScores(final Map<Allele, List<Haplotype>> alleleMapper, final int desiredNumOfAlleles) {
```
lbergelson: Could you add a check at the start to test that the actual number of alleles is >= the desired number of alleles and return if that's the case

lbergelson:
This needs a test that covers the rest of the new code in `HaplotypeCallerGenotypingEngine` as well.  An integration test that runs on a single line vcf with tons of alleles would be a good example (i.e. one that didn't run in reasonable time before, but now will run just fine thanks to your fix.

lbergelson:
@SHuang-Broad Back to you with some minor refactoring requests, a question, and request for additional test.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
SHuang-Broad: done, though not sure if the change is what you suggested. will talk in person.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
SHuang-Broad: done.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
SHuang-Broad: I'm not sure if `Allele`s have a quantitative feature that would sensibly break the tie. 
What we need is some quantitative feature from the haplotypes that 
1) comes from the actual data at hand, 
2) can be used to judge, heuristically, the alleles,
3) no two alleles would be equal judged by that property
There could be another way to remove alt alleles even before the map from the alleles to the haplotypes is created, but that might deserve another separate fix.
Needs to think more about that.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
SHuang-Broad: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -256,6 +274,104 @@ public CalledHaplotypes assignGenotypeLikelihoods(final List<Haplotype> haplotyp
...
+        return alleleMapper.entrySet().stream().filter(p -> allelesToRetain.contains(p.getKey()))
+                .collect(Collectors.toMap(p->p.getKey(), p->p.getValue()));
+    }
+
+    /**
+     * A utility class that provides ordering information, given best and second best haplotype scores.
+     * If there's a tie between the two alleles when comparing their best haplotype score, the second best haplotype score
+     * is used for breaking the tie. In the case that one allele doesn't have a second best allele, i.e. it has only one
+     * supportive haplotype, its second best score is set as {@link Double#NEGATIVE_INFINITY}.
+     * TODO: in the extremely unlikely case that two alleles, having the same best haplotype score, neither have a second
```
lbergelson: @SHuang-Broad Would it be possible to use the bases in the alleles themselves to sort in the worst case tie?  Two different alleles are going to have different bases by definition, and if they're equally scored it wouldn't matter which is sorted first.

SHuang-Broad: Yes, you are right. In this case I could rely on the `Allele.compareTo()` then, as the quantitative part probably doesn't matter any more here.

SHuang-Broad:
Tests are failing because of partial rebase on Adam's fix for HDF5. Hopefully they will pass when Adam's fix is merged in and this branch rebased on top of that.

davidbenjamin:
@SHuang-Broad I approve.

PR #601, opened 2016-07-05
Changed the default value for Het calling stringency from 5 to 30

asmirnov239:
Please review


src/main/java/org/broadinstitute/hellbender/tools/exome/GetBayesianHetCoverage.java
```diff
@@ -162,7 +162,7 @@
             shortName = HET_CALLING_STRINGENCY_SHORT_NAME,
             optional = true
     )
-    protected double hetCallingStringency = 5;
+    protected double hetCallingStringency = 30;
```
mbabadi: I agree with this -- this is something that I wanted to change for a while, thank you!

mbabadi:
I agree with your changed @asmirnov239, feel free to merge.

PR #608, opened 2016-07-08
Ak calculate genotype posteriors with originals

akiezun:
@ldgauthier new version of this pull request, now with the original code in the first commit


src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriors.java
```diff
@@ -134,89 +80,89 @@
...
- *   -o output.withPosteriors.vcf
+ *   -O output.withPosteriors.vcf
  * </pre>
  *
  */
-@DocumentedGATKFeature( groupName = HelpConstants.DOCS_CAT_VARMANIP, extraDocs = {CommandLineGATK.class} )
-public class CalculateGenotypePosteriors extends RodWalker<Integer,Integer> {
+@CommandLineProgramProperties(
+        summary = "Given a VCF with genotype likelihoods from the HaplotypeCaller, UnifiedGenotyper, or another source which provides" +
+                  "unbiased genotype likelihoods, calculate the posterior genotype state and likelihood given allele frequency" +
```
ldgauthier: As David B. pointed out, I played fast and loose with the terminology in the original code, but this should be "calculate the posterior genotype state and probability" -- posteriors are no longer likelihoods.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriors.java
```diff
@@ -134,89 +80,89 @@
...
+                  "information from both the samples themselves and input VCFs describing allele frequencies in related populations.",
```
ldgauthier: The input samples are not necessarily used, for instance if there aren't enough I believe.  Input VCFs are also optional.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriors.java
```diff
@@ -134,89 +80,89 @@
...
+        oneLineSummary = "Calculate genotype posterior likelihoods given panel data",
```
ldgauthier: Again, should be posterior probabilities.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriors.java
```diff
@@ -242,55 +188,62 @@
...
     /**
      * Skip application of family-based priors.  Note: if pedigree file is absent, family-based priors will be skipped.
      */
-    @Argument(fullName="skipFamilyPriors",shortName="skipFam",doc="Skip application of family-based priors", required=false)
+    @Argument(fullName="skipFamilyPriors",shortName="skipFam",doc="Skip application of family-based priors", optional=true)
     public boolean skipFamilyPriors = false;
 
-    @Output(doc="File to which variants should be written")
-    protected VariantContextWriter vcfWriter = null;
+    @Argument(fullName="pedigree", shortName="ped", doc="Pedigree file", optional=true)
```
ldgauthier: Docs?

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/PosteriorLikelihoodsUtils.java
```diff
@@ -1,68 +1,23 @@
...
 
 import java.util.*;
 
-public class PosteriorLikelihoodsUtils {
+public final class PosteriorLikelihoodsUtils {
+
+    private PosteriorLikelihoodsUtils(){}
 
+    /**
+     * Calculates phred-scaled posterior probabilities for genotypes given the data and allele frequency priors.
```
ldgauthier: data here is genotype likelihoods, right?

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/PosteriorLikelihoodsUtils.java
```diff
@@ -1,68 +1,23 @@
...
+     */
     public static VariantContext calculatePosteriorGLs(final VariantContext vc1,
```
ldgauthier: I think it would make @davidbenjamin happy if we renamed this.  Maybe calculatePosteriorProbs?  I think maybe in the original version of the code the PLs (GLs) were modified, but now the likelihoods are preserved and PPs (Phred-scaled posteriors) are output, so this is both mathematically wrong and wrong in terms of the code.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/PosteriorLikelihoodsUtils.java
```diff
@@ -214,7 +174,8 @@ public static VariantContext calculatePosteriorGLs(final VariantContext vc1,
     }
 
     // convenience function for a single genotypelikelihoods array. Just wraps.
-    protected static double[] calculatePosteriorGLs(final double[] genotypeLikelihoods,
+    @VisibleForTesting
+    static double[] calculatePosteriorGLs(final double[] genotypeLikelihoods,
```
ldgauthier: Also rename here

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/PosteriorLikelihoodsUtils.java
```diff
@@ -330,29 +294,31 @@ else if ( alleleCountsFromVCF instanceof String || alleleCountsFromVCF instanceo
      * @param integerListContainingVCField - Object returned by a call to VariantContext::getAttribute()
      * @return - array of ints
      */
+    @SuppressWarnings("unchecked")
     public static int[] extractInts(final Object integerListContainingVCField) {
```
ldgauthier: This is sort of gross, but I blame HTSJDK and its refusal to use the type information it already has from the annotation headers.  Maybe it's worth documenting that if we're working with a integerListContainingVCField that's read directly out of the file it will be a String but if it gets pulled from a VariantContext object built elsewhere in the code it will be an Integer or a List, depending on how many alleles there are at the site (if I recall correctly).

src/test/java/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriorsIntegrationTest.java
```diff
@@ -1,139 +1,90 @@
...
-import org.broadinstitute.gatk.engine.walkers.WalkerTest;
+import org.broadinstitute.hellbender.CommandLineProgramTest;
+import org.broadinstitute.hellbender.utils.test.IntegrationTestSpec;
 import org.testng.annotations.Test;
 
-import java.util.Arrays;
+import java.io.IOException;
+import java.util.Collections;
+
+public final class CalculateGenotypePosteriorsIntegrationTest extends CommandLineProgramTest {
```
ldgauthier: @akiezun How do I access these test results?  I want to double check that the code is still doing the right thing in all these cases.

src/test/java/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriorsIntegrationTest.java
```diff
@@ -1,139 +1,90 @@
...
 
-    @Test(enabled = true)
+    @Test
     //use the first 20 variants to save time; they have a nice range of AC from 4 to over 4000
-    public void testUsingDiscoveredAF() {
-        WalkerTestSpec spec = new WalkerTestSpec(
-                "-T CalculateGenotypePosteriors --no_cmdline_in_header" +
-                        " -o %s" +
-                        " -R " + b37KGReference +
+    public void testUsingDiscoveredAF() throws IOException {
```
ldgauthier: PPs should be != to PLs because they use input samples

src/test/java/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriorsIntegrationTest.java
```diff
@@ -1,139 +1,90 @@
...
     }
 
-    @Test(enabled = true)
-    public void testInputINDELs() {
-        WalkerTestSpec spec = new WalkerTestSpec(
-                "-T CalculateGenotypePosteriors --no_cmdline_in_header -useACoff" +
-                        " -o %s" +
-                        " -R " + b37KGReference +
+    @Test
+    public void testInputINDELs() throws IOException {
```
ldgauthier: INDELs should be skipped (I can't remember if that means PPs = PLs or PPs are missing)

src/test/java/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriorsIntegrationTest.java
```diff
@@ -1,139 +1,90 @@
...
     }
 
-    @Test(enabled = true)
-    public void testSingleParentFamily() {
-        WalkerTestSpec spec = new WalkerTestSpec(
-                "-T CalculateGenotypePosteriors --no_cmdline_in_header " +
-                        " -o %s" +
-                        " -R " + b37KGReference +
+    @Test
+    public void testSingleParentFamily() throws IOException {
```
ldgauthier: I can't remember what's supposed to happen here.  I think I put in the code for posteriors if only one parent is called, but obviously it's not informative for most genotype combinations.

src/test/java/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriorsIntegrationTest.java
akiezun: they are in the PR

ldgauthier:
Code looks fine -- some doc cleanup would be good.

I want to review the new tests on chr20 -- are the results stored somewhere like we do in GATK3?

src/test/java/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriorsIntegrationTest.java
ldgauthier: Ohhhhh, so you're comparing to a file in the repo instead of md5 shenanigans?

akiezun:
test data and results are included in this PR. In GATK4 we include all test data in the repo so that anyone can test the gatk

akiezun:
assigning back to @ldgauthier - reassign back when you've reviewed the test results

src/test/java/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriorsIntegrationTest.java
akiezun: yes

src/test/resources/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriors/expectedCGP_testFamilyPriors.vcf
```diff
@@ -0,0 +1,42 @@
...
+##INFO=<ID=QD,Number=1,Type=Float,Description="Variant Confidence/Quality by Depth">
+##INFO=<ID=ReadPosRankSum,Number=1,Type=Float,Description="Z-score from Wilcoxon rank sum test of Alt vs. Ref read position bias">
+##INFO=<ID=set,Number=1,Type=String,Description="Source VCF for the merged record in CombineVariants">
+##contig=<ID=20,length=63025520,assembly=b37>
+##contig=<ID=21,length=48129895,assembly=b37>
+##reference=human_b37_both.fasta
+##source=CalculateGenotypePosteriors
+##source=SelectVariants
+#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	NA12878	NA12891	NA12892
+20	564477	.	A	G	0	.	AC=0;AF=0.00;AN=6;DP=32;FS=0.000;MLEAC=0;MLEAF=0.00;MQ0=0;PG=0,0,0;set=Intersection	GT:AD:DP:GQ:JL:JP:PL:PP	0/0:19,0:19:66:9:9:0,57,632:0,66,672	0/0:3,0:3:9:9:9:0,9,109:0,9,164	0/0:10,0:10:30:9:9:0,30,356:0,30,411
```
ldgauthier: Since you moved these to chromosome 20, are these actually the correct reference bases for these positions?  I'm skeptical that this VCF is valid.

src/test/java/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriorsIntegrationTest.java
ldgauthier: Looks like PPs = PLs, as intended.

src/test/resources/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriors/expectedCGP_testSingleParentFamily.vcf
```diff
@@ -0,0 +1,82 @@
...
+##INFO=<ID=VariantType,Number=1,Type=String,Description="Variant type description">
+##INFO=<ID=culprit,Number=1,Type=String,Description="The annotation which was the worst performing in the Gaussian mixture model, likely the reason why the variant was filtered out">
+##INFO=<ID=set,Number=1,Type=String,Description="Source VCF for the merged record in CombineVariants">
+##contig=<ID=20,length=63025520,assembly=b37>
+##contig=<ID=21,length=48129895,assembly=b37>
+##reference=human_g1k_v37.fasta
+##source=CalculateGenotypePosteriors
+##source=SelectVariants
+#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	PAIR_CHILD	PAIR_FATHER	TRIO_CHILD	TRIO_FATHER	TRIO_MOTHER	UNRELATED_INDIVIDUAL_CONTROL
+20	10109	.	A	T	99	PASS	AC=0;AF=0.00;AN=12;TestDescription=HomRef/HomRef/HomRef	GT:GQ:JL:JP:PL:PP	0/0:50:.:.:0,50,200	0/0:50:.:.:0,50,200	0/0:97:45:46:0,50,200:0,97,299	0/0:50:45:46:0,50,200:0,50,250	0/0:50:45:46:0,50,200:0,50,250	0/0:50:.:.:0,50,200
```
ldgauthier: Ditto here with chr change

src/test/resources/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriors/expectedCGP_testFamilyPriors.vcf
akiezun: does it matter? the reference is not used AFAIK

src/test/java/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriorsIntegrationTest.java
ldgauthier: Nope, looks like I never got around to doing parent-offspring pairs.  PPs are omitted.

ldgauthier:
Test results look good because they're identical.  I'm a little concerned that you probably have invalid VCFs though.

src/test/resources/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriors/expectedCGP_testFamilyPriors.vcf
ldgauthier: If the reference isn't used then why did you have to move it from chr1?

akiezun: there's a limitation in gatk4 when a reference is needed for `-L` to work. I think that was why i moved them

ldgauthier: But it shouldn't be that much slower without the -L, right?  At least it wasn't in GATK3, it just gave bad time estimates.

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriors.java
akiezun: fixed

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriors.java
akiezun: what wording would you prefer to see here?

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriors.java
akiezun: fixed. I'm assuming you want to rename `PosteriorLikelihoodsUtils` and `PosteriorLikelihoodsUtilsUnitTest` accordingly, right?

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriors.java
akiezun: nice catch. added doc from gatk3

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/PosteriorLikelihoodsUtils.java
akiezun: i think the data is the variant contexts

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/PosteriorLikelihoodsUtils.java
akiezun: renamed to `calculatePosteriorProbs`

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/PosteriorLikelihoodsUtils.java
akiezun: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/variantutils/PosteriorLikelihoodsUtils.java
akiezun: added that comment verbatim

src/test/java/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriorsIntegrationTest.java
akiezun: so the output is incorrect? I see PPs different from PLs in `expectedCGP_testUsingDiscoveredAF.vcf`, eg:
`GT:AD:DP:GQ:PL:PP       0/0:4,0:4:7:0,12,187:0,7,183`

src/test/java/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriorsIntegrationTest.java
akiezun: ok, so this is correct, right?

src/test/java/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriorsIntegrationTest.java
akiezun: so output is correct, right?

src/test/resources/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriors/expectedCGP_testFamilyPriors.vcf
akiezun: you're right. i was able to move back to chr1 without a problem.

akiezun: thanks for noticing

src/test/resources/org/broadinstitute/hellbender/tools/walkers/variantutils/CalculateGenotypePosteriors/expectedCGP_testSingleParentFamily.vcf
akiezun: fixed - back to chr1

akiezun:
merging. comments and concerns addressed

PR #609, opened 2016-07-11
remove svd code from protected because it's in public now

### LINKED ISSUE #550: remove SVD code from protected - available in public now
svd is now available in public. That code should be deleted from protected

akiezun:
fixes #550 

@davidbenjamin can you have a look?


davidbenjamin:
Sure, will do.

src/main/java/org/broadinstitute/hellbender/tools/exome/DecomposeSingularValues.java
```diff
@@ -84,7 +84,7 @@ protected void runPipeline(final JavaSparkContext ctx) {
     private void writeMatrix(final RealMatrix v, final File outputFilename) throws IOException {
         final List<String []> textTable = new ArrayList<>();
         for (int i = 0; i < v.getRowDimension(); i ++){
-            textTable.add(Arrays.stream(v.getRow(i)).mapToObj(Double :: toString).toArray(String[]::new));
+            textTable.add(Arrays.stream(v.getRow(i)).mapToObj(Double::toString).toArray(String[]::new));
```
davidbenjamin: While you're at it, change `v` to a more matrixy name.

akiezun: like Trinity or Morpheus?

src/test/java/org/broadinstitute/hellbender/utils/svd/SVDUnitTest.java
```diff
@@ -1,21 +1,50 @@
...
+                {new SparkSingularValueDecomposer(SparkContextFactory.getTestSparkContext())}
+        };
+    }
+
+    @Test(dataProvider = "makeSVDers")
+    public void testBasicFunction(final SingularValueDecomposer decomposer){
+        LoggingUtils.setLoggingLevel(Log.LogLevel.INFO);
+        try {
+            final ReadCountCollection rcc = ReadCountCollectionUtils.parse(CONTROL_PCOV_FULL_FILE);
+            final SVD svd = decomposer.createSVD(rcc.counts());
```
davidbenjamin: SVD testing does not need to be coupled with `ReadCountCollection` and pcov files.  Rather than referring to files, I would rely on the following necessary and sufficient assertions.  I will use the notation that M is the original matrix and M = U S V* is the SVD, where S is a matrix with the singular values on its diagonal and zeroes elsewhere

* U is unitary
* V is unitary
* M = U S V*
* pseudoinverse = V S' U* where S' is the transpose of the element-by-element inverse of the non-zero elements of S.

This should work for an arbitrary random matrix and thus you don't need any test files.

src/main/java/org/broadinstitute/hellbender/tools/exome/DecomposeSingularValues.java
davidbenjamin: I was thinking `m` or `mat` but feel free to use your imagination.

davidbenjamin:
@akiezun Not much to say, but I did have an opinion about the unit test.

src/test/java/org/broadinstitute/hellbender/utils/svd/SVDUnitTest.java
akiezun: such unit tests exist in public. in protected, the tests check that it all works with the readcounts etc

davidbenjamin: Hmmm if that's the case I'm not sure the test here is needed at all.  I mean, it's not really testing an interaction with `ReadCountCollection` since `rcc` only enters via its `counts()`.  So it seems to me like this is in effect just testing the consistency of the four test files and the fact that parsing them into a `ReadCountCollection` works.

akiezun: do you want me to keep them or delete them?

davidbenjamin: If you find my reasoning convincing, delete.

src/main/java/org/broadinstitute/hellbender/tools/exome/DecomposeSingularValues.java
akiezun: done

src/test/java/org/broadinstitute/hellbender/utils/svd/SVDUnitTest.java
akiezun: ok, will delete

src/test/java/org/broadinstitute/hellbender/tools/exome/DecomposeSingularValuesIntegrationTest.java
```diff
@@ -86,15 +86,30 @@ private void assertSVDValues(final File outputFileV, final File outputFileS, fin
...
+            assertUnitaryMatrix(svd.getU());
             Assert.assertTrue(MatrixUtils.isSymmetric(sDiag, 1e-32));
 
         } catch (final IOException ioe) {
             Assert.fail("Could not open test file: " + CONTROL_PCOV_FULL_FILE, ioe);
         }
     }
 
+    /**
+     * Assrt that the given matrix is unitary.
```
davidbenjamin: `Assert`

src/test/java/org/broadinstitute/hellbender/tools/exome/DecomposeSingularValuesIntegrationTest.java
```diff
@@ -86,15 +86,30 @@ private void assertSVDValues(final File outputFileV, final File outputFileS, fin
...
+     * @param m
+     */
+    public static void assertUnitaryMatrix(final RealMatrix m){
+        final RealMatrix mInv = new QRDecomposition(m).getSolver().getInverse();
```
davidbenjamin: Same comments as in gatk-public PR.  Check that row dimension == column dimension, then check `m.multiply(m.transpose()) == m.transpose().multiply(m) == identity`.

davidbenjamin:
Two more little comments, after which you can merge.

src/test/java/org/broadinstitute/hellbender/tools/exome/DecomposeSingularValuesIntegrationTest.java
akiezun: same issue as in https://github.com/broadinstitute/gatk/pull/1998/files - U may not be square

davidbenjamin: Same answer as there -- Apache Commons has a weird SVD, but nothing we can do about it.

src/test/java/org/broadinstitute/hellbender/tools/exome/DecomposeSingularValuesIntegrationTest.java
akiezun: fixed

Issue #598, opened 2016-07-01
delete files that were moved to gatk


lbergelson:
We currently have files that were moved from gatk-protected to gatk but never deleted in gatk-protected.   See https://github.com/broadinstitute/gatk/pull/1888.  Since these moved without changing packages we have duplicate files with the same exact identifiers.  We should figure out how to prevent this in the future, possibly by giving gatk-protected its own package.  

We should check the affected files before deleting because I think there may have diverged from the versions in public.

droazen:
Issue moved to [broadinstitute/gatk #2869](https://github.com/broadinstitute/gatk/issues/2869) via [**ZenHub**](https://www.zenhub.com/)

PR #610, opened 2016-07-11
delete tsv package from protected

akiezun:
fix for https://github.com/broadinstitute/gatk-protected/issues/598

@lbergelson can you review?


akiezun:
@samuelklee can you review? It's almost entirely a straightforward delete of code that was moved.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCaller.java
```diff
@@ -180,6 +180,11 @@
     protected int defaultAssemblyRegionPadding() { return 100; }
 
     @Override
+    protected int defaultMaxReadsPerAlignmentStart() {
+        return 50;
```
samuelklee: just nitpicking, but the vertical whitespacing here is different from the other nearby methods

samuelklee:
one very minor comment, otherwise looks good to me!

PR #616, opened 2016-07-18
HaplotypeCallerSpark

lbergelson:
Initial Implementation of HaplotypeCallerSpark.  Passes concordance tests but not exact tests.  Annotations may be wonky and performance is likely not very good due to a whole bam shuffle.

FeatureSide inputs aren't implemented yet so -dbsnp and -comp don't work. 

Corresponding public pr is https://github.com/broadinstitute/gatk/pull/2016

SerializationTestUtils.java moved to public.


src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
```diff
@@ -0,0 +1,329 @@
...
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+@CommandLineProgramProperties(summary = "HaplotypeCaller on Spark", oneLineSummary = "HaplotypeCaller on Spark", programGroup = SparkProgramGroup.class)
+public final class HaplotypeCallerSpark extends GATKSparkTool {
```
akiezun: add class comment

src/main/java/org/broadinstitute/hellbender/tools/spark/SparkReadShard.java
```diff
@@ -0,0 +1,52 @@
...
+
+
+import org.broadinstitute.hellbender.engine.ReadShard;
+import org.broadinstitute.hellbender.utils.SimpleInterval;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+
+import java.io.Serializable;
+import java.util.Iterator;
+
+public class SparkReadShard implements ReadShard, Serializable {
```
akiezun: add class javadoc
make final

src/main/java/org/broadinstitute/hellbender/tools/spark/SparkReadShard.java
```diff
@@ -0,0 +1,52 @@
...
+    private static final long serialVersionUID = 1L;
+
+    private final ShardBoundary key;
+    private final Iterable<GATKRead> reads;
+
+    public SparkReadShard(ShardBoundary key, Iterable<GATKRead> reads){
+        this.key = key;
```
akiezun: null checks

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCaller.java
```diff
@@ -200,7 +216,15 @@ public AssemblyRegionEvaluator assemblyRegionEvaluator() {
...
     @Override
     public void onTraversalStart() {
-        hcEngine = new HaplotypeCallerEngine(hcArgs, getHeaderForReads(), referenceArguments.getReferenceFileName());
+        final CachingIndexedFastaSequenceFile referenceReader;
+        final File reference = new File(referenceArguments.getReferenceFileName());
+        try {
+            referenceReader = new CachingIndexedFastaSequenceFile(reference);
+        } catch (FileNotFoundException e) {
+            throw new UserException.CouldNotReadInputFile(reference, e);
+        }
```
akiezun: extract this code to a method returning `CachingIndexedFastaSequenceFile`

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -30,6 +32,8 @@
  */
 public final class HaplotypeCallerGenotypingEngine extends GenotypingEngine<AssemblyBasedCallerArgumentCollection> {
 
+    private static final Logger logger = LogManager.getLogger(GenotypingEngine.class);
```
akiezun: use `HaplotypeCallerGenotypingEngine.class` here

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/ReadThreadingAssemblerArgumentCollection.java
```diff
@@ -23,7 +24,11 @@
      */
     @Advanced
     @Argument(fullName="kmerSize", shortName="kmerSize", doc="Kmer size to use in the read threading assembler", optional = true)
-    public List<Integer> kmerSizes = Arrays.asList(10, 25);
+    public List<Integer> kmerSizes = new ArrayList<>();
```
akiezun: are these changes needed because of serialization? If yes, does this mean you're pulling in the argument to serialization via a closure (closures are always serialized by Java serialization and not Kryo)? Is yes, maybe that's a problem - we should not avoid serializing big closures

src/test/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSparkIntegrationTest.java
```diff
@@ -0,0 +1,191 @@
...
+        runCommandLine(args);
+
+        final double concordance = calculateConcordance(output, gatk3Output);
+        Assert.assertTrue(concordance >= 0.99, "Concordance with GATK 3.5 in VCF mode is < 99%");
+    }
+
+    /*
+ * Test that in VCF mode we're >= 99% concordant with GATK3.5 results
+ */
+    @Test(expectedExceptions = SparkException.class) //this should be a UserException, but spark exceptions are not unwrapped yet
```
akiezun: so this test actually blows up? if yes, why does it blow up?

src/main/java/org/broadinstitute/hellbender/tools/spark/SparkReadShard.java
```diff
@@ -0,0 +1,52 @@
...
+import org.broadinstitute.hellbender.utils.SimpleInterval;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+
+import java.io.Serializable;
+import java.util.Iterator;
+
+public class SparkReadShard implements ReadShard, Serializable {
+    private static final long serialVersionUID = 1L;
+
+    private final ShardBoundary key;
```
akiezun: what is `ShardBoundary`? It's not imported and it's not in the same package (?)

akiezun: Oh I see, it's an inner class of ReadShard. That's too fancy. Can you take that class out and make it a top-level? It does not seem to need to be an non-static inner class (ie does not the pointer to enclosing instance as far as i can tell)

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
```diff
@@ -0,0 +1,329 @@
...
+    private static final long serialVersionUID = 1L;
+
+    @Argument(fullName= StandardArgumentDefinitions.OUTPUT_LONG_NAME, shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME, doc = "Put the output vcf here")
+    public String output;
+
+    @ArgumentCollection
+    public final ShardingArgumentCollection shardingArgs = new ShardingArgumentCollection();
+
+    public static class ShardingArgumentCollection implements Serializable {
```
akiezun: should this arg collection be used also in HaplotypeCaller?

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
```diff
@@ -0,0 +1,329 @@
...
+        @Argument(fullName="readShardPadding", shortName="readShardPadding", doc = "Each read shard has this many bases of extra context on each side. Read shards must have as much or more padding than assembly regions.", optional = true)
+        public int readShardPadding = HaplotypeCaller.DEFAULT_READSHARD_PADDING;
+
+        @Argument(fullName = "minAssemblyRegionSize", shortName = "minAssemblyRegionSize", doc = "Minimum size of an assembly region", optional = true)
+        public int minAssemblyRegionSize = HaplotypeCaller.DEFAULT_MIN_ASSEMBLY_REGION_SIZE;
+
+        @Argument(fullName = "maxAssemblyRegionSize", shortName = "maxAssemblyRegionSize", doc = "Maximum size of an assembly region", optional = true)
+        public int maxAssemblyRegionSize = HaplotypeCaller.DEFAULT_MAX_ASSEMBLY_REGION_SIZE;
+
+        @Argument(fullName = "assemblyRegionPadding", shortName = "assemblyRegionPadding", doc = "Amount of additional bases of context to include around each assembly region", optional = true)
```
akiezun: `Amount of additional bases` -> `Number of additional bases`

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
```diff
@@ -0,0 +1,329 @@
...
+     * @param ctx the spark context
+     * @param reads the reads variants should be called from
+     * @param header the header that goes with the reads
+     * @param reference the reference to use when calling
+     * @param intervals the intervals to restrict calling to
+     * @param hcArgs haplotype caller arguments
+     * @param shardingArgs arguments to control how the assembly regions are sharded
+     * @return an RDD of Variants
+     */
+    public static JavaRDD<VariantContext> callVariantsWithHaplotypeCaller(
```
akiezun: why is this method public?

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
```diff
@@ -0,0 +1,329 @@
...
+            final AuthHolder authHolder,
+            final JavaSparkContext ctx,
+            final JavaRDD<GATKRead> reads,
+            final SAMFileHeader header,
+            final ReferenceMultiSource reference,
+            final List<SimpleInterval> intervals,
+            final HaplotypeCallerArgumentCollection hcArgs,
+            final ShardingArgumentCollection shardingArgs) {
+        Utils.validateArg(hcArgs.dbsnp.dbsnp == null && hcArgs.comps.isEmpty(),
```
akiezun: split in two lines to make a better error message

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
```diff
@@ -0,0 +1,329 @@
...
+            return variantContexts.stream().filter(shardBoundary::contains).collect(Collectors.toList());
+        };
+    }
+
+    /**
+     * WriteVariants, this is currently going to be horribly slow and explosive on a full size file since it performs a collect.
+     */
+    private void writeVariants(JavaRDD<VariantContext> variants) {
+        final List<VariantContext> collectedVariants = variants.collect();
+        final List<VariantContext> sortedVariants = collectedVariants.stream().sorted((o1, o2) -> IntervalUtils.compareLocatables(o1, o2, getReferenceSequenceDictionary())).collect(Collectors.toList());
```
akiezun: that call to `getReferenceSequenceDictionary()` is going to be done a lot of times. Can you call once and reuse? Wont save much (collect dominates, that's for sure) but it's a simple change

akiezun: also add a comment here that it will be replaced by a shared write at some point (with a ticket number)

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
```diff
@@ -0,0 +1,329 @@
...
+            final AuthHolder authHolder,
+            final Broadcast<ReferenceMultiSource> reference,
+            final Broadcast<HaplotypeCallerArgumentCollection> hcArgsBroadcast,
+            final ShardingArgumentCollection assemblyArgs,
+            final SAMFileHeader header) {
+        return shard -> {
+            final ReferenceMultiSource referenceMultiSource = reference.value();
+            final ReferenceMultiSourceToReferenceDataSourceAdaptor refDataSource = new ReferenceMultiSourceToReferenceDataSourceAdaptor(referenceMultiSource, authHolder);
+            final ReferenceContext refContext = new ReferenceContext(refDataSource, shard.getPaddedInterval());
+            final ReferenceMultiSourceToReferenceSequenceFileAdaptor referenceReader = new ReferenceMultiSourceToReferenceSequenceFileAdaptor(reference.getValue(), authHolder);
```
akiezun: you already have `reference.getValue()` stored in the `referenceMultiSource` value so no reason to get it again

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
```diff
@@ -0,0 +1,329 @@
...
+        private static final long serialVersionUID = 1L;
+
+        private final ReferenceMultiSource source;
+        private final AuthHolder auth;
+        private final SAMSequenceDictionary sequenceDictionary;
+
+        public ReferenceMultiSourceToReferenceSequenceFileAdaptor(final ReferenceMultiSource source, final AuthHolder auth) {
+            this.source = source;
+            this.auth = auth;
+            this.sequenceDictionary = getSequenceDictionary();
```
akiezun: calling virtual method during object initialization in constructor is bad practice

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
```diff
@@ -0,0 +1,329 @@
...
+     */
+    private static FlatMapFunction<ReadShard, Tuple2<AssemblyRegion, SimpleInterval>> shardsToAssemblyRegions(
+            final AuthHolder authHolder,
+            final Broadcast<ReferenceMultiSource> reference,
+            final Broadcast<HaplotypeCallerArgumentCollection> hcArgsBroadcast,
+            final ShardingArgumentCollection assemblyArgs,
+            final SAMFileHeader header) {
+        return shard -> {
+            final ReferenceMultiSource referenceMultiSource = reference.value();
+            final ReferenceMultiSourceToReferenceDataSourceAdaptor refDataSource = new ReferenceMultiSourceToReferenceDataSourceAdaptor(referenceMultiSource, authHolder);
```
akiezun: this adapter dancing between `ReferenceSource` and `ReferenceDataSource` is crazy. And then ReferenceSequenceFile is yet another thing. Can this be fixed/unified somehow?

akiezun:
this looks fine to me. I don't understand why some things are the way they are so i put in some questions.

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
lbergelson: done

src/main/java/org/broadinstitute/hellbender/tools/spark/SparkReadShard.java
lbergelson: done

src/main/java/org/broadinstitute/hellbender/tools/spark/SparkReadShard.java
lbergelson: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCaller.java
lbergelson: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
lbergelson: done

src/test/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSparkIntegrationTest.java
lbergelson: It blows up because allele-specific annotations aren't supported in the base haplotype caller for vcf mode.  It is a copy of a matching test in HaplotypeCallerIntegrationTest.  I've added a comment.

src/main/java/org/broadinstitute/hellbender/tools/spark/SparkReadShard.java
lbergelson: made it a top level class

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
lbergelson: I think it should be, but David was doing some parallel work and asked that I not touch that.  I opened #626 to track putting it in service elsewhere.

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
lbergelson: changed here and in AssemblyRegionWalker

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
lbergelson: It's public and static so that It's easy to use haplotype caller as part of a longer pipeline.  I.e. you can just call this on the results of ApplyBQSR and get variants out.  I can remove it / change it if you think there is a different better way to make it modular.  Or add a comment to make it more obvious what it's for.

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
lbergelson: done

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
lbergelson: both done, good catch on getReferenceSequenceDictionary()

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
lbergelson: fixed

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
lbergelson: fixed

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/ReadThreadingAssemblerArgumentCollection.java
lbergelson: I think we didn't have an Arrays.asList() serializer registered which was causing kryo failures.  I added that but forgot to revert this change.

lbergelson: This is more annoying that it should be.  I thought that registering an 'AllelesAsListSerializer' would let 'Kryo' work with these.  However, spark has a number of hardcoded serializers that override the user's choices.  These include a buggy version of 'AllelesAsListSerializer' that is incapable of serializing primitive arrays.  (See [KryoSerializer line 136](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/serializer/KryoSerializer.scala#L136))  I filed a bug report with Chill (https://github.com/twitter/chill/issues/255), but I think it will be a long time before this is addressed and available  (since it would mean reving the chill version in spark...)

lbergelson: Replaced this with Lists.asArrayList() call.

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
lbergelson: This is a longstanding problem, and one that needs to be addressed.  I think it's out of scope of this pull request to fix the proliferation of reference data types though.

I did unify the adaptors classes though, since they're so similar.  (which implies that we really should be able to unify these interfaces...)

src/test/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerIntegrationTest.java
```diff
@@ -311,7 +310,7 @@ public void testHaplotypeCallerRemoveAltAlleleBasedOnHaptypeScores() throws IOEx
     /*
      * Calculate rough concordance between two vcfs, comparing only the positions, alleles, and the first genotype.
      */
-    private double calculateConcordance( final File actual, final File expected ) {
+    public static double calculateConcordance( final File actual, final File expected ) {
```
lbergelson: @droazen I'm moving these to `VariantContextTestUtils` in public, so test files won't depend on each other, but I'm going to do it with the next round of updates if that's alright to avoid needing to do public pr just for this.

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    /**
+     * @return an {@link OverlapDetector} loaded with {@link ShardBoundary}
+     * based on the -L intervals
+     */
+    private static OverlapDetector<ShardBoundary> getShardBoundaryOverlapDetector(final SAMFileHeader header, final List<SimpleInterval> intervals, final int readShardSize, final int readShardPadding) {
+        final OverlapDetector<ShardBoundary> shardBoundaryOverlapDetector = new OverlapDetector<>(0, 0);
+        intervals.stream()
+                .flatMap(interval -> Shard.divideIntervalIntoShards(interval, readShardSize, readShardSize, readShardPadding, header.getSequenceDictionary()).stream())
```
droazen: Call the overload that sets shard size == shard step for you.

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
```diff
@@ -0,0 +1,319 @@
...
+            final ReferenceMultiSource referenceMultiSource = reference.value();
+            final ReferenceMultiSourceAdaptor referenceSource = new ReferenceMultiSourceAdaptor(referenceMultiSource, authHolder);
+            final ReferenceContext refContext = new ReferenceContext(referenceSource, shard.getPaddedInterval());
+
+            //TODO load features as a side input
+            final FeatureContext features = new FeatureContext();
+
+            final Iterable<AssemblyRegion> assemblyRegions = AssemblyRegion.createFromReadShard(
+                    shard, header, refContext, features,
+                    new HaplotypeCallerEngine(hcArgsBroadcast.value(), header, referenceSource),
```
droazen: Try to find a way to re-use the HC engine instead of re-creating it.

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
```diff
@@ -0,0 +1,319 @@
...
+
+            return StreamSupport.stream(assemblyRegions.spliterator(), false)
+                    .map( a -> new Tuple2<>(a, shard.getInterval()))
+                    .collect(Collectors.toList());
+        };
+    }
+
+
+    @VisibleForTesting
+    static final class ReferenceMultiSourceAdaptor implements ReferenceSequenceFile, ReferenceDataSource, Serializable{
```
droazen: Add a one-line comment explaining this thing's existence

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
```diff
@@ -0,0 +1,319 @@
...
+                final ReferenceBases bases = source.getReferenceBases(auth.asPipelineOptionsDeprecated(), new SimpleInterval(contig, (int) start, (int) stop));
+                return new ReferenceSequence(contig, sequenceDictionary.getSequenceIndex(contig), bases.getBases());
+            } catch (final IOException e) {
+                throw new GATKException(String.format("Failed to load reference bases for %s:%d-%d", contig, start, stop));
+            }
+        }
+
+        @Override
+        public void close() {
+            // should this do something?
```
droazen: Close your `ReferenceMultiSource`?

src/test/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSparkIntegrationTest.java
```diff
@@ -0,0 +1,195 @@
...
+import org.broadinstitute.hellbender.utils.test.SparkTestUtils;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.Collections;
+
+import static org.broadinstitute.hellbender.tools.walkers.haplotypecaller.HaplotypeCallerIntegrationTest.TEST_FILES_DIR;
+import static org.broadinstitute.hellbender.tools.walkers.haplotypecaller.HaplotypeCallerIntegrationTest.calculateConcordance;
```
droazen: questionable use of static imports

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
droazen: Adaptor -> Adapter

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/LikelihoodEngineArgumentCollection.java
```diff
@@ -6,11 +6,15 @@
 import org.broadinstitute.hellbender.utils.QualityUtils;
 import org.broadinstitute.hellbender.utils.pairhmm.PairHMM;
 
+import java.io.Serializable;
+
+import java.io.Serializable;
```
lbergelson: done

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
```diff
@@ -0,0 +1,319 @@
...
+ *
+ * This is an implementation of {@link HaplotypeCaller} using spark to distribute the computation.
+ * It is still in an early stage of development and does not yet support all the options that the non-spark version does.
+ */
+@CommandLineProgramProperties(summary = "HaplotypeCaller on Spark", oneLineSummary = "HaplotypeCaller on Spark", programGroup = SparkProgramGroup.class)
+public final class HaplotypeCallerSpark extends GATKSparkTool {
+    private static final long serialVersionUID = 1L;
+
+    @Argument(fullName= StandardArgumentDefinitions.OUTPUT_LONG_NAME, shortName = StandardArgumentDefinitions.OUTPUT_SHORT_NAME, doc = "Put the output vcf here")
+    public String output;
```
lbergelson: done, it's not

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
```diff
@@ -0,0 +1,319 @@
...
+            final AuthHolder authHolder,
+            final JavaSparkContext ctx,
+            final JavaRDD<GATKRead> reads,
+            final SAMFileHeader header,
+            final ReferenceMultiSource reference,
+            final List<SimpleInterval> intervals,
+            final HaplotypeCallerArgumentCollection hcArgs,
+            final ShardingArgumentCollection shardingArgs) {
+        Utils.validateArg(hcArgs.dbsnp.dbsnp == null, "HaplotypeCallerSpark does not yet support -D or --dbsnp arguments" );
+        Utils.validateArg(hcArgs.comps.isEmpty(), "HaplotypeCallerSpark does not yet support -comp or --comp arguments" );
```
lbergelson: done

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
```diff
@@ -0,0 +1,319 @@
...
+            final AuthHolder authHolder,
+            final SAMFileHeader header,
+            final Broadcast<ReferenceMultiSource> referenceBroadcast,
+            final Broadcast<HaplotypeCallerArgumentCollection> hcArgsBroadcast) {
+        return regionAndInterval -> {
+            final ReferenceMultiSourceAdaptor referenceReader = new ReferenceMultiSourceAdaptor(referenceBroadcast.getValue(), authHolder);
+            final HaplotypeCallerEngine hcEngine = new HaplotypeCallerEngine(hcArgsBroadcast.value(), header, referenceReader);
+            final List<VariantContext> variantContexts = hcEngine.callRegion(regionAndInterval._1(), new FeatureContext());
+            final SimpleInterval shardBoundary = regionAndInterval._2();
+            return variantContexts.stream().filter(shardBoundary::contains).collect(Collectors.toList());
```
lbergelson: fixed

src/test/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSparkIntegrationTest.java
lbergelson: fixed

src/main/java/org/broadinstitute/hellbender/tools/HaplotypeCallerSpark.java
lbergelson: as far as I can tell there isn't really anything to close with a 2-bit

PR #620, opened 2016-07-21
GenotypeGVCF speedup by reusing merger and calculators

akiezun:
15% speedup to GenotypeGVCFs by reusing merger and calculators

@lbergelson can you review?


lbergelson:
👍

PR #622, opened 2016-07-25
use hdf5 from maven

akiezun:
using hdf5 bindings library from maven central (source here: https://github.com/broadinstitute/hdf5-java-bindings)

- deleted all HDF5 setup info from Readme because it should work automagically now. 
- removed install steps from travis
- removed hardwired locations and java libarry path dancing from build.gradle
- updated gradle to 2.13 (something blew up without it, dont remember what)
- changed imports to use new library
- loading the library on start of any tool that uses it and throw UserException if that fails

@lbergelson can you have a look? 
btw, how/when should the library be loaded? Should we put it on the top of main somewhere (would require moving this to public, which is not a bad idea in itself)


akiezun:
(sorry about the multiple commits. will squash before merging)

akiezun:
@SHuang-Broad was able to use this branch on his 'almost vanilla' osx 10.11 install

src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateCoverageComponents.java
```diff
@@ -80,6 +81,9 @@
 
     @Override
     protected void runPipeline(final JavaSparkContext ctx) {
+        if (! new HDF5Library().load(null)){
```
lbergelson: Why are you passing in null?  Shouldn't we be passing in a tempdir to use?

src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateCoverageComponents.java
```diff
@@ -80,6 +81,9 @@
...
+            throw new UserException.HardwareFeatureException("Cannot load the required HDF5 library");
```
lbergelson: Should these warnings be more verbose?  ex.  "HDF5 is supported on OSX and x86-64, if you are not running on one of these platforms you'll need to provide your own native hdf5 binaries by specifying....."

(is it possible to specify a different binary than we package?)

lbergelson:
@akiezun I'm not sure where the right place to load this is.  I think that it depends on what happens if you try to call a method that relies on it but it isn't loaded?  Do we get a sane error?  

Since HDF5 native is mandatory for the tools that rely on it, I think that it's not too terrible to have an upfront check in those tools to make sure that it works and a helpful error message if it doesn't.  That only makes sense if you get a reasonable error message if you forget to load it manually though.  

I would prefer if we don't push it into main unless we also had some code that uses it in public, since I think that would add additional complexity and another potential point of failure without much benefit.  Better to keep public with the minimal number of dependencies.

lbergelson:
+61 -1356 is a nice change!

akiezun:
@lbergelson are you done reviewing? how do you want to proceed?

src/test/java/org/broadinstitute/hellbender/utils/hdf5/HDF5LibraryUnitTest.java
```diff
@@ -43,32 +46,24 @@
     @SuppressWarnings("FieldCanBeLocal")
     private static double TEST_PON_VERSION = 1.3;
 
-    @Test(groups = "supported")
+    @BeforeClass
     public void testIsSupported() {
-        Assert.assertTrue(HDF5Library.isSupported());
+        Assert.assertTrue(new HDF5Library().load(null));
```
lbergelson: Do we want to make these skip if it's not on an architecture we expect to be able to support?  i.e. power?

lbergelson:
@akiezun Done reviewing, interested to hear what the error is for running hdf5 calls with an unloaded library. 

I think it's probably best to just do what you've done and put the library loading at the start of each of the tools that use the library.  That way it's very clear what's happening.  It's a bit awkward to have to load it at the start of every tool, but since it's required for those tools to work at all, it's better to have a check upfront instead of failing later on.  We wouldn't want to fail during main though, since most tools are unaffected by HDF5 failures.  

The only other places I see that we could put it would be in some static block which always has it's own problems,  or extend the GatkTool base classes with a `requiresHDF5` method which seems like it's integrating this niche thing to deeply into our core classes.

I might make a static method somewhere that does the load check and fail so that the error messages remain consistent between tools.  

Maybe that should be something we should add to the GatkNative interface itself, instead of having just `boolean load()` we could have `boolean tryLoad` and `void loadOrDie`.  That way the library itself could provide more detailed information about why it failed to load correctly.

src/test/java/org/broadinstitute/hellbender/utils/hdf5/HDF5LibraryUnitTest.java
akiezun: probably not. The reason is that we have no java equivalent for hdf5. So I'd prefer to know right away when stuff stops working on travis rather than have it skip silently.

lbergelson: I guess the tests for the tools are going to fail as well on power, so we're not changing the status quo here really.

src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateCoverageComponents.java
akiezun: i will improve the message. maybe it is possible. i dont think we should mention it unless we've tested that it works. I'm just going to say `Cannot load the required HDF5 library. HDF5 is currently supported on x86-64 architecture and Linux or OSX systems.`

src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateCoverageComponents.java
akiezun: passing null means using the default temp dir. I added a note.

akiezun:
merging then. enter tickets or PRs for improvements

PR #647, opened 2016-08-19
Fixed inconsistent output of NaN values in ACNV -> ACS conversion.

### LINKED ISSUE #617: Absolute compatibility issue
Hi GATK team, 

There is a small issue with the ACS style output of GATK-ACNV
Segments that have non-nan values for mu major and mu minor but nan values for sigma cause ABSOLUTE to crash.

The fix would be to output the segments as NaN for mu major, mu minor and f if any of the other values are NaN. 

(In this case it was a segment with a length of 3 targets)

-Dimitri

samuelklee:
Closes #617.


samuelklee:
@LeeTL1220, could you please review?  It's a short one!

LeeTL1220:
Ignoring the coveralls issue...

@samuelklee Good to go... I'm merging.

PR #655, opened 2016-08-30
Port supporting classes for Mutect2

davidbenjamin:


takutosato:
@davidbenjamin you should be able to git-mention me in GATK-protected now. 

I will familiarize myself with your port and the GATK4 code in general (but won't do a full-on review just yet?)

davidbenjamin:
@takutosato You can proceed with a full review.  Without tests it's not guaranteed that it all *works*, but without the code we can't write tests.  In your review, just do your best to confirm that all the changes to conform to GATK4 probably replicate the old behavior.  I'm sure some mistakes will slip through, but as long as there aren't too many we should be able to clean them up with debugging.

I have tried to make the commits as user-friendly as possible.  Each ported class starts with a direct port that is merely a copy-and-paste with updated package names.  I then commit one change at a time.  You may find it helpful to trace these in your IDE to see with red underlined code each code is addressed to.  Also, please ask me about anything.  Some of these changes were only clear after digging into GATK3 and GATK4 source code and there's no need to duplicate that effort.

davidbenjamin:
@kcibul Your eyes are always welcome, too.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/PerAlleleCollection.java
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * may be a container for allele fractions for ALT alleles in a variant context. While
+ *
+ *   PerAlleleCollection<Double> alleleCount = PerAlleleCollection.createPerRefAndAltAlleleCollection()
+ *
+ * may hold the allele counts for the REF allele and all ALT alleles in a variant context.
+ *
+ *
+ **/
+public class PerAlleleCollection<X> {
```
magicDGS: Request: could this class be in public? It does not contain any logic for Mutect, and I find that it could have quite interesting usages for other GATK-framework users....

davidbenjamin: If @droazen consents, sure.

davidbenjamin:
Superseded by PR #664.

PR #658, opened 2016-08-31
fix hdf5 test failures

lbergelson:
This fixes it for me in my docker image.  @LeeTL1220 Could you test this.  Don't merge this yet, I'd like to wait until we merge https://github.com/broadinstitute/hdf5-java-bindings/pull/6 and have a non-snapshot version.


lbergelson:
@LeeTL1220 I've updated this with the released version.  It should be good to go.  Please test that it works in your docker script.

PR #660, opened 2016-09-01
Changed allele-fraction model to use slice sampling.

samuelklee:
The allele-fraction model was previous sampled using adaptive Metropolis-Hastings sampling.  However, we did not implement thinning of samples, which resulted in many autocorrelated posterior samples, leading to repeated values in reported posterior deciles.  This PR changes the allele-fraction model to use slice sampling instead.

Also:
-Removed AlleleFractionSamplersUnitTest, which were specifically testing MH sampling.  I also think the allele-fraction samplers are essentially covered by the AlleleFractionModellerUnitTest.
-Minor-allele fraction is a bounded random variable, so minor changes to the SliceSampler class were required.  Unit tests for sampling of a beta distribution were added to cover this case.
-Tweaked the failure thresholds for other MCMC tests.
-Optimized imports, miscellaneous cleanup.

EDIT:  Looks like this bumps up the runtime about 2-3x (due to the increased number of likelihood evaluations required for slice sampling) to about 30-40 min for an exome.  I would guess that the autocorrelation scale is higher than this, so it's probably faster than doing MH and thinning.

Although we can adjust some of the default parameters (like number of MCMC refits per similar-segment merging iteration, similar-segment merging thresholds, etc.) to bring this down, I think we could probably optimize the AF likelihood calculation as well.


src/main/java/org/broadinstitute/hellbender/utils/mcmc/SliceSampler.java
```diff
@@ -68,23 +68,27 @@ public SliceSampler(final RandomGenerator rng, final Function<Double, Double> lo
      * @return              sample drawn from the probability density function
      */
     public double sample(final double xInitial) {
-        final double xSample = Math.min(Math.max(xInitial, xMin + EPSILON), xMax - EPSILON);
+        Utils.validateArg(xMin < xInitial && xInitial < xMax, "Initial point in slice sampler is not within specified range.");
```
samuelklee: putting a reminder to myself to update the javadoc for xInitial

src/main/java/org/broadinstitute/hellbender/tools/exome/allelefraction/AlleleFractionModeller.java
```diff
@@ -172,11 +175,11 @@ public void fitMCMC(final int numSamples, final int numBurnIn) {
...
 
     //use width of a probability distribution given the position of its mode (estimated from Gaussian approximation) as step size
-    private static double getStepSizeFromApproximatePosteriorWidthAtMode(final Function<Double, Double> logPDF, final double mode) {
+    private static double approximatePosteriorWidthAtMode(final Function<Double, Double> logPDF, final double mode) {
         final double absMode = Math.abs(mode);
-        final double EPSILON = Math.min(1e-6, absMode / 2);    //adjust scale is mode is very near zero
-        final double DEFAULT = absMode / 10;                   //if "mode" is not close to true mode of logPDF, approximation may not apply; just use 1/10 of absMode in this case
-        final double secondDerivative = (logPDF.apply(mode + EPSILON) - 2*logPDF.apply(mode) + logPDF.apply(mode - EPSILON))/(EPSILON*EPSILON);
-        return secondDerivative < 0 ? Math.sqrt(-1.0 / secondDerivative) : DEFAULT;
+        final double epsilon = Math.min(1e-6, absMode / 2);    //adjust scale is mode is very near zero
```
davidbenjamin: *if* mode is. . .  Typo was originally mine, of course.

src/main/java/org/broadinstitute/hellbender/utils/mcmc/SliceSampler.java
```diff
@@ -68,23 +68,27 @@ public SliceSampler(final RandomGenerator rng, final Function<Double, Double> lo
...
+        double xLeft = xInitial - width * rng.nextDouble();
         double xRight = xLeft + width;
 
         int k = MAXIMUM_NUMBER_OF_DOUBLINGS;
         //expand slice by doubling until it brackets logPDF
-        while (k > 0 && (logPDF.apply(xLeft) > logSliceHeight || logPDF.apply(xRight) > logSliceHeight) &&
-                xLeft > xMin && xRight < xMax) {
+        double logPDFLeft = xLeft > xMin ? logPDF.apply(xLeft) : Double.NEGATIVE_INFINITY;
+        double logPDFRight = xRight < xMax ? logPDF.apply(xRight) : Double.NEGATIVE_INFINITY;
+        while (k > 0 && ((logSliceHeight < logPDFLeft || logSliceHeight < logPDFRight))) {
```
davidbenjamin: IMO the idiom of `k = MAXIMUM. . . while (k > 0 . . .) . . . k--` is less clear than `count = 0 . . . while (count < MAXIMUM . . .) . . . count++`.

src/test/java/org/broadinstitute/hellbender/tools/exome/SegmentUtilsUnitTest.java
```diff
@@ -22,10 +22,10 @@
 
     @Test
     public void testReadUncalledSegments() {
-        final File TEST_DIR = new File("src/test/resources/org/broadinstitute/hellbender/tools/exome/caller");
-        final File TEST_SEGMENTS = new File(TEST_DIR,"segments.tsv");
+        final File testDir = new File("src/test/resources/org/broadinstitute/hellbender/tools/exome/caller");
```
davidbenjamin: why not give this class scope?

davidbenjamin:
@samuelklee Looks good.  Address little complaints and merge.

src/main/java/org/broadinstitute/hellbender/tools/exome/allelefraction/AlleleFractionModeller.java
samuelklee: done

src/main/java/org/broadinstitute/hellbender/utils/mcmc/SliceSampler.java
samuelklee: Just following the way it's done in Neal 2003.  Will add a comment somewhere to this effect!

src/test/java/org/broadinstitute/hellbender/tools/exome/SegmentUtilsUnitTest.java
samuelklee: I guess it was like this originally because it is only used in this test, but done.

davidbenjamin:
@samuelklee good to go.

samuelklee:
thanks @davidbenjamin!  for the record, also made one additional change to prevent a (literal) edge case in the slice sampler.

PR #664, opened 2016-09-01
Port Mutect2

davidbenjamin:


davidbenjamin:
We don't have tests yet, but manual review of the commented-out test I wrote looked good.  Let's get this in the code base and then write tests.

Note: @LeeTL1220 OK'ed the temporary hit in coverage.  It will be corrected soon.

davidbenjamin:
@takutosato This replaces the previous PR assigned to you.

@kcibul here's the port to GATK4.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/PerAlleleCollection.java
```diff
@@ -31,75 +32,54 @@
...
      */
-    public void set(Allele allele, X newValue){
-        if (allele == null || newValue == null){
-            throw new IllegalArgumentException("allele or newValue is null");
-        }
-        if (allele.isReference() && type == Type.ALT_ONLY){
-            throw new IllegalArgumentException("Collection stores values for alternate alleles only");
-        }
+    public void set(Allele allele, X value){
+        Utils.nonNull(allele, "allele is null");
```
takutosato: love these new argument checking utility functions!

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/M2ArgumentCollection.java
```diff
@@ -13,6 +13,7 @@
 import java.util.List;
 
 public class M2ArgumentCollection extends AssemblyBasedCallerArgumentCollection {
+    private static final long serialVersionUID = 9341L;
```
takutosato: What's serial version UID for? (I will leave a bunch of comments/questions on github - maybe we can go through them in person on Tuesday?)

takutosato: nvm, google gave me an answer

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/MuTect2.java
```diff
@@ -1065,4 +1002,56 @@ public String getTumorSampleName(){
         }
         return annotatedCalls;
     }
+
+    //TODO: See if we can share code with HaplotypeCaller's isActive
```
takutosato: Yea we definitely should!

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/MuTect2.java
```diff
@@ -612,6 +583,19 @@ public ActivityProfileState isActive(final RefMetaDataTracker tracker, final Ref
         return annotateVCs(calledHaplotypes, metaDataTracker);
     }
 
+    @Override
```
takutosato: Maybe add a TODO here to mark that the method is duplicated from HC and that we should do something about it?

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/MuTect2.java
```diff
@@ -954,23 +908,6 @@ private Haplotype createReferenceHaplotype(final AssemblyRegion assemblyRegion,
         return ReferenceConfidenceModel.createReferenceHaplotype(assemblyRegion, assemblyRegion.getAssemblyRegionReference(referenceReader), paddedReferenceLoc);
     }
 
-    /**
-     * Clean up reads/bases that overlap within read pairs
-     *
-     * @param reads the list of reads to consider
-     */
-    private void cleanOverlappingReadPairs(final List<GATKRead> reads, final Set<String> normalSampleNames) {
```
takutosato: This method was not used in GATK3 M2, but GATK4 HC does. Might be useful to keep it (and find the right place to use it).

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/MuTect2.java
```diff
@@ -0,0 +1,1057 @@
...
+        final Set<String> filters = new HashSet<>();
+
+        final Integer eventCount = (Integer) eventDistanceAttributes.get(GATKVCFConstants.EVENT_COUNT_IN_HAPLOTYPE_KEY);
+        final Integer maxEventDistance = (Integer) eventDistanceAttributes.get(GATKVCFConstants.EVENT_DISTANCE_MAX_KEY);
+
+        //TODO: verify that vc.getStart() is the right second argument to use here
+        final Collection<VariantContext> panelOfNormalsVC = featureContext.getValues(MTAC.normalPanelFeatureInput, vc.getStart());
+        final VariantContext ponVc = panelOfNormalsVC.isEmpty()?null:panelOfNormalsVC.iterator().next();
+
+        if (ponVc != null) {
```
takutosato: How about:

```java
if (!panelOfNormalsVC.isEmpty()) { filters.add(GATKVCFConstants.PON_FILTER_NAME); }
```

takutosato: I think the correct thing to do here is to filter only if the alleles match. Maybe the genotype is important too (e.g. if the variant in PON is hom var then don't filter).

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/SomaticGenotypingEngine.java
```diff
@@ -0,0 +1,558 @@
...
+    private static AFCalculatorProvider DUMMY_AF_CALCULATOR_PROVIDER = new AFCalculatorProvider() {
+        public AFCalculator getInstance(final int ploidy, final int maximumAltAlleles) { return null; }
+    };
+
+    private final static Logger logger = Logger.getLogger(SomaticGenotypingEngine.class);
+
+    public SomaticGenotypingEngine(final M2ArgumentCollection configuration,
+                                   final SampleList samples,
+                                   final boolean doPhysicalPhasing,
+                                   final M2ArgumentCollection MTAC,
```
takutosato: we are passing `MTAC` as argument twice here

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/MuTect2.java
```diff
@@ -1041,5 +1041,8 @@ public ActivityProfileState isActive(final AlignmentContext context, final Refer
             return new ActivityProfileState( ref.getInterval(), prob, ActivityProfileState.Type.NONE, null);
         }
     }
+
+    public static boolean hasBadMate(final GATKRead rec) {
```
takutosato: We should document (and perhaps rename); GATK3 has the following docs about it: "If mates in a pair are mapping to different contigs, it is likely that at least one of them is in the wrong place."

ldgauthier: To be fair, that could also be the correct mapping for a pair of read supporting a structural variant, which are very common (compared with germline) in some tumor types. MuTect2 can't call SVs, but one of the mates being wrong isn't the only possibility.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/MuTect2.java
```diff
@@ -69,8 +69,10 @@
  *     -jar GenomeAnalysisTK.jar \
  *     -T MuTect2 \
  *     -R reference.fasta \
- *     -I:tumor tumor.bam \
- *     -I:normal normal.bam \
+ *     -I tumor.bam \
+ *     -I normal.bam \
+ *     -tumor tumorSampleName.bam \ // as in the BAM header
```
takutosato: should it just say `-tumor tumorSampleName`?

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/MuTect2.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/MuTect2.java
davidbenjamin: I filed an issue for this.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/MuTect2.java
davidbenjamin: 1. Done.  2. put in a TODO and a github issue

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/SomaticGenotypingEngine.java
davidbenjamin: true.  fixed it.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/MuTect2.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/MuTect2.java
davidbenjamin: done, with the SV caveat

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/MuTect2.java
```diff
@@ -0,0 +1,1057 @@
...
+
+        final AssemblyRegion regionForGenotyping = assemblyResult.getRegionForGenotyping();
+        logReadInfo(DEBUG_READ_NAME, regionForGenotyping.getReads(), "Present in region for genotyping");
+
+        // filter out reads from genotyping which fail mapping quality based criteria
+        //TODO - why don't do this before any assembly is done? Why not just once at the beginning of this method
+        //TODO - on the originalAssemblyRegion?
+        //TODO - if you move this up you might have to consider to change referenceModelForNoVariation
+        //TODO - that does also filter reads.
+        final Collection<GATKRead> filteredReads = filterNonPassingReads(regionForGenotyping);
```
takutosato: To reflect on these TODOs, one reason to filter reads here (as opposed to before local reassembly)  is `keepRG`; if this argument is specified we use all read groups for local reassembly and only this read group for calling variants. 

That said, `keepRG` does not seem too useful to me. Might wanna just get rid of it.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/MuTect2.java
```diff
@@ -395,7 +396,10 @@ private VariantAnnotatorEngine initializeVCFOutput() {
         // if printTCGAsampleHeader, we already checked for exactly 1 tumor and 1 normal in printTCGAsampleHeader assignment in initialize()
         final List<String> outputSampleNames = printTCGAsampleHeader ? Arrays.asList("TUMOR", "NORMAL") : samplesList.asListOfSamples();
 
-        vcfWriter.writeHeader(new VCFHeader(headerInfo, outputSampleNames));
+        //TODO: figure out what was outputting a null line (there was exactly 1) and remove this hack!!!
```
takutosato: I would guess (the absence of) `clusteredReadPosition` annotator is causing the problem here

src/test/java/org/broadinstitute/hellbender/tools/walkers/mutect/MuTect2IntegrationTest.java
```diff
@@ -19,9 +19,15 @@
...
+    //TODO: note this test works if you run it locally and mount /seq/references
+    //TODO: it's commented out because Travis will fail, but it's useful to have for development until
+    //TODO: real imntegration tests come
+    //TODO: currently, results look reasonable on what is admittedly a pretty tiny test
+
     /*
      * Run the tool
      */
+    /*
     @Test
```
takutosato: minor detail but how about replacing `@test` with the `@ignore` testNG annotation instead of comment?

takutosato:
Review complete. @davidbenjamin thank you for dividing up the changes into small, related commits; it made reviewing so much easier. Back to you.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/MuTect2.java
davidbenjamin: I can't see how it would be used either, but perhaps @ldgauthier @kcibul would know its purpose.

src/test/java/org/broadinstitute/hellbender/tools/walkers/mutect/MuTect2IntegrationTest.java
davidbenjamin: Done.  That's much cleaner.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/MuTect2.java
davidbenjamin: I added some `TODO`s and filed an issue.  I would correct it but `GATKVCFHeaderLines` is in gatk public.

davidbenjamin: I added some `TODO`s and filed an issue.  I would correct it but `GATKVCFHeaderLines` is in gatk public.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/MuTect2.java
davidbenjamin: Since HC uses it inside `finalizeRegion`, the natural place to deal with this would be when we fix the code duplication between that and M2's equivalent method, issue #670.

davidbenjamin:
@takutosato back to you.  I followed all the straightforward suggestions and filed issues for the rest.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/MuTect2.java
ldgauthier: The way we do the DeBruijn-like graph assembly, variants with fewer reads are more prone to being missed because they happen to fall on dangling tails or because they end up in a large kmer (with at least one other variant) that isn't merged back to the reference and gets pruned.  I can't speak to the original intention, but it will make the assembly slightly more robust to include the low MQ reads.

davidbenjamin: I could believe that explanation.  If that's the case we may be able to improve things by using somatic-specific parameters for assembly instead of borrowing them from `HaplotypeCaller`.

takutosato:
@davidbenjamin the latest commit looks good. Please squash and merge!

PR #695, opened 2016-09-08
Added feature to take user provided target list

mbabadi:
It can be used for:

(1) passing the tool unpadded target lists
(2) passing the tool a target list with PAR regions excluded


asmirnov239:
Looks good! Go ahead and merge

PR #699, opened 2016-09-09
Extract code shared between Mutect2 and HaplotypeCaller

### LINKED ISSUE #668: Code duplication of realignReadsToTheirBestHaplotype
Virtually identical methods exist in `Mutect2` and `HaplotypeCallerEngine`.  Extract a common method, perhaps putting in a new `HaplotypeCallerUtils` class.

### LINKED ISSUE #670: Code duplication of finalizeAssemblyRegion between Mutect2 and HaplotypeCaller
This one seems straightforward -- there doesn't appear to be any somatic-specific logic, so a common method can be extracted and put in a utils class.

### LINKED ISSUE #675: Organize Mutect2 command line argument
Some arguments belong in `Mutect2ArgumentCollection`.  The annotation arguments probably belong in or already exist in an argument colleciton used by `HaplotypeCaller`. et cetera

davidbenjamin:
Closes #668, closes #670, closes #675.  Various other code improvements.

@takutosato I did the same mini-integration test as before to sanity-check output. 

@kcibul keeping you in the loop.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerUtils.java
@@ -1,0 +1,181 @@
+package org.broadinstitute.hellbender.tools.walkers.haplotypecaller;
+
+import htsjdk.samtools.SAMFileHeader;
+import htsjdk.samtools.reference.IndexedFastaSequenceFile;
+import htsjdk.samtools.util.Locatable;
+import org.broadinstitute.hellbender.engine.AssemblyRegion;
+import org.broadinstitute.hellbender.exceptions.UserException;
+import org.broadinstitute.hellbender.utils.QualityUtils;
+import org.broadinstitute.hellbender.utils.SimpleInterval;
+import org.broadinstitute.hellbender.utils.clipping.ReadClipper;
+import org.broadinstitute.hellbender.utils.fasta.CachingIndexedFastaSequenceFile;
+import org.broadinstitute.hellbender.utils.fragments.FragmentCollection;
+import org.broadinstitute.hellbender.utils.fragments.FragmentUtils;
+import org.broadinstitute.hellbender.utils.genotyper.ReadLikelihoods;
+import org.broadinstitute.hellbender.utils.genotyper.SampleList;
+import org.broadinstitute.hellbender.utils.haplotype.Haplotype;
+import org.broadinstitute.hellbender.utils.read.AlignmentUtils;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+import org.broadinstitute.hellbender.utils.read.ReadCoordinateComparator;
+import org.broadinstitute.hellbender.utils.read.ReadUtils;
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.util.*;
+
+/**
+ * Created by davidben on 9/8/16.
+ */
+public class AssemblyBasedCallerUtils {
+    /**
+     * Returns a map with the original read as a key and the realigned read as the value.
+     * <p>
+     *     Missing keys or equivalent key and value pairs mean that the read was not realigned.
+     * </p>
+     * @return never {@code null}
+     */
+    public static Map<GATKRead, GATKRead> realignReadsToTheirBestHaplotype(final ReadLikelihoods<Haplotype> originalReadLikelihoods, final Haplotype refHaplotype, final Locatable paddedReferenceLoc) {
+        final Collection<ReadLikelihoods<Haplotype>.BestAllele> bestAlleles = originalReadLikelihoods.bestAlleles();
+        final Map<GATKRead, GATKRead> result = new HashMap<>(bestAlleles.size());
+
+        for (final ReadLikelihoods<Haplotype>.BestAllele bestAllele : bestAlleles) {
+            final GATKRead originalRead = bestAllele.read;
+            final Haplotype bestHaplotype = bestAllele.allele;
+            final boolean isInformative = bestAllele.isInformative();
+            final GATKRead realignedRead = AlignmentUtils.createReadAlignedToRef(originalRead, bestHaplotype, refHaplotype, paddedReferenceLoc.getStart(), isInformative);
+            result.put(originalRead, realignedRead);
+        }
+        return result;
+    }
+
+    public static void finalizeRegion(final AssemblyRegion region,
+                                      final boolean errorCorrectReads,
+                                      final boolean dontUseSoftClippedBases,
+                                      final byte minTailQuality,
+                                      final SAMFileHeader readsHeader,
+                                      final SampleList samplesList) {
+        if ( region.isFinalized() ) {
+            return;
+        }
+
+        // Loop through the reads hard clipping the adaptor and low quality tails
+        final List<GATKRead> readsToUse = new ArrayList<>(region.getReads().size());
+        for( final GATKRead myRead : region.getReads() ) {
+            final byte minTailQualityToUse = errorCorrectReads ? HaplotypeCallerEngine.MIN_TAIL_QUALITY_WITH_ERROR_CORRECTION : minTailQuality;
+            GATKRead clippedRead = ReadClipper.hardClipLowQualEnds(myRead, minTailQualityToUse);
+
+            // remove soft clips if we cannot reliably clip off adapter sequence or if the user doesn't want to use soft clips at all
+            // otherwie revert soft clips so that we see the alignment start and end assuming the soft clips are all matches
+            // TODO -- WARNING -- still possibility that unclipping the soft clips will introduce bases that aren't
+            // TODO -- truly in the extended region, as the unclipped bases might actually include a deletion
+            // TODO -- w.r.t. the reference.  What really needs to happen is that kmers that occur before the
+            // TODO -- reference haplotype start must be removed
+            clippedRead = dontUseSoftClippedBases || ! ReadUtils.hasWellDefinedFragmentSize(clippedRead) ?
+                    ReadClipper.hardClipSoftClippedBases(clippedRead) : ReadClipper.revertSoftClippedBases(clippedRead);
+
+            clippedRead = clippedRead.isUnmapped() ? clippedRead : ReadClipper.hardClipAdaptorSequence(clippedRead);
+            if ( ! clippedRead.isEmpty() && clippedRead.getCigar().getReadLength() > 0 ) {
+                clippedRead = ReadClipper.hardClipToRegion( clippedRead, region.getExtendedSpan().getStart(), region.getExtendedSpan().getEnd() );
+                if ( region.readOverlapsRegion(clippedRead) && clippedRead.getLength() > 0 ) {
+                    readsToUse.add(clippedRead);
+                }
+            }
+        }
+
+        // TODO -- Performance optimization: we partition the reads by sample 4 times right now; let's unify that code.
+        // final List<GATKRead> downsampledReads = DownsamplingUtils.levelCoverageByPosition(ReadUtils.sortReadsByCoordinate(readsToUse), maxReadsInRegionPerSample, minReadsPerAlignmentStart);
+        Collections.sort(readsToUse, new ReadCoordinateComparator(readsHeader)); // TODO: sort may be unnecessary here
+
+        // handle overlapping read pairs from the same fragment
+        cleanOverlappingReadPairs(readsToUse, samplesList, readsHeader);
+
+        region.clearReads();
+        region.addAll(readsToUse);
+        region.setFinalized(true);
+    }
+
+    /**
+     * Clean up reads/bases that overlap within read pairs
+     *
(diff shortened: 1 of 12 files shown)


src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/MuTect2.java
```diff
@@ -607,47 +456,40 @@ public void apply(final AssemblyRegion region, final ReferenceContext referenceC
         //TODO: verify that vc.getStart() is the right second argument to use here
         final Collection<VariantContext> panelOfNormalsVC = featureContext.getValues(MTAC.normalPanelFeatureInput, vc.getStart());
 
-        //TODO: is this the correct criterion for filtering?
-        //TODO: Takutp points out that we're filtering based on the PoN even if the alleles are different!
-        //TODO: and what about genotypes?
+        //TODO: Takuto points out that we're filtering even if the alleles are different!  And what about genotypes?
```
kcibul: although it's true that the PON will filter out gemline events (where alleles would matter), if you have a normal it does a WAY better job at that.  Then then PON is primarily used as a "noise filter" (and has a huge effect) but in those cases we don't really trust the data other than "this is a noisy site... don't trust the data" and the alleles don't matter.  

Tumor-only calling is a different thing... and you probably want some of each of the above ;)

kcibul:
So nice to see this being cleaned up!  Before I tried to touch the HC as little as possible... great to see the refactoring here!

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/MuTect2.java
davidbenjamin: Ah, that makes a lot of sense.  I'll put in a comment about that.

davidbenjamin:
@kcibul Yeah, I've been told a lot of these changes are things you had wanted to do before Mutect was fully integrated into the GATK.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerEngine.java
```diff
@@ -811,54 +717,8 @@ public void shutdown() {
...
-        // final List<GATKRead> downsampledReads = DownsamplingUtils.levelCoverageByPosition(ReadUtils.sortReadsByCoordinate(readsToUse), maxReadsInRegionPerSample, minReadsPerAlignmentStart);
-        Collections.sort(readsToUse, new ReadCoordinateComparator(readsHeader)); // TODO: sort may be unnecessary here
-
-        // handle overlapping read pairs from the same fragment
-        cleanOverlappingReadPairs(readsToUse);
-
-        region.clearReads();
-        region.addAll(readsToUse);
-        region.setFinalized(true);
+    private void finalizeRegion(final AssemblyRegion region) {
```
takutosato: Now that `finalizeRegion` is in the util class, we should get rid of this private method and prepend the util class (`AssemblyBasedCallerUtils.finalizeRegion()`) when we call it.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCaller.java
```diff
@@ -209,15 +209,11 @@ public void onTraversalStart() {
 
     @Override
     public void apply(final AssemblyRegion region, final ReferenceContext referenceContext, final FeatureContext featureContext ) {
```
takutosato: (Thinking out loud here)

`apply()` is a one-liner now so duplication is no longer as big of an issue, but it is still a little unsatisfying. I wonder if HC and M2 should share code via an interface with default methods (a mixin?). But then how's that different from having a util class? 

I will come back to this after I go through the PR.

droazen: You also have to think about the issue of sharing code across the walker and spark versions of tools -- for that purpose, static utility methods are by far the best approach.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/MuTect2.java
```diff
@@ -931,32 +691,8 @@ public static void logReadInfo(final String readName, final GATKRead rec, final
...
-            final GATKRead realignedRead = AlignmentUtils.createReadAlignedToRef(originalRead, bestHaplotype, refHaplotype, paddedReferenceLoc.getStart(), isInformative);
-            result.put(originalRead,realignedRead);
-        }
-        return result;
-    }
-
-    //TODO: is there a faster way to do this?
     private boolean isReadFromNormal(final GATKRead rec) {
-        return normalSampleName != null && normalReadGroupIDs.contains(rec.getReadGroup());
+        return normalSampleName != null && ReadUtils.getSampleName(rec, header) == normalSampleName;
```
takutosato: `.equals(normalSampleName)` instead of `== normalSampleName `

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/ReadThreadingAssemblerArgumentCollection.java
```diff
@@ -137,4 +139,21 @@
     @Hidden
     @Argument(fullName="minObservationsForKmerToBeSolid", shortName="minObservationsForKmerToBeSolid", doc = "A k-mer must be seen at least these times for it considered to be solid", optional = true)
     public int minObservationsForKmerToBeSolid = 20;
+
+    public ReadThreadingAssembler createReadThreadingAssembler(final boolean debug, final byte minBaseQualityScore) {
```
takutosato: Awesome!

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerArgumentCollection.java
```diff
@@ -81,12 +70,7 @@
     @Advanced
     @Argument(fullName = "indelSizeToEliminateInRefModel", shortName = "ERCIS", doc = "The size of an indel to check for in the reference model", optional = true)
     public int indelSizeToEliminateInRefModel = 10;
-    
-    /**
-     * Bases with a quality below this threshold will reduced too the minimum usable qualiy score (6).
-     */
-    @Argument(fullName = "base_quality_score_threshold", shortName = "bqst", doc = "Base qualities below this threshold will be reduced to the minimum (" + QualityUtils.MIN_USABLE_Q_SCORE + ")", optional = true)
-    public byte BASE_QUALITY_SCORE_THRESHOLD = PairHMM.BASE_QUALITY_SCORE_THRESHOLD;
```
takutosato: I think `pcrErrorMode` belongs in `LikelihoodEngineArgumentCollection.java` too; it's a parameter for the likelihood engine. Then `createLikelihoodCalculationEngine` doesn't need to take `pcrErrorMode` as an argument (but this is not the primary reason for moving it. We want to move it because it semantically belongs in `LikelihoodEngineArgumentCollection.java`)

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerUtils.java
```diff
@@ -0,0 +1,184 @@
...
+        return phredScaledGlobalReadMismappingRate < 0 ? - Double.MAX_VALUE : QualityUtils.qualToErrorProbLog10(phredScaledGlobalReadMismappingRate);
+    }
+
+    // create the assembly using just high quality reads (eg Q20 or higher).  We may want to use lower
+    // quality reads in the PairHMM downstream, so we can't use a ReadFilter
+    public static AssemblyRegion assemblyRegionWithWellMappedReads(final AssemblyRegion originalAssemblyRegion,
+                                                                   final int minMappingQuality,
+                                                                   final SAMFileHeader readsHeader) {
+        final AssemblyRegion result = new AssemblyRegion(originalAssemblyRegion.getSpan(), originalAssemblyRegion.getSupportingStates(), originalAssemblyRegion.isActive(), originalAssemblyRegion.getExtension(), readsHeader);
+        for (final GATKRead rec : originalAssemblyRegion.getReads()) {
```
takutosato: use a stream for this for loop?

```java
originalAssemblyRegion.getReads().stream().filter(rec -> rec.getMappingQuality() >= minMappingQuality).forEach(rec -> result.add(rec));
```

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerArgumentCollection.java
davidbenjamin: You're right.  A question that then arises is the different defaults between `Mutect2` (`HOSTILE`) and `HaplotypeCaller` (`CONSERVATIVE`).  If it moves to `LikelihoodEngineArgumentCollection` they will share a default, so one tool would need to specify its previous default on the command line.  I don't consider this a big deal, but do you know if it makes much of a difference if M2 is run with the "wrong" `pcrErrorMode`?

ldgauthier: Spoiler alert: the PCR error model doesn't really do anything right now anyway.

Kris originally used the same error mode as HC, but there were still a lot of 3bp deletions (80% sure it was deletions not insertions) in the M2 calls so he cranked it up, but it doesn't so much and we just filter those out after th fact anyway.

kcibul: Yeah -- laura is right... we hoped that using the PCR error model with a lot more hostility would have killed these FPs, but I couldn't make it hostile enough so we put in a hard filter instead.

But would be good to confirm that (unless Laura has?) by running the benchmark set with it both off and on (exomes specifically IIRC).

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -138,27 +138,14 @@ public CalledHaplotypes assignGenotypeLikelihoods(final List<Haplotype> haplotyp
...
-        if (activeAllelesToGenotype == null ) {
-            throw new IllegalArgumentException("activeAllelesToGenotype must be non-null");
-        }
-        if (! refLoc.contains(activeRegionWindow)) {
-            throw new IllegalArgumentException("refLoc must contain activeRegionWindow");
-        }
+        Utils.nonEmpty(haplotypes, "haplotypes input should be non-empty and non-null");
+        Utils.validateArg(readLikelihoods != null && readLikelihoods.numberOfSamples() > 0, "readLikelihoods input should be non-empty and non-null");
+        Utils.validateArg(ref != null && ref.length > 0, "ref bytes input should be non-empty and non-null");
+        Utils.nonNull(refLoc, "refLoc must b non-null");
```
takutosato: refLoc must be* non-null

takutosato:
Review done. Back to you @davidbenjamin

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
davidbenjamin: r u saying something about my diction?

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerArgumentCollection.java
davidbenjamin: In light of comments, I implemented @takutosato's suggestion.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerUtils.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
davidbenjamin: done

davidbenjamin:
@takutosato I did everything you said.  Back to you.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
takutosato: hahaha

takutosato:
looks good! ready to merge. 👍

PR #707, opened 2016-09-16
More refactoring shared M2/HC code

davidbenjamin:


davidbenjamin:
@takutosato More of the same.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -629,39 +626,36 @@ protected static void cleanUpSymbolicUnassembledEvents( final List<Haplotype> ha
...
-        }
+        final Allele ref = mergedVC.getReference();
+        result.put(ref, new ArrayList<>());
 
-        for( final Haplotype h : haplotypes ) {
-            if( h.getEventMap().get(loc) == null ) {
-                eventMapper.get(refEvent).add(h);
+        //Note: we can't use the alleles implied by eventsAtThisLoc because they are not yet merged to a common reference
+        //For example, a homopolymer AAAAA reference with a single and double deletion would yield (i) AA* A and (ii) AAA* A
+        //in eventsAtThisLoc, when in mergedVC it would yield AA* AA A
```
takutosato: If I understood this correctly you meant to say `AAA* AA A`?

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -629,39 +626,36 @@ protected static void cleanUpSymbolicUnassembledEvents( final List<Haplotype> ha
...
+        for (final Haplotype h : haplotypes) {
+            final VariantContext haplotypeVC = h.getEventMap().get(loc);
+            if (haplotypeVC == null) {    //no events --> this haplotype supports the reference at this locus
+                result.get(ref).add(h);
             } else {
-                for( final VariantContext vcAtThisLoc : eventsAtThisLoc ) {
-                    if( h.getEventMap().get(loc).hasSameAllelesAs(vcAtThisLoc) ) {
-                        eventMapper.get(new Event(vcAtThisLoc)).add(h);
+                for (int n = 0; n < eventsAtThisLoc.size(); n++) {
+                    if (haplotypeVC.hasSameAllelesAs(eventsAtThisLoc.get(n))) {
```
takutosato: @davidbenjamin can we discuss this piece of code in person?

takutosato: Notes:

- `eventsAtThisLoc` and `haplotypes` share the same variant contexts (alleles). mergedVC may have alleles that those two don't have (see David's example above). 
- mixing the index `n` between `mergedVC` and `eventsAtThisLoc` may be dangerous as hinted by the comment in original code

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
```diff
@@ -137,131 +59,93 @@
...
 
-        // setup the likelihood calculation engine
-        if ( MTAC.likelihoodArgs.phredScaledGlobalReadMismappingRate < 0 ) MTAC.likelihoodArgs.phredScaledGlobalReadMismappingRate = -1;
+        assemblyEngine = AssemblyBasedCallerUtils.createReadThreadingAssembler(MTAC);
 
-        log10GlobalReadMismappingRate = AssemblyBasedCallerUtils.calculateLog10GlobalReadMismappingRate(MTAC.likelihoodArgs.phredScaledGlobalReadMismappingRate);
+        MIN_TAIL_QUALITY = (byte)(MTAC.minBaseQualityScore - 1);
 
-        likelihoodCalculationEngine = AssemblyBasedCallerUtils.createLikelihoodCalculationEngine(MTAC.likelihoodArgs, log10GlobalReadMismappingRate);
+        if ( MTAC.likelihoodArgs.phredScaledGlobalReadMismappingRate < 0 ) MTAC.likelihoodArgs.phredScaledGlobalReadMismappingRate = -1;
```
takutosato: curly braces around the if statement

takutosato: Also, HC doesn't check if ` MTAC.likelihoodArgs.phredScaledGlobalReadMismappingRate < 0`. Wondering if it's intentional or just a vestigial piece of GATK3 HC from a year and a half ago. Need to come back to this once review is finished.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
```diff
@@ -137,131 +59,93 @@
...
-
-    //TODO: END OF HACK ALERT
-
 
+    protected SampleList samplesList;
+    protected boolean printTCGAsampleHeader = false;
 
-    private VariantContextWriter vcfWriter;
+    protected CachingIndexedFastaSequenceFile referenceReader;
+    protected ReadThreadingAssembler assemblyEngine;
```
takutosato: these variables should be `private`

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
```diff
@@ -578,28 +407,22 @@ protected int getCountOfNonRefEvents(final ReadPileup pileup, final byte refBase
     }
 
     private double calculateF(final ReadPileup pileup, final byte refBase, final byte minBaseQual) {
```
takutosato: I think it's slightly better to use streams here. Taking your code from elsewhere in the project:

```java
int totalCount = (int) StreamSupport.stream(pileup.spliterator(), false).count();
int altCount = (int) StreamSupport.stream(pileup.spliterator(), false)
        .filter(p -> p.isDeletion() || p.getQual() > minBaseQual)
        .filter(p -> isNonRef(refBase, p))
        .count();
return (double) altCount / totalCount;
```

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerUtils.java
```diff
@@ -187,6 +188,10 @@ public static ReadThreadingAssembler createReadThreadingAssembler(final Assembly
         return assemblyEngine;
     }
 
+    public static HaplotypeBAMWriter createBamWriter(final AssemblyBasedCallerArgumentCollection args, final SAMFileHeader header) {
```
takutosato: Instead of returning `null` when bamout path is not defined, we should use an `Optional`. So we will have something like:

```java

public static Optional<HaplotypeBAMWriter> createBamWriter(final AssemblyBasedCallerArgumentCollection args, final SAMFileHeader header) {
        return args.bamOutputPath != null ? Optional.of(HaplotypeBAMWriter.create(args.bamWriterType, new File(args.bamOutputPath), header)) : Optional.empty();
}

```

This clearly signals to the reader that the variable is empty sometimes.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerUtils.java
```diff
@@ -178,4 +181,11 @@ public static AssemblyRegion assemblyRegionWithWellMappedReads(final AssemblyReg
                 .forEach(result::add);
         return result;
     }
+
+    public static VariantContext makeMergedVariantContext(final List<VariantContext> vcs) {
+        final List<String> sources = vcs.stream().map(VariantContext::getSource).collect(Collectors.toList());
```
takutosato: maybe it's a commonly used term that I'm just not familiar with. TODO: discuss with @davidbenjamin

takutosato:
@davidbenjamin back to you!

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -520,7 +520,7 @@ private VariantContext addNonRefSymbolicAllele(final VariantContext mergedVC) {
     }
 
 
-    private Map<String, List<GATKRead>> overlappingFilteredReads(final Map<String, List<GATKRead>> perSampleFilteredReadList, final SimpleInterval loc) {
+    private static Map<String, List<GATKRead>> overlappingFilteredReads(final Map<String, List<GATKRead>> perSampleFilteredReadList, final SimpleInterval loc) {
```
davidbenjamin: `static` here is essentially an annotation that this method doesn't depend on the particular instance of `HaplotypeCallerGenotypingEngine`, which is a useful hint to anyone reading the code.  More generally, I think that any method that *can* be declared `static` *should* be declared `static`, kind of like `final`.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -556,19 +557,11 @@ private VariantContext addNonRefSymbolicAllele(final VariantContext mergedVC) {
                                                                       final byte[] ref,
                                                                       final SimpleInterval refLoc,
                                                                       final List<VariantContext> activeAllelesToGenotype) {
-        final boolean in_GGA_mode = !activeAllelesToGenotype.isEmpty();
+        //TODO: maintaining previous behavior, which cleared the TreeSet from buildEventMaps before adding only the GGA positions
+        //TODO: This seems wrong -- GGA mode shouldn't discard non-GGA loci, right?
```
davidbenjamin: I had always assumed that GGA mode is supposed to discover alleles as usual but *in addition* force specified alleles to be considered.  However, I recently saw some other code that considers *only* the GGA alleles, as was your understanding.

As for how GGA mode works, take a look at `HaplotypeCallerEngine::callRegion`.  Near the top of the method you will find the following code to grab the GGA alleles for a particular assembly region:

```java
for ( final VariantContext vc : features.getValues(hcArgs.alleles) ) {
    if ( vc.isNotFiltered() ) {
        givenAlleles.add(vc); // do something with these VCs during GGA mode
    }
}
```
The key piece of code here is `features.getValues(hcArgs.alleles)`, where `features` is a `FeatureContext`, that is, an interval of an input vcf corresponding to the assembly region, and `hcArgs.alleles` is a `FeatureInput<VariantContext>`, that is, a vcf.  Thus we can see that on the command line one provides a vcf full of GGA alleles, as opposed to an explicit list of alleles, and that `HaplotypeCallerEngine` queries this vcf for alleles overlapping each assembly region.

PS I didn't know how this worked until you asked this question.  GGA mode is definitely not common knowledge.
`

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -634,8 +628,15 @@ protected static void cleanUpSymbolicUnassembledEvents( final List<Haplotype> ha
         haplotypes.removeAll(haplotypesToRemove);
     }
 
-    protected static Map<Allele, List<Haplotype>> createAlleleMapper(final Map<VariantContext, Allele> mergeMap, final Map<Event, List<Haplotype>> eventMap ) {
-        return mergeMap.entrySet().stream().collect(Collectors.toMap(e -> e.getValue(), e -> eventMap.get(new Event(e.getKey())), (x,y)->x, LinkedHashMap<Allele, List<Haplotype>>::new));
+
+    protected static Map<Allele, List<Haplotype>> createAlleleMapper(List<VariantContext> eventsAtThisLoc, Map<Event, List<Haplotype>> eventMapper, VariantContext mergedVC) {
```
davidbenjamin: My rule of thumb is use the `static` keyword whenever allowed, because why not.  I checked a few stackexchange threads, such as [this one](http://stackoverflow.com/questions/538870/should-private-helper-methods-be-static-if-they-can-be-static) and on the whole people agree.  As you said, `static` *does* signify that the code can be called from outside the class, but that's not necessarily a bad thing.  For example, it would make things much easier if anyone ever wanted to share the code with a new class or to extract it to a utility class.  For now, though, the `protected` modifier tells us that it's not currently used outside of this class.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
davidbenjamin: Good catch!  Fixed it.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -561,7 +561,7 @@ private VariantContext addNonRefSymbolicAllele(final VariantContext mergedVC) {
 
         if( activeAllelesToGenotype.isEmpty() ) {
             haplotypes.stream().map(h -> h.getEventMap().get(loc))
-                    .filter(vc -> vc != null && !containsVCWithMatchingAlleles(eventsAtThisLoc, vc))
+                    .filter(vc -> vc != null && eventsAtThisLoc.stream().noneMatch(vc::hasSameAllelesAs))
```
davidbenjamin: Good catch.  Fixed.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -672,24 +666,6 @@ protected static void cleanUpSymbolicUnassembledEvents( final List<Haplotype> ha
         return new EventMap(haplotype, ref, refLoc, sourceNameToAdd);
     }
 
-    protected static final class Event {
```
davidbenjamin: When I figured out that `Event` was just a wrapper for `VariantContext` with a different `hashCode` I wanted to bang my forehead against a wall.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerUtils.java
davidbenjamin: It's not standard as far as I know, and I like the name you suggested.  Fixed.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -218,6 +210,16 @@ public CalledHaplotypes assignGenotypeLikelihoods(final List<Haplotype> haplotyp
...
     }
 
+    protected VariantContext makeAnnotatedCall(byte[] ref, SimpleInterval refLoc, FeatureContext tracker, SAMFileHeader header, VariantContext mergedVC, ReadLikelihoods<Allele> readAlleleLikelihoods, VariantContext call) {
+        final SimpleInterval locus = new SimpleInterval(mergedVC.getContig(), mergedVC.getStart(), mergedVC.getEnd());
+        final SimpleInterval refLocInterval= new SimpleInterval(refLoc);
+        final ReferenceDataSource refData = new ReferenceMemorySource(new ReferenceBases(ref, refLocInterval), header.getSequenceDictionary());
+        final ReferenceContext referenceContext = new ReferenceContext(refData, locus, refLocInterval);
+        final VariantContext untrimmedResult =  annotationEngine.annotateContext(call, tracker, referenceContext, readAlleleLikelihoods.toPerReadAlleleLikelihoodMap(), a -> true);
+        return call.getAlleles().size() == mergedVC.getAlleles().size() ? untrimmedResult
+                : GATKVariantContextUtils.reverseTrimAlleles(untrimmedResult);
```
davidbenjamin: I don't know.  This is copied from existing behavior.  I promised David Roazen not to affect HaplotypeCaller in any way.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -683,4 +673,10 @@ public PloidyModel getPloidyModel() {
     public IndependentSampleGenotypesModel getGenotypingModel() {
         return genotypingModel;
     }
+
+    // check whether all alleles of a vc, including the ref but excluding the NON_REF allele, are one base in length
+    protected static GenotypeLikelihoodsCalculationModel getGLModel(final VariantContext vc) {
+        final boolean isSNP = vc.getAlleles().stream().filter(a -> !a.isSymbolic()).allMatch(a -> a.length() == 1);
```
davidbenjamin: Yes.  `VariantContext::isSNP` only returns `true` when every alt allele is a SNP, and therefore returns false when there is a NON_REF symbolic allele.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/SomaticGenotypingEngine.java
```diff
@@ -297,6 +262,41 @@ public CalledHaplotypes callMutations (
         return new CalledHaplotypes(outputCalls, calledHaplotypes);
     }
 
+    private void addStrandBiasAnnotationsAndFilter(Map<String, Integer> originalNormalReadQualities, VariantContext mergedVC, PerReadAlleleLikelihoodMap tumorPRALM, PerAlleleCollection<Double> altAlleleFractions, Allele alleleWithHighestTumorLOD, VariantContextBuilder callVcb) {
```
davidbenjamin: I'm uncomfortable with the number of arguments some of these extracted methods have -- it's a code smell, for sure -- but it's a start.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
davidbenjamin: Good point.  In fact this code is entirely pointless.  In `AssemblyBasedCallerUtils::createdLikelihoodCalculationEngine` we have this line: `final double log10GlobalReadMismappingRate = likelihoodArgs.phredScaledGlobalReadMismappingRate < 0 ? - Double.MAX_VALUE`, so setting to `-1` if it was already negative makes no difference.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
```diff
@@ -415,29 +289,24 @@ private VariantAnnotatorEngine initializeVCFOutput() {
...
+        writeBamOutput(assemblyResult, readLikelihoods, calledHaplotypes);
+
+        if( MTAC.debug) { logger.info("----------------------------------------------------------------------------------"); }
+        return applyAnnotationsAndFilters(calledHaplotypes, featureContext);
+    }
+
+    private void writeBamOutput(AssemblyResultSet assemblyResult, ReadLikelihoods<Haplotype> readLikelihoods, HaplotypeCallerGenotypingEngine.CalledHaplotypes calledHaplotypes) {
         if ( MTAC.bamOutputPath != null ) {
             final Set<Haplotype> calledHaplotypeSet = new HashSet<>(calledHaplotypes.getCalledHaplotypes());
             if (MTAC.disableOptimizations)
```
davidbenjamin: Fixed.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
davidbenjamin: Fixed.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
```diff
@@ -729,63 +552,60 @@ private boolean isReadFromNormal(final GATKRead rec) {
...
         return annotatedCalls;
     }
 
-    //TODO: See if we can share code with HaplotypeCaller's isActive
-    private class Mutect2AssemblyRegionEvaluator implements  AssemblyRegionEvaluator {
-        public ActivityProfileState isActive(final AlignmentContext context, final ReferenceContext ref, final FeatureContext featureContext) {
-            if( context == null || context.getBasePileup().isEmpty() ) {
-                return new ActivityProfileState(ref.getInterval(), 0.0);
-            }
+    public ActivityProfileState isActive(final AlignmentContext context, final ReferenceContext ref, final FeatureContext featureContext) {
```
davidbenjamin: Yes we should.  Fixed in both classes.  There is never a reason to omit this annotation.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerUtils.java
davidbenjamin: Done

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
davidbenjamin: I really wanted to stream this but both the total and alt counts require filtering by base quality and `isDeletion`.  So we could either have two streams with duplicate filtering (inelegant), or `collect` a common filtered stream to an intermediate list, which we would then stream again (also inelegant).  I could also imagine something with `MutableInt`s but I couldn't think of anything that beat the original `for` loop.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/SomaticGenotypingEngine.java
```diff
@@ -119,22 +119,7 @@ public CalledHaplotypes callMutations (
...
-            // TODO: this varaible needs a descriptive name
-            final Map<VariantContext, Allele> mergeMap = new LinkedHashMap<>();
-
-            mergeMap.put(null, mergedVC.getReference()); // the reference event (null) --> the reference allele
-            for(int i = 0; i < eventsAtThisLoc.size(); i++) {
-                // TODO: as noted below, this operation seems dangerous. Understand how things can go wrong.
-                mergeMap.put(eventsAtThisLoc.get(i), mergedVC.getAlternateAllele(i)); // BUGBUG: This is assuming that the order of alleles is the same as the priority list given to simpleMerge function
-            }
-
-            /** TODO: the code in the for loop up to here needs refactor. The goal, as far as I can tell, is to create two things: alleleMapper and mergedVC
```
davidbenjamin: The code used to require two `Map`s:
```java
final Map<Event, List<Haplotype>> eventMapper = createEventMapper(loc. . .
yada yada yada
final Map<VariantContext, Allele> mergeMap = new LinkedHashMap<>();
```
which were then joined to create a `Map<Allele, List<Haplotype>>`.  These `Map`s no longer exist.  I feel like the code is still not transparent enough, however.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
davidbenjamin: I put back a `TODO` more explicit than the original.  There is more refactoring to be done but it's not obvious how to do it.

davidbenjamin:
@takutosato I addressed all of your comments.  Let's talk soon about the code in `createAlleleMapper`.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
takutosato: Got it. I always wondered why `private static` methods were ever helpful but now I understand.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
takutosato: I see. I was confused by `FeatureContext` vs `FeatureInput` but now it makes sense. Thanks!

takutosato: Found a typo (not yours) in `HaplotypeCallerEngine:assembleReads`. `giveAlleles` -> `givenAlleles`

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
```diff
@@ -530,9 +528,13 @@ private VariantContext addNonRefSymbolicAllele(final VariantContext mergedVC) {
         final List<VariantContext> eventsAtThisLoc = new ArrayList<>();
 
         if( activeAllelesToGenotype.isEmpty() ) {
-            haplotypes.stream().map(h -> h.getEventMap().get(loc))
-                    .filter(vc -> vc != null && eventsAtThisLoc.stream().noneMatch(vc::hasSameAllelesAs))
-                    .forEach(eventsAtThisLoc::add);
+            final List<VariantContext> events = haplotypes.stream().map(h -> h.getEventMap().get(loc))
+                    .filter(Objects::nonNull).collect(Collectors.toList());
```
takutosato: I have a hunch that there is a more direct/concise way create a set of unique events at this locus. I just spent about 20 minutes and failed to come up with a solution. Will try again later.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
takutosato: Got it

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
takutosato: Good call.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
takutosato: I missed filtering by base quality for `total count`. I agree with you, for loop is better.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
takutosato: Looks like removing this line is not in the latest commit.

takutosato:
@davidbenjamin back to you. Should be good to go after removing that one line about  `phredScaledGlobalReadMismappingRate`.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
davidbenjamin: Are you sure?  I can't find any mention of "phred" in Mutect2Engine.java in the latest commit.

davidbenjamin:
@takutosato I *think* that one line is gone. . . if so, back to you for final review.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
takutosato: you are right - sorry about that.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerGenotypingEngine.java
takutosato: Spent another 20 minutes and couldn't come up with an alternative solution. Moving on.

takutosato:
@davidbenjamin merge time! 👍

PR #714, opened 2016-09-21
Fixing minor issue in docker file where cache could be erroneous.

### LINKED ISSUE #713: Clean up Dockerfile apt-get commands.
https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/

Note that we need to combine some apt-get install and apt-get update commands.

LeeTL1220:
Closes #713


samuelklee:
looks good to me, merge once tests pass!

PR #722, opened 2016-09-27
Fixes an issue of GetBayesianHetCoverage with non- ACTG (e.g. N, X, etc) reference bases

mbabadi:
... by filtering out non-ACTG bases when (1) walking the reference and (2) making a pileup.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/pulldown/BayesianHetPulldownCalculator.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/pulldown/BayesianHetPulldownCalculator.java
@@ -184,6 +184,20 @@ public final class BayesianHetPulldownCalculator {
     }
 
     /**
+     * Checks of a given base is in BASES
+     * @param base a nucleotide
+     * @return boolean
+     */
+    private boolean isProperBase(final Nucleotide base) {
+        for (final Nucleotide properBase : BASES) {
+            if (base.equals(properBase)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /**
      * Returns map of base-pair to error probabilities at a given locus. All reads are considered (not just ACTG)
      * @param locus locus
      * @return map of base-pair to error probabilities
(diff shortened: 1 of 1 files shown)


mbabadi:
@LeeTL1220 I disabled indexing in SamLocusIterator by default. In GetHetCoverage, it is pretty much always disabled (it is enabled only if #SNPS < 25000). There must a bug in SamLocusIterator since indexing leads to a neverending iterator sometimes. I found another example. Could you check if this PR fixes your problematic case as well?

LeeTL1220:
@mbabadi ok... I'll get to it asap....

LeeTL1220:
@mbabadi Hey, as a reminder, I have this bug for both GetBayesian and regular GetHetCoverage...

LeeTL1220:
@mbabadi In other words, I am not going to test this, since you made GetBayesian more like GetHet

src/main/java/org/broadinstitute/hellbender/tools/exome/pulldown/BayesianHetPulldownCalculator.java
```diff
@@ -184,6 +184,20 @@ private double getSingletonLogLikelihood(final Map<Nucleotide, List<BaseQuality>
...
+        for (final Nucleotide properBase : BASES) {
+            if (base.equals(properBase)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /**
      * Returns map of base-pair to error probabilities at a given locus. All reads are considered (not just ACTG)
```
LeeTL1220: Is this comment still applicable?

samuelklee:
Not sure why this was assigned to me---should I review as well?

LeeTL1220:
@mbabadi Not much.... but give it a look.  Feel free to merge after that.

@samuelklee whoops!

src/main/java/org/broadinstitute/hellbender/tools/exome/pulldown/BayesianHetPulldownCalculator.java
mbabadi: It is still applicable. The improper base calls, however, will be ignored in the inference.

PR #754, opened 2016-10-21
updating gradle to 3.1

lbergelson:
this allows use of gradle composite builds which should help reduce pain of developing gatk/gatk-protected together


lbergelson:
@LeeTL1220 Coveralls is incorrectly marking these builds as failed.  Could you take a look at this when you get a chance?

PR #756, opened 2016-10-26
Fix regression in HaplotypeCaller with shard boundary check

droazen:
Somehow during refactoring, the line in the HaplotypeCaller that checks
whether an event starts in the current shard got changed to instead check
that the event is completely contained in the current shard. This means that
events spanning shard boundaries would be dropped completely.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCaller.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCaller.java
@@ -14,6 +14,7 @@ import org.broadinstitute.hellbender.engine.*;
 import org.broadinstitute.hellbender.engine.filters.ReadFilter;
 import org.broadinstitute.hellbender.exceptions.UserException;
 import org.broadinstitute.hellbender.utils.fasta.CachingIndexedFastaSequenceFile;
+import org.broadinstitute.hellbender.utils.SimpleInterval;
 
 import java.io.File;
 import java.io.FileNotFoundException;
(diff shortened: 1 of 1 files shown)


droazen:
@davidbenjamin please review. If the change was not intentional and the fix looks ok to you, I'll try to devise a unit test that fails before the fix and passes after it.

Note that this may be applicable to Mutect as well (we should check).

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCaller.java
```diff
@@ -212,7 +213,7 @@ public void apply(final AssemblyRegion region, final ReferenceContext referenceC
         hcEngine.callRegion(region, featureContext).stream()
                 // Only include calls that start within the current read shard (as opposed to the padded regions around it).
                 // This is critical to avoid duplicating events that span shard boundaries!
-                .filter(call -> getCurrentReadShardBounds().contains(call))
+                .filter(call -> getCurrentReadShardBounds().contains(new SimpleInterval(call.getContig(), call.getStart(), call.getStart())))
```
davidbenjamin: If I understand the change correctly, this fixes previous behavior where large insertions starting within one read shard and ending in another would be filtered, and replaces it with what the comment says to do.  Seems right to me.

davidbenjamin:
@droazen PS This fix definitely applies to Mutect as well.  Can you make the same change there before merging?

droazen:
Sure, before merging I'll extend the fix to Mutect, and craft a unit test to ensure this stays fixed.

lbergelson:
opened #1081 to remember this in mutect

PR #759, opened 2016-10-31
Bring back standard read filters to Mutect2

takutosato:
We apply these filters in GATK3. Also moved `HaplotyeeCallerEngine.makeStandardHCReadFilters` to AssemblyBasedCallerUtils.java.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerUtils.java
@@ -7,10 +7,6 @@ import htsjdk.samtools.util.Locatable;
 import htsjdk.variant.variantcontext.VariantContext;
 import org.apache.logging.log4j.Logger;
 import org.broadinstitute.hellbender.engine.AssemblyRegion;
-import org.broadinstitute.hellbender.engine.filters.MappingQualityReadFilter;
-import org.broadinstitute.hellbender.engine.filters.ReadFilter;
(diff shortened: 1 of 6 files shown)


takutosato:
@davidbenjamin could you review?

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/AssemblyBasedCallerUtils.java
```diff
@@ -269,4 +273,27 @@ public static AssemblyResultSet assembleReads(final AssemblyRegion region,
             throw e;
         }
     }
+
+    /**
+     * @return the default set of read filters for HC (includes the MQ threshold filter) or M2 (does not include the MQ filter)
+     */
+    public static List<ReadFilter> makeStandardReadFilterList(final boolean includeMQThresholdFilter) {
```
droazen: I'd rather not have the make read filters methods for Mutect and HaplotypeCaller entangled like this -- it is very problematic for adding additional filters in the future to one tool but not the other. Can you restore `HaplotyeeCallerEngine.makeStandardHCReadFilters()`, and give Mutect a completely separate method for creating its filters?

takutosato: @droazen will do. Thanks for reviewing the changes. @davidbenjamin please wait for my next commit.

takutosato:
@davidbenjamin this PR is ready for review again.

davidbenjamin:
@takutosato looks good.

PR #785, opened 2016-11-15
Mainly trivial code cleaning

davidbenjamin:
in particular, using `Utils.validateArg()` and `Utils.nonNull` wherever possible.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/ReadCountCollection.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/ReadCountCollection.java
@@ -92,19 +92,12 @@ public final class ReadCountCollection implements Serializable {
             Utils.nonNull(targets,"the input targets cannot be null");
             Utils.nonNull(columnNames,"the column names cannot be null");
             Utils.nonNull(counts,"the counts cannot be null");
-            if (columnNames.contains(null)) {
-                throw new IllegalArgumentException("column names contain nulls");
-            } else if (targets.contains(null)) {
-                throw new IllegalArgumentException("there are some null targets");
-            } else if (counts.getRowDimension() != targets.size()) {
-                throw new IllegalArgumentException("number of count rows does not match the number of targets");
-            } else if (counts.getColumnDimension() != columnNames.size()) {
-                throw new IllegalArgumentException("number of count columns does not match the number of column names");
-            } else if (new HashSet<>(targets).size() != targets.size()) {
-                throw new IllegalArgumentException("targets contain duplicates");
-            } else if (new HashSet<>(columnNames).size() != columnNames.size()) {
-                throw new IllegalArgumentException("column names contain duplicates");
-            }
+            Utils.containsNoNull(columnNames, "column names contain nulls");
+            Utils.containsNoNull(targets, "there are some null targets");
+            Utils.validateArg(counts.getRowDimension() == targets.size(), "number of count rows does not match the number of targets");
+            Utils.validateArg(counts.getColumnDimension() == columnNames.size(), "number of count columns does not match the number of column names");
+            Utils.validateArg(new HashSet<>(targets).size() == targets.size(), "targets contain duplicates");
+            Utils.validateArg(new HashSet<>(columnNames).size() == columnNames.size(), "column names contain duplicates");
             this.targets = Collections.unmodifiableList(new ArrayList<>(targets));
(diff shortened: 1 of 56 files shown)


davidbenjamin:
@takutosato can you review, but if you feel like the third commit will be more than a very brief effort just do the first two commits and shelve the third for later?

src/test/java/org/broadinstitute/hellbender/utils/hmm/HeavyStateTestHMModel.java
```diff
@@ -18,13 +21,9 @@
...
-        if (numStates <= 1) {
-            throw new IllegalArgumentException("bad number of states");
-        } else if (heavyState < 0 || heavyState >= numStates) {
-            throw new IllegalArgumentException("bad heavy state");
-        } else if (heavyStateWeight <= (1.0 / numStates) || heavyStateWeight >= 1.0) {
-            throw new IllegalArgumentException("bad heavy state weight");
-        }
+        Utils.validateArg(numStates > 1, "bad number of states");
+        Utils.validIndex(heavyState, numStates);
+        ParamUtils.inRange(heavyStateWeight, (1.0 / numStates), 1.0, "bad heavy state weight");
```
takutosato: `ParamUtils.inRange` accepts a closed interval, whereas the original code accepted an open interval. Don't know what `heavyStateWeight` represents, but if it's equal to 1.0, I think the behavior changes; old code throws an exception, new code doesn't. 

That said, `heavyStateWeight` is a double so == comparison is not exact so maybe that doesn't actually make a difference. Either way I'd like to hear your thoughts on this

takutosato:
@davidbenjamin back to you

src/main/java/org/broadinstitute/hellbender/tools/exome/ReadCountCollection.java
```diff
@@ -92,19 +92,12 @@ private ReadCountCollection(final List<Target> targets, final List<String> colum
...
-                throw new IllegalArgumentException("targets contain duplicates");
-            } else if (new HashSet<>(columnNames).size() != columnNames.size()) {
-                throw new IllegalArgumentException("column names contain duplicates");
-            }
+            Utils.containsNoNull(columnNames, "column names contain nulls");
+            Utils.containsNoNull(targets, "there are some null targets");
+            Utils.validateArg(counts.getRowDimension() == targets.size(), "number of count rows does not match the number of targets");
+            Utils.validateArg(counts.getColumnDimension() == columnNames.size(), "number of count columns does not match the number of column names");
+            Utils.validateArg(new HashSet<>(targets).size() == targets.size(), "targets contain duplicates");
+            Utils.validateArg(columnNames.size() == columnNames.size(), "column names contain duplicates");
```
davidbenjamin: good catch

src/main/java/org/broadinstitute/hellbender/utils/pca/PCA.java
```diff
@@ -150,14 +150,12 @@ public int numComponentsToAccountForVariance(final double proportion) {
...
             return null;
-        } else if (input.stream().anyMatch(Objects::isNull)) {
-            throw new IllegalArgumentException(String.format("the input %s list must not contain nulls", roleName));
-        } else if (input.stream().sorted().distinct().count() != input.size()) {
-            throw new IllegalArgumentException(String.format("the input %s list must not contain repeats", roleName));
-        } else {
-            return Collections.unmodifiableList(new ArrayList<>(input));
         }
-
+        Utils.validateArg(input.stream().noneMatch(Objects::isNull), () ->
```
davidbenjamin: true

src/test/java/org/broadinstitute/hellbender/utils/hmm/HeavyStateTestHMModel.java
davidbenjamin: It's fine not to `throw` when ` heavyStateWeight == 1.0`.  The idea here is to have an HMM biased toward transitions to a particular "heavy" state.  /since the expected behavior is obvious, it's useful for testing.  If the weight is 1.0, it's complete bias -- the HMM will never go to another state, which is a valid extreme case.

davidbenjamin:
Back to @takutosato.  I fixed everything you caught.

takutosato:
@davidbenjamin looks good. please merge.

PR #790, opened 2016-11-15
Do not initialize pair HMM when we have zero reads

takutosato:


takutosato:
@davidbenjamin please review

davidbenjamin:
Great job tracking down the source of this error @takutosato!

takutosato:
thanks!

PR #803, opened 2016-11-29
Cleaning code

davidbenjamin:


davidbenjamin:
@takutosato This addresses your comments on a recent PR, then does a few more things to clean M2, then binges on IntelliJ's automatic code cleanups.  You can skim the last several commits if you trust the IDE and the Travis tests.

davidbenjamin:
Also, this is low priority -- much lower than the M2 integration tests PR.

davidbenjamin:
Some notes on the commits:

* got rid of M2 debug -- earlier we discussed keeping this because conditional breakpoints are slow in the IDE, but personally I would rather just write an one-off `if` statement and put a breakpoint inside rather permanently clutter the code. 

* inlined redundant variables -- replaced the pattern `result = something; return result` with simply `return something` in several places

* remove unnecessary throws clauses -- deleted `throws Exception` from methods that do not in fact throw these exceptions.

* remove redundant super -- no need to say `super.method()` when you can just say `method()`.  Otherwise, what's the point of inheritance?

* remove redundant this -- it's not Python.

* replace fully qualified names with imports -- eg `java.lang.Util.List` --> `List`.

PR #810, opened 2016-12-02
Simplify Mutect2 read filtering

davidbenjamin:
Closes issue #672

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
@@ -13,6 +13,7 @@ import org.apache.commons.lang3.math.NumberUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.broadinstitute.hellbender.engine.*;
+import org.broadinstitute.hellbender.engine.filters.MappingQualityReadFilter;
 import org.broadinstitute.hellbender.engine.filters.ReadFilter;
 import org.broadinstitute.hellbender.engine.filters.ReadFilterLibrary;
 import org.broadinstitute.hellbender.engine.filters.WellformedReadFilter;
@@ -38,6 +39,7 @@ import org.broadinstitute.hellbender.utils.variant.GATKVCFConstants;
 import org.broadinstitute.hellbender.utils.variant.GATKVCFHeaderLines;
 
 import java.util.*;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
@@ -66,6 +68,7 @@ public final class Mutect2Engine implements AssemblyRegionEvaluator {
     private SAMFileHeader header;
 
     private static final int MIN_READ_LENGTH = 30;
+    private static final int READ_QUALITY_FILTER_THRESHOLD = 20;
 
     private SampleList samplesList;
     private boolean printTCGAsampleHeader = false;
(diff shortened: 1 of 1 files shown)


davidbenjamin:
@takutosato Not very high priority, but should be quick.  HaplotypeCaller is unaffected, and now Mutect does the same thing as HaplotypeCaller.  Tumor and normal filtering is now the same.  The old idea was to make normal filtering more permissive to retain as much evidence of variation in the normal as possible, but when you think about it we're looking for a variant genotype with allele fraction 1/2 or 1 in the normal, so a few stray alt reads aren't going to matter.  And even if they did, they would *not* be evidence of the variant existing in the normal.  They might be evidence that the *site* is noisy, but that's a separate issue.

droazen:
@davidbenjamin This exact change was PR'd a few months ago, and I requested in the review that the filtering for HC and Mutect be kept separate, so that we can easily adjust them independently, rather than calling into a shared method.

davidbenjamin:
@droazen I would be surprised if we ever needed different read filtering in M2, but we're still experimenting and I see your point.  I will re-separate the methods.  @takutosato there will still be some non-trivial changes here because I'm making the M2 filtering simpler.

davidbenjamin:
Okay, tests are passing again, not sure what Travis was up to.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
```diff
@@ -454,26 +452,18 @@ private boolean isNonRef(final byte refBase, final PileupElement p) {
...
-            if (isReadFromNormal(rec)) {
-                if( rec.getLength() < MIN_READ_LENGTH ) {
-                    readsToRemove.add(rec);
-                }
-            } else if( rec.getLength() < MIN_READ_LENGTH || rec.getMappingQuality() < MTAC.MIN_MAPPING_QUALITY_SCORE || hasBadMate(rec) ||
-                    (MTAC.keepRG != null && !rec.getReadGroup().equals(MTAC.keepRG)) ) {
-                readsToRemove.add(rec);
-            }
-        }
+    private Set<GATKRead> filterNonPassingReads( final AssemblyRegion assemblyRegion) {
```
takutosato: also, rename the method to `filterReads`

takutosato:
@davidbenjamin I just pushed my commit. Here are the highlights:

* `hasBadMate()` and `ReadFilterLibrary.MATE_ON_SAME_CONTIG_OR_NO_MAPPED_MATE` are equivalent
* call `regionForGenotyping.removeAll(readsToRemove)` outside of the method to make the side-effect conspicuous
* compose a Predicate from smaller Predicates. I thought it'd be overkill to create a subclass of `ReadFilter` for each criterion.

What do you think?

davidbenjamin:
These are great changes!  I will deal with your previous comments in a new commit on top of yours and submit for re-review.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
```diff
@@ -153,6 +150,7 @@ private void initialize() {
         // In Mutect2 we intentionally keep poorly mapped reads for reassembly
```
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
davidbenjamin: done by you

davidbenjamin:
@takutosato back to you.  I loved the `Predicate` idea and realized that the composed `Predicate` could be a class-level field.  Then the method to identify reads to remove became a one-liner, which I inlined.

takutosato:
awesome! the code looks good. please merge!

PR #815, opened 2016-12-06
Delete classes used only in our obsolete XHMM port

davidbenjamin:
@asmirnov239 @samuelklee here you go.   Letting @mbabadi know about these changes just in case the new coverage model needs any of these old classes, which I seriously doubt.


asmirnov239:
@davidbenjamin Sorry for taking so long, was deciding whether I want we want to keep some classes for future use. Everything looks good, go ahead and merge!

Issue #733, opened 2016-10-06
Emit all alleles in Mutect


davidbenjamin:
We should emit _all_ alleles with their posterior probabilities or somatic qual scores or whatever.  This will make it easier to unite all filtering in a second post-processing step.

davidbenjamin:
Shortly after posting this issue, Sarah Calvo pointed out to us that this is an outright bug as far as her work on somatic variants in mitochondria is concerned.

chandrans:
Hey David. What do you mean by "all alleles" You mean like any potential allele that is at the position? Like is a GVCF? Thanks!

davidbenjamin:
Hi Sheila, we weren't thinking as permissive as a gvcf, more like a vcf, i.e. we will emit all alleles that we believe actually exist as somatic mutations.  This will only affect a small number of sites.

chandrans:
Okay got it. Thanks. Hopefully users will be happy with this! We have had some complaints about it not being permissive enough especially at messy sites.

davidbenjamin:
@chandrans It's good to hear that (the complaints, that is).  Are these complaints on the forum?

chandrans:
@davidbenjamin Yes, from the forum. I have been emailing Takuto with some interesting MuTect2 issues/comments from users. Would you like to be in on these threads?

davidbenjamin:
@chandrans yes please!

chandrans:
Awesome. Thanks David. Will do.

davidbenjamin:
Closed by #816.

Issue #807, opened 2016-12-02
Make sure we can make a Mutect2 PoN in GATK4


davidbenjamin:
Did this get lost or break in the port to GATK4?  We need to run GATK4 M2 in create-PoN mode, then run M2 with this PoN to see if this works.

davidbenjamin:
Closed by #816.

Issue #735, opened 2016-10-06
Put Mutect filtration into a second pass


davidbenjamin:
`HaplotypeCaller` and `VQSR` are a two-pass team in which first we call variants permissively and then filter.  There are several advantages of this approach:
- Rather than scattering filtration all over the variant calling code, we filter in a single place, i.e. a vcf record with all of its annotations.
- It lets users go to town with manual review if they so desire.
- While calculating likelihoods is fundamentally an online method, suitable for a walker, filtering is suited to a machine learning approach that learns about artifacts from the entire call set.
- The two-pass approach is modular and extensible because we can add filters without affecting the variant calling code.
- The two-pass approach will let us use model based recalibration instead of hard filters.

Note that this can be done initially with no change in behavior.

davidbenjamin:
Closed by #816.

PR #816, opened 2016-12-08
Refactoring to prepare Mutect2 for model improvements

davidbenjamin:
* Closes issues #733, #735, and #807.
* Gets rid of a lot of arbitrary filters that have probably become obsolete
* moves filtering into a second pass
* deals with a lot of little TODOs
* emits all somatic alleles
* annotates the output vcf more than before

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2FilteringEngine.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2FilteringEngine.java
@@ -1,22 +1,24 @@
 package org.broadinstitute.hellbender.tools.walkers.mutect;
 
 import htsjdk.variant.variantcontext.VariantContext;
-import org.broadinstitute.hellbender.engine.FeatureContext;
 import org.broadinstitute.hellbender.utils.variant.GATKVCFConstants;
 
-import java.util.Collection;
+import java.util.*;
 
 /**
  * Created by davidben on 9/15/16.
  */
 public class Mutect2FilteringEngine {
-    final M2ArgumentCollection MTAC;
 
-    public Mutect2FilteringEngine(final M2ArgumentCollection MTAC) {
-        this.MTAC = MTAC;
+    private Mutect2FilteringEngine() { }
+
+    private static void applyTriallelicFilter(final VariantContext vc, final Collection<String> filters) {
+        if (vc.getNAlleles() > 2) {
+            filters.add(GATKVCFConstants.TRIALLELIC_SITE_FILTER_NAME);
+        }
     }
 
-    public void applySTRFilter(final VariantContext vc, final Collection<String> filters) {
+    private static void applySTRFilter(final VariantContext vc, final Collection<String> filters) {
         // STR contractions, such as ACTACTACT -> ACTACT, are overwhelmingly false positives so we hard filter by default
         if (vc.isIndel()) {
             final int[] rpa = vc.getAttributeAsList(GATKVCFConstants.REPEATS_PER_ALLELE_KEY).stream()
@@ -33,7 +35,7 @@ public class Mutect2FilteringEngine {
         }
     }
 
-    public void applyClusteredReadPositionFilter(final VariantContext vc, final Collection<String> filters) {
+    private static void applyClusteredReadPositionFilter(final M2ArgumentCollection MTAC, final VariantContext vc, final Collection<String> filters) {
         if (MTAC.ENABLE_CLUSTERED_READ_POSITION_FILTER) {
             final Double tumorFwdPosMedian = (Double) vc.getAttribute(GATKVCFConstants.MEDIAN_LEFT_OFFSET_KEY);
             final Double tumorRevPosMedian = (Double) vc.getAttribute(GATKVCFConstants.MEDIAN_RIGHT_OFFSET_KEY);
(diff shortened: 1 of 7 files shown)


davidbenjamin:
@takutosato this one's hefty.  By putting all the unfiltered calls in memory at once we will be able to filter calls more intelligently.  

I got rid of several filters without thinking too hard about them.  These are perturbations to the model to improve it slightly.  When we change the model there is no reason to think that the same perturbations will be relevant.  Therefore, any filter that is not manifestly sensible is not worth keeping.  It is quite possible that this will temporarily hurt performance.  However, the price of keeping these ad hoc parts of the model is that every change to the code interacts with these mystery bits unpredictably, slowing down development.  This is unacceptable, hence we must delete ruthlessly.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/SomaticGenotypingEngine.java
```diff
@@ -184,6 +167,20 @@ public CalledHaplotypes callMutations(
         return new CalledHaplotypes(outputCalls, calledHaplotypes);
     }
 
+    // compute the likelihoods of: AA, AB, AC. . . where A is ref and B, C. . . are somatic alts
+    private PerAlleleCollection<Double> getHetGenotypeLogOdds(ReadLikelihoods<Allele> likelihoods,
+                                                              final boolean isNormal,
+                                                              final Strand strand) {
+        final String sample = isNormal ? matchedNormalSampleName : tumorSampleName;
+        final OptionalDouble givenAltAlleleFraction = isNormal ? OptionalDouble.of(0.5) : NO_FIXED_TUMOR_ALT_FRACTION;
```
takutosato: Change `OptionalDouble.of(0.5)` to `GERMLINE_HET_ALT_FRACTION` (currently unused) or change `NO_FIXED_TUMOR_ALT_FRACTION` to `OptionalDouble.empty().` I prefer the former.

src/test/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2IntegrationTest.java
```diff
@@ -63,6 +66,49 @@ public void testDreamTumorNormal(final File tumorBam, final String tumorSample,
...
+                "-I", normalBam.getAbsolutePath(),
+                "-normal", normalSample,
+                "-R", b37_reference_20_21,
+                "-L", "20",
+                "-O", ponVcf.getAbsolutePath(),
+                "-artifact_detection_mode"
+        };
+
+        runCommandLine(createPonArgs);
+        new Main().instanceMain(makeCommandLineArgs(Arrays.asList("-F", ponVcf.getAbsolutePath()), "IndexFeatureFile"));
```
takutosato: As discussed, please describe that this command creates an index file (or does it make the index file visible?) in a comment

src/test/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2IntegrationTest.java
```diff
@@ -63,6 +66,49 @@ public void testDreamTumorNormal(final File tumorBam, final String tumorSample,
...
+                "-L", "20",
+                "-O", outputVcf.getAbsolutePath()
+
+        };
+
+        runCommandLine(callWithPonArgs);
+
+        final Set<String> outputKeys = StreamSupport.stream(new FeatureDataSource<VariantContext>(outputVcf).spliterator(), false)
+                .filter(vc -> vc.getFilters().isEmpty())
+                .map(vc -> keyForVariant(vc)).collect(Collectors.toSet());
```
takutosato: since we expect outputKeys to be empty we could just `count()` after `.filter()` and check that it equals 0

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2FilteringEngine.java
```diff
@@ -42,7 +35,7 @@ public void applySTRFilter(final VariantContext vc, final Collection<String> fil
         }
     }
 
-    public void applyClusteredReadPositionFilter(final VariantContext vc, final Collection<String> filters) {
+    public static void applyClusteredReadPositionFilter(final M2ArgumentCollection MTAC, final VariantContext vc, final Collection<String> filters) {
```
takutosato: shouldn't all `apply*Filter()` methods be private?

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2FilteringEngine.java
```diff
@@ -55,24 +48,20 @@ public void applyClusteredReadPositionFilter(final VariantContext vc, final Coll
...
             filters.add(GATKVCFConstants.PON_FILTER_NAME);
         }
     }
 
-    public void applyGermlineVariantFilter(final VariantContext vc, final Collection<String> filters, final FeatureContext featureContext) {
+    public static void applyGermlineVariantFilter(final M2ArgumentCollection MTAC, final VariantContext vc, final Collection<String> filters) {
         if (!vc.hasAttribute(GATKVCFConstants.NORMAL_LOD_KEY)) {
             return;
         }
         final int loc = vc.getStart();
```
takutosato: intelliJ says `loc` is not used

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2FilteringEngine.java
```diff
@@ -97,12 +86,25 @@ public void applyStrandBiasFilter(final VariantContext vc, final Collection<Stri
...
 
-    public void applyEventDistanceFilters(final VariantContext vc, final Collection<String> filters) {
+    public static void applyEventDistanceFilters(final VariantContext vc, final Collection<String> filters) {
         final Integer eventCount = vc.getAttributeAsInt(GATKVCFConstants.EVENT_COUNT_IN_HAPLOTYPE_KEY, -1);
         if (eventCount >= 3) {
             filters.add(GATKVCFConstants.HOMOLOGOUS_MAPPING_EVENT_FILTER_NAME);
         }
     }
 
+    public static Set<String> calculateFilters(final M2ArgumentCollection MTAC, final VariantContext vc) {
```
takutosato: ideally it'd be cleaner if we can just concatenate the result of each filter in `filters`, instead of passing `filters` to each method and let it modify `filters` as a side effect. If you could put a TODO here I can give it a try

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/SomaticGenotypingEngine.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/SomaticGenotypingEngine.java
```diff
@@ -215,18 +219,18 @@ private VariantContext addGenotypes(boolean hasNormal, VariantContext mergedVC,
         // if we are calling with a normal, build the genotype for the sample to appear in vcf
         if (hasNormal) {
             final PerAlleleCollection<Integer> normalAlleleDepths = getRefAltCount(mergedVC, likelihoods, matchedNormalSampleName, Strand.BOTH);
-            final int normalRefAlleleDepth = normalAlleleDepths.getRef();
-            final int normalAltAlleleDepth = normalAlleleDepths.getAlt(alleleWithHighestTumorLOD);
-            final double normalAlleleFraction = (double) normalAltAlleleDepth / ( normalRefAlleleDepth + normalAltAlleleDepth);
+            final int normalTotalDepth = likelihoods.sampleReadCount(likelihoods.indexOfSample(matchedNormalSampleName));
+            final double[] normalAlleleFractions = alleles.stream()
```
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2FilteringEngine.java
```diff
@@ -16,6 +16,12 @@ public Mutect2FilteringEngine(final M2ArgumentCollection MTAC) {
         this.MTAC = MTAC;
     }
 
+    public void applyTriallelicFilter(final VariantContext vc, final Collection<String> filters) {
+        if (vc.getNAlleles() > 2) {
```
davidbenjamin: Addressed in a later commit -- once this filter is reached only alleles passing TLOD and NLOD are in the vcf.  (I don't think of TLOD and NLOD as filtering, but rather as part of somatic genotyping).

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2FilteringEngine.java
```diff
@@ -73,14 +73,12 @@ public void applyGermlineVariantFilter(final VariantContext vc, final Collection
         final int loc = vc.getStart();
         final boolean siteInCosmic = !featureContext.getValues(MTAC.cosmicFeatureInput, loc).isEmpty();
         final boolean siteInDbsnp = !featureContext.getValues(MTAC.dbsnp.dbsnp, loc).isEmpty();
-        final double normalLodFilterThreshold = siteInDbsnp && !siteInCosmic ? MTAC.NORMAL_DBSNP_LOD_THRESHOLD : MTAC.NORMAL_LOD_THRESHOLD;
-
-        final double normalLod = vc.getAttributeAsDouble(GATKVCFConstants.NORMAL_LOD_KEY, 0.0);
-
-        if (normalLod < normalLodFilterThreshold) {
-            filters.add(GATKVCFConstants.GERMLINE_RISK_FILTER_NAME);
+        if (siteInDbsnp && !siteInCosmic ) {
```
davidbenjamin: We already check the normal lod as part of the first pass.  This filter checks for dbSNP and then applies a *stricter* lod threshold.  Note that this reproduces old behavior (I think) and I'm only keeping it because I know we're about to change things.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2FilteringEngine.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2FilteringEngine.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2FilteringEngine.java
davidbenjamin: I put in a `TODO` and would be happy if you cleaned it up.

src/test/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2IntegrationTest.java
davidbenjamin: done

src/test/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2IntegrationTest.java
davidbenjamin: done

davidbenjamin:
Back to @takutosato if tests pass.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2FilteringEngine.java
takutosato: Got it. Now I see that we get rid of bad alleles in `addGenotypes`.

takutosato:
Looks good!

PR #822, opened 2016-12-14
Small changes to Mutect2

takutosato:
We need changes in the first two commits to run Mutect2.


takutosato:
Hi @davidbenjamin, here are the few changes I mentioned today. Could you take a look?

PR #829, opened 2016-12-19
Mutect2 and HC always use NON_ZERO_REFERENCE_LENGTH_ALIGNMENT read filter

takutosato:
This filter takes place pre-reassembly and removes reads that do not take up any reference bases. Without the filter we get an error like this one: `java.lang.IllegalArgumentException: Invalid interval. Contig:3 start:100170602 end:100170601`


davidbenjamin:
Although, if @droazen approves, you can keep the HC change as well.

droazen:
Can you add a unit or integration test that fails before the change and passes after (for both tools)?

Merge after adding test :+1:

takutosato:
@davidbenjamin I added integration test sto HC and M2. Without the change (i.e. if you check out just the first commit), the tests should fail with `java.lang.IllegalArgumentException: Invalid interval. Contig:20 start:5624960 end:5624959`

I made the test an integration test, not a unit test as @droazen suggested. This was because I don't understand the read filter code enough to identify the appropriate unit.

davidbenjamin:
@takutosato You may squash and merge.  I think that an integration test is fine for our purposes.

PR #836, opened 2016-12-21
Extract common parent for M2 and HC GenotypingEngines.  Fixes #811.

### LINKED ISSUE #811: SomaticGenotypingEngine should not inherit from HaplotypeCallerGenotypingEngine
`SomaticGenotypingEngine` uses very few methods from `HaplotypeCallerGenotypingEngine`.  Their primary role is to assign genotypes and likelihoods, the code of which they don't share at all.  It would be much more natural to let both inherit from an abstract `AssemblyBasedCallerGenotypingEngine` or to extract a few common static methods into `AssemblyBasedCallerUtils`.  @droazen what would you recommend?

davidbenjamin:
@takutosato This PR just moves methods around a bit.  It fixes the aesthetically offensive inheritance of `SomaticGenotypingEngine` from `HaplotypeCallerGenotypingEngine`.


takutosato:
@davidbenjamin please rebase and merge

davidbenjamin:
Thanks @takutosato!  Since this is not critical for progress on M2, I will wait for approval from @droazen, *which can occur after the break*.

davidbenjamin:
@droazen what do you think?

Issue #834, opened 2016-12-21
Hardcoded paths, which do not exist outside of the Broad, should not be put into json 


LeeTL1220:
json accompanying WDL should be viewed as a template.

Both m2 json files contain hardcoded paths that users outside of the Broad (or on the cloud) cannot access.  

I propose using the convention in the CNV json...  Populate parameters that have default values (these would work outside the Broad) and parameters that are needed by the user use ``$__name_of_parameter__``.  This makes these parameters easy to replace in python, using templates (or in bash with sed, etc etc)

LeeTL1220:
Thoughts?  @samuelklee @davidbenjamin @takutosato @asmirnov239 

@vdauwera will need to have these template json files for support of best practice workflows.

LeeTL1220:
Tagging as MuTect2, since those WDLs are in question.

takutosato:
sounds good to me

vdauwera:
:+1:

samuelklee:
Sounds good to me as well.

LeeTL1220:
@takutosato We can put custom wdl in the ``unsupported`` directory.

LeeTL1220:
(or create a new sandbox repo)

takutosato:
I will put the broad in-house wdls in a palantir repo.

davidbenjamin:
Closed by @takutosato in PR #846.

PR #846, opened 2017-01-05
generalize mutect2 wdl and json for users external to the broad

takutosato:
@davidbenjamin I removed hardcoded file paths in the mutect2 wdl for users outside of the Broad as suggested by @LeeTL1220 in #834. Also, the file paths in .json were replaced with with `$__parameter_name__`.  Lastly, I removed files that are of interest only to members of DSDE-methods (e.g. `mutect2-dream-eval-sge.wdl`) from this repo. They will go to the palantir repo.


davidbenjamin:
Looks good!

PR #847, opened 2017-01-06
New tool that counts false positives in a vcf to be used for mutect2 NA12878 normal-normal validation

takutosato:
@davidbenjamin I wrote a simple tool for mutect2 NA12878 validation. It takes a list of vcfs, counts the number of SNP and INDEL variants (i.e. PASS variants) in each file, and prints the counts in a tsv format. Please take a look!

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/CountFalsePositives.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/CountFalsePositives.java
@@ -2,6 +2,7 @@ package org.broadinstitute.hellbender.tools.walkers.mutect;
 
 import htsjdk.variant.variantcontext.VariantContext;
 import org.apache.commons.io.FilenameUtils;
+import org.apache.commons.lang3.tuple.Pair;
 import org.broadinstitute.hellbender.cmdline.CommandLineProgram;
 import org.broadinstitute.barclay.argparser.Argument;
 import org.broadinstitute.barclay.argparser.CommandLineProgramProperties;
@@ -9,14 +10,16 @@ import org.broadinstitute.hellbender.cmdline.programgroups.VariantProgramGroup;
 import org.broadinstitute.hellbender.engine.FeatureDataSource;
 import org.broadinstitute.hellbender.exceptions.UserException;
 import org.broadinstitute.hellbender.utils.Utils;
+import org.broadinstitute.hellbender.utils.tsv.DataLine;
+import org.broadinstitute.hellbender.utils.tsv.TableColumnCollection;
+import org.broadinstitute.hellbender.utils.tsv.TableWriter;
 
 import java.io.File;
-import java.io.FileWriter;
 import java.io.IOException;
-import java.io.PrintWriter;
 import java.nio.file.Files;
 import java.nio.file.Paths;
-import java.util.stream.Stream;
+import java.util.List;
+import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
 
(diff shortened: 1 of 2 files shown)


davidbenjamin:
@takutosato I'm looking at it.  My preliminary pre-review comment is that I will probably want you to use one of our official classes for writing tables eg. `TableWriter`.

takutosato:
Didn't know about `TableWriter`. Sounds useful. Will look into it!

takutosato:
@davidbenjamin I pushed the version with `TableWriter`. I changed the output format a little but other than that it does the same thing as before.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/CountFalsePositives.java
```diff
@@ -69,10 +69,24 @@ private static String countFalsePositivesAndBuildString(File m2vcf) {
...
+        private FalsePositiveTableWriter(final File output) throws IOException {
+            super(output, new TableColumnCollection("filename", "snp", "indel"));
+        }
+
+        @Override
+        protected void composeLine(final File vcf, final DataLine dataLine) {
+            Utils.nonNull(vcf, "the vcf file cannot be null");
+            Utils.nonNull(dataLine, "dataLine cannot be null");
+
+            Pair<Long, Long> falsePositives = countFalsePositives(vcf);
```
davidbenjamin: Rename the variable so that the `Pair` is self-documenting e.g. `snpAndIndelFalsePositiveCounts`.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/CountFalsePositives.java
```diff
@@ -36,31 +39,28 @@
...
-                PrintWriter outputWriter = new PrintWriter(new FileWriter(output))
-        ) {
-            sb = vcfFiles.reduce(sb,
-                    (acc, f) -> acc.append(countFalsePositivesAndBuildString(new File(f))),
-                    StringBuilder::append);
-            outputWriter.print(sb.toString());
+        try ( FalsePositiveTableWriter writer = new FalsePositiveTableWriter(output) ) {
+            List<String> vcfFilePaths = Files.readAllLines(Paths.get(vcfList.toString()));
+            List<File> vcfFiles = vcfFilePaths.stream().map(File::new).collect(Collectors.toList());
+            writer.writeAllRecords(vcfFiles);
```
davidbenjamin: I think it's preferable to take all logic out of the `TableWriter`, so that you calculate false positives here and not inside `composeLine`.  This may require creating another nested class called eg `FalsePositiveRecord` that is basically a struct of the filename and snp and indel counts, but t's worth it.

davidbenjamin:
@takutosato back to you

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/CountFalsePositives.java
```diff
@@ -69,10 +69,24 @@ private static String countFalsePositivesAndBuildString(File m2vcf) {
...
+        return Pair.of(snpFalsePositives, indelFalsePositives);
+    }
 
-        StringBuilder sb = new StringBuilder();
-        sb.append(String.format("%s\t%d\t%s\n", sampleName, snpFalsePositives, "snp"));
-        sb.append(String.format("%s\t%d\t%s\n", sampleName, indelFalsePositives, "indel"));
-        return sb.toString();
+    private class FalsePositiveTableWriter extends TableWriter<File> {
+        private FalsePositiveTableWriter(final File output) throws IOException {
+            super(output, new TableColumnCollection("filename", "snp", "indel"));
```
takutosato: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/CountFalsePositives.java
takutosato: done

takutosato:
@davidbenjamin thanks for reviewing! back to you.

davidbenjamin:
@takutosato Squash.  Merge.  Rinse.  Repeat.

PR #850, opened 2017-01-11
Artifact-in-normal filter AND two-pass M2 calling/filtering

### LINKED ISSUE #821: Filter artifacts common to tumor and normal
If the normal has a lot of alt reads but not enough to call a germline het, *and* if its alt allele fraction is similar to the tumor's, it could be due to a shared artifact.  (For example, even if sample prep and sequencing were completely different, there may be a common alignment error).  Let's model this.  A command line that reveals three relevant instances is:

```java
#!/bin/bash
ref=/humgen/gsa-hpprojects/GATK/bundle/2.8/b37/human_g1k_v37.fasta
tumor=/humgen/gsa-scr1/schandra/cooperjam_MuTect2PASSVariantsAreInNormals/GATK_error/AMUM.post.subset.bam
normal=/humgen/gsa-scr1/schandra/cooperjam_MuTect2PASSVariantsAreInNormals/GATK_error/AMUM.normal.subset.bam
java -jar gatk-protected.jar Mutect2 -R $ref -I $tumor -I $normal -tumor AMUM.post -normal AMUM.normal -L 6:32609886-32610486 -L 7:15725829-15726429 -L 8:75944109-75944709 -O results.vcf
```

davidbenjamin:
Closes #821.

### DIFF
--- a/scripts/mutect2_wdl/mutect2.wdl
+++ b/scripts/mutect2_wdl/mutect2.wdl
@@ -62,6 +62,22 @@ task GatherVCFs {
   }
 }
 
+task Filter {
+  File gatk4_jar
+  File unfiltered_calls
+  String output_vcf_name
+
+  command <<<
+  	java -Xmx4g -jar ${gatk4_jar} FilterMutectCalls \
+  	-V ${unfiltered_calls} \
+    -O ${output_vcf_name}.vcf
+  >>>
+
+  output {
+    File output_vcf = "${output_vcf_name}.vcf"
+  }
+}
+
 # Warning: this task does not work in the cloud.
 task SplitIntervals {
   File picard_jar
@@ -130,19 +146,25 @@ workflow Mutect2 {
         dbsnp_index = dbsnp_index,
         cosmic = cosmic,
         cosmic_index = cosmic_index
-    } 
-	}
+    }
+  }
 
   call GatherVCFs {
     input:
       picard_jar = picard_jar,
       input_vcfs = M2.output_vcf,
-      output_vcf_name = "${tumor_sample_name}-vs-${normal_sample_name}"
+      output_vcf_name = "${tumor_sample_name}-vs-${normal_sample_name}-unfiltered"
+  }
+
+call Filter {
+    input:
+      gatk4_jar = gatk4_jar,
+      unfiltered_calls = GatherVCFs.output_vcf,
+      output_vcf_name = "${tumor_sample_name}-vs-${normal_sample_name}-filtered"
   }
 
   output {
-    File mutect2_vcf = GatherVCFs.output_vcf
-  }
-
-
+        File unfiltered_vcf = GatherVCFs.output_vcf
+        File filtered_vcf = Filter.output_vcf
+    }
 }
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/FilterMutectCalls.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/FilterMutectCalls.java
@@ -1,0 +1,75 @@
+package org.broadinstitute.hellbender.tools.walkers.mutect;
+
+import htsjdk.variant.variantcontext.VariantContext;
+import htsjdk.variant.variantcontext.VariantContextBuilder;
+import htsjdk.variant.variantcontext.writer.VariantContextWriter;
+import htsjdk.variant.vcf.VCFFilterHeaderLine;
+import htsjdk.variant.vcf.VCFHeader;
+import htsjdk.variant.vcf.VCFHeaderLine;
+import org.broadinstitute.barclay.argparser.Argument;
+import org.broadinstitute.barclay.argparser.ArgumentCollection;
+import org.broadinstitute.barclay.argparser.CommandLineProgramProperties;
+import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
+import org.broadinstitute.hellbender.cmdline.programgroups.VariantProgramGroup;
+import org.broadinstitute.hellbender.engine.*;
+import org.broadinstitute.hellbender.utils.variant.GATKVCFHeaderLines;
+import org.broadinstitute.hellbender.utils.variant.GATKVariantContextUtils;
+
+import java.io.File;
+import java.util.*;
+
+@CommandLineProgramProperties(
+        summary = "Filter somatic SNPs and indels called by Mutect2",
+        oneLineSummary = "Filter somatic SNPs and indels called by Mutect2",
+        programGroup = VariantProgramGroup.class
+)
+public final class FilterMutectCalls extends VariantWalker {
+
+    @Argument(fullName= StandardArgumentDefinitions.OUTPUT_LONG_NAME,
+            shortName=StandardArgumentDefinitions.OUTPUT_SHORT_NAME,
+            doc="The output filtered VCF file", optional=false)
+    private final String outputVcf = null;
+
+    @ArgumentCollection
+    protected M2FiltersArgumentCollection MTFAC = new M2FiltersArgumentCollection();
+
+    private VariantContextWriter vcfWriter;
+
+    private final List<VariantContext> unfilteredCalls = new ArrayList<>();
+
+
+    @Override
+    public void onTraversalStart() {
+        final VCFHeader inputHeader = getHeaderForVariants();
+        final Set<VCFHeaderLine> headerLines = new HashSet<>(inputHeader.getMetaDataInSortedOrder());
+        Mutect2FilteringEngine.M_2_FILTER_NAMES.stream().map(GATKVCFHeaderLines::getFilterLine).forEach(headerLines::add);
+        headerLines.add(new VCFFilterHeaderLine(Mutect2FilteringEngine.ARTIFACT_IN_NORMAL_FILTER_NAME, "artifact_in_normal"));
+
+        final VCFHeader vcfHeader = new VCFHeader(headerLines, inputHeader.getGenotypeSamples());
+        vcfWriter = GATKVariantContextUtils.createVCFWriter(new File(outputVcf), getBestAvailableSequenceDictionary(), lenientVCFProcessing);
+        vcfWriter.writeHeader(vcfHeader);
+    }
+
+    @Override
+    public Object onTraversalSuccess() {
+        // TODO: implement sophisticated filtering
+        for (final VariantContext vc : unfilteredCalls) {
+            final VariantContextBuilder vcb = new VariantContextBuilder(vc);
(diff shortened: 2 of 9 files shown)


davidbenjamin:
@takutosato can you take a look?  BTW, I verified that the wdl works -- /dsde/working/davidben/mutect/test_two_pass_wdl is set up to run the wdl on our integration test bams, which is really quick,

lbergelson:
Why are you separating out the calling from the filtering?  It seems like a backwards step for usability?  Are you going to maintain the option to do filtering as you call?

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/FilterMutectCalls.java
```diff
@@ -0,0 +1,74 @@
...
+        Mutect2FilteringEngine.M_2_FILTER_NAMES.stream().map(GATKVCFHeaderLines::getFilterLine).forEach(headerLines::add);
+        headerLines.add(new VCFFilterHeaderLine(Mutect2FilteringEngine.ARTIFACT_IN_NORMAL_FILTER_NAME, "artifact_in_normal"));
+
+        final VCFHeader vcfHeader = new VCFHeader(headerLines, inputHeader.getGenotypeSamples());
+        vcfWriter = GATKVariantContextUtils.createVCFWriter(new File(outputVcf), getBestAvailableSequenceDictionary(), lenientVCFProcessing);
+        vcfWriter.writeHeader(vcfHeader);
+    }
+
+    @Override
+    public Object onTraversalSuccess() {
```
lbergelson: Is there a reason it does all the work in onSuccess instead of streaming it during apply?

davidbenjamin: @lbergelson Soon streaming won't be an option because we will need to store all calls in memory and do cool machine learning stuff with all of them simultaneously.  Putting the work into `onSuccess` anticipates this.

davidbenjamin:
I see no need to maintain an option to call and filter within a single CLI.  In terms of usability, additional somatic tools like contamination and oxo-G mean that everyone should run M2 from the WDL anyway.  Also, calling is so much more expensive than filtering that from a development point of view it makes sense to be able to play around with filtering without much CPU cost.  Users might also want to do the same.  Finally, there's a lot to be said for mimicking the HC/VQSR division of labor for uniformity within the GATK.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/FilterMutectCalls.java
lbergelson: Ah, I see.  So this is the first step in the fabled VQSR for cancer?

lbergelson:
Fair enough.  Is the filtering going to be multisample filtering like VQSR?

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/SomaticGenotypingEngine.java
```diff
@@ -258,18 +262,19 @@ private void addStrandBiasAnnotations(final ReadLikelihoods<Allele> likelihoods,
      * @return                                      genotype likelihoods for homRef and het for each alternate allele
      */
     private PerAlleleCollection<Double> getHetGenotypeLogLikelihoods(final ReadLikelihoods<Allele> likelihoods,
-                                                                     final String sampleName,
+                                                                     final String sampleNameForAlleleFractions,
+                                                                     final String sampleNameForLikelihoods,
                                                                      final OptionalDouble givenAltAlleleFraction,
                                                                      final Strand strand) {
-        final PerAlleleCollection<Double> alleleFractions = getAlleleFractions(likelihoods, sampleName, strand);
+        final PerAlleleCollection<Double> alleleFractions = getAlleleFractions(likelihoods, sampleNameForAlleleFractions, strand);
```
takutosato: We should not compute `alleleFractions` if `givenAltAlleleFraction ` is not empty.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/SomaticGenotypingEngine.java
```diff
@@ -189,13 +193,13 @@ public CalledHaplotypes callMutations(
...
     private PerAlleleCollection<Double> getHetGenotypeLogOdds(ReadLikelihoods<Allele> likelihoods,
-                                                              final boolean isNormal,
-                                                              final Strand strand) {
-        final String sample = isNormal ? matchedNormalSampleName : tumorSampleName;
-        final OptionalDouble givenAltAlleleFraction = isNormal ? GERMLINE_HET_ALT_FRACTION : NO_FIXED_TUMOR_ALT_FRACTION;
-        final PerAlleleCollection<Double> hetGenotypeLogLks = getHetGenotypeLogLikelihoods(likelihoods, sample, givenAltAlleleFraction, strand);
+                                                              final boolean isGermline,
+                                                              final Strand strand, final String sample) {
+        final OptionalDouble givenAltAlleleFraction = isGermline ? GERMLINE_HET_ALT_FRACTION : NO_FIXED_TUMOR_ALT_FRACTION;
+        final String sampleForAlleleFractions = (sample == matchedNormalSampleName && isGermline) ? matchedNormalSampleName : tumorSampleName;
```
takutosato: As I understand the code we have 3 use cases of `getHetGenotypeLogOdds`:

1. tumor lod: use the allele fraction of tumor and reads from tumor
2. normal lod: use 0.5 (`GERMLINE_HET_ALT_FRACTION`) as the allele fraction and reads from normal
3. normal artifact lod: use the allele fraction of _tumor_ and reads from normal

With the current code, for case 3) we use the allele fraction of the normal. My understanding is `sampleForAlleleFractions` should always be set to  `tumorSampleName `.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/FilterMutectCalls.java
davidbenjamin: Yes it is.

scripts/mutect2_wdl/mutect2.wdl
```diff
@@ -130,19 +146,25 @@ workflow Mutect2 {
...
-	}
+    }
+  }
 
   call GatherVCFs {
     input:
       picard_jar = picard_jar,
       input_vcfs = M2.output_vcf,
-      output_vcf_name = "${tumor_sample_name}-vs-${normal_sample_name}"
+      output_vcf_name = "${tumor_sample_name}-vs-${normal_sample_name}_unfiltered"
```
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2.java
```diff
@@ -146,14 +144,8 @@ public void onTraversalStart() {
         m2Engine.writeHeader(vcfWriter, sequenceDictionary);
     }
 
-    //TODO: implement sophisticated filtering
```
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2FilteringEngine.java
```diff
@@ -65,22 +73,36 @@ private static void applyGermlineVariantFilter(final M2ArgumentCollection MTAC,
...
-        if (MTAC.ENABLE_STRAND_ARTIFACT_FILTER) {
+    // filter out anything called in tumor that would also be called in the normal if it were treated as a tumor.
+    // this handles shared artifacts, such as ones due to alignment and any shared aspects of sequencing
+    private static void applyArtifactInNormalFilter(final M2FiltersArgumentCollection MTFAC, final VariantContext vc, final Collection<String> filters) {
+        if (!vc.hasAttribute(SomaticGenotypingEngine.NORMAL_ARTIFACT_LOD_ATTRIBUTE)) {
+            return;
+        }
+
+        // take maximum of artifact-in-normal log odds
+        final double normalArtifactLod = MathUtils.arrayMax(GATKProtectedVariantContextUtils.getAttributeAsDoubleArray(vc, SomaticGenotypingEngine.NORMAL_ARTIFACT_LOD_ATTRIBUTE, () -> null, -1));
```
davidbenjamin: I agree.  Done.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/SomaticGenotypingEngine.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/SomaticGenotypingEngine.java
davidbenjamin: I did case 3) this way intentionally, the idea being that the artifact might not have the same allele fraction in the normal as in the tumor, but its presence in the normal is still useful evidence.  The way it works now is "If there were no tumor sample and we pretended the normal was a tumor, would we call this allele?"

davidbenjamin:
@takutosato back to you.  How do you feel about `getHetGenotypeLogOdds`?

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/SomaticGenotypingEngine.java
takutosato: I see, that makes sense.

davidbenjamin:
@takutosato I like getting rid of the intermediate variables.  As for calculating allele fractions inside or outside the likelihoods function, it's a toss-up to me.  On the one hand it's nice if the method doesn't do too much, on the other hand, the allele fractions are really latent variables and as such can be thought of as an implementation detail of the likelihoods.  Since this refactoring is coming soon I would say you can go ahead with the changes in this branch but don't work too hard on the allele fractions.

What should I do with this PR for now?

takutosato:
@davidbenjamin I see your point. Let's hold off on my changes and merge your branch as is. We can discuss in person whether to put in my changes when you return to office. How does that sound?

takutosato:
Actually I want to request one change in your code.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/SomaticGenotypingEngine.java
takutosato: `sample.equals(matchedNormalSampleName)`

davidbenjamin: done

PR #859, opened 2017-01-13
Remove tools.jar dependency.

cmnbroad:
We really shouldn't have a compile dependency on tools.jar; doing so propagates it through the uber.jar. Building and testing seem to pass without it. If there is another gradle target (or some runtime code) that requires it, we can try to address that. Otherwise the dependency should be removed.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -77,7 +77,6 @@ build.dependsOn installDist
 check.dependsOn installDist
 
 dependencies {
-    compile files("${System.properties['java.home']}/../lib/tools.jar")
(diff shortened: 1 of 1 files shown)


lbergelson:
I can't figure out why we had this originally.. It looks like I added it in some very early iteration of the build file, but I can't for the life of me think why.  I probably copied it from something that was actually using the unsafe classes and didn't remove it.  Seems right to get rid of it.

lbergelson:
@LeeTL1220 Any reservations about removing this?

LeeTL1220:
No.  So long as all tests pass.

PR #860, opened 2017-01-17
removing duplicate code from ReadCountsReader

lbergelson:
removing some duplicated code from ReadCountsReader constructors
removing an unused type parameter


lbergelson:
@asmirnov239 I removed the code in those duplicated constructors that you were talking about when we were taking a look at the line reader.   Could you review?

asmirnov239:
Thanks, that looks good!

PR #876, opened 2017-01-27
strand artifact annotation bug fix

takutosato:
* use strand-agnostic allele fraction when we compute the strand-specific alt genotype likelihoods
* only sum over reads in one strand when we compute the strand-specific ref genotype likelihoods 

Strand-specific tumor lods look much more reasonable now


takutosato:
Hi @davidbenjamin, here are the bug fixes for the gatk3 strand artifact filter. The bug was only in the strand-artifact filter code after all.

davidbenjamin:
@takutosato great catch and great fix

PR #888, opened 2017-01-31
add runtime clauses and fix glob

takutosato:
@davidbenjamin This change adds runtime clauses to `mutect2.wdl`. Now this single wdl can run in the cloud and on SGE. Users must supply the runtime parameters (e.g. `$__docker__`) themselves. I have a script that populates those parameters for broad internal uses; this script will live in the palantir repo under `Analysis/mutect2-evaluation`. Cromwell ignores runtime clauses when irrelevant (e.g. when we submit the jobs to SGE).

I also modified the code surrounding `glob` so it works on cromwell v24; in v24 and later, cromwell `glob` looks for files in the current working directory only.

### DIFF
--- a/scripts/mutect2_wdl/mutect2.wdl
+++ b/scripts/mutect2_wdl/mutect2.wdl
@@ -46,6 +46,12 @@ task M2 {
     -O ${tumor_sample_name}-vs-${normal_sample_name}.vcf
   }
 
+  runtime {
+    docker: "$__docker__"
+    memory: "$__large_memory__"
+    disks: "local-disk " + $__large_disk__ + " HDD"
+  }
+
   output {
     File output_vcf = "${tumor_sample_name}-vs-${normal_sample_name}.vcf"
   }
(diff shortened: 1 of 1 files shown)


scripts/mutect2_wdl/mutect2.wdl
```diff
@@ -56,6 +62,12 @@ task GatherVCFs {
     OUTPUT=${output_vcf_name}.vcf
   >>>
 
+  runtime {
+    docker: "$__docker__"
```
takutosato: As suggested by @davidbenjamin make docker an argument to the wdl. similarly for memory, disk.

scripts/mutect2_wdl/mutect2.wdl
```diff
@@ -91,10 +109,23 @@ task SplitIntervals {
     I=${intervals} \
     O=intervalFileDir \
     SCATTER_COUNT=${scatterCount}
+
+    find $PWD/intervalFileDir -iname *.interval_list | sort > interval_lists.txt
```
takutosato: Confirm that copying the files is necessary.

davidbenjamin: @takutosato In that case you can go ahead and rebase and I will review the rebase.

davidbenjamin:
@takutosato You explained everything to my satisfaction and may squash and merge.

takutosato:
Thanks @davidbenjamin. I split them into two commits in case we want to revisit the glob fix.

PR #889, opened 2017-02-01
PlotSegmentedCopyRatio and PlotACNVResults now require a sequence dictionary as input.

### LINKED ISSUE #686: Allow sequence dictionary as input for CNV/ACNV plotting.
No reason that we should be limited to hg19 except for the centromere locations, which we can turn off or make optional.

samuelklee:
@achevali please review (don't be alarmed by the size of the PR, most of it is test files and relatively trivial refactorings), @sooheelee please feel free to comment.

Contigs and lengths in the sequence dictionary are used to determine the appropriate genomic range to plot over and the relative horizontal scaling of the contigs.  A few notes on behavior:

- Contigs are plotted in the order given in the dictionary and are labeled by their corresponding names.
- Contig names containing ":" will fail.  Although this is technically allowed by the SAM specification, I don't think it should be a problem---but let me know if you think otherwise.
- If there is data present on contigs that are missing from the sequence dictionary, a warning is thrown and only the contigs that are present in the dictionary are plotted.
- Conversely, contigs without data are still plotted as empty regions as long as they are specified in the dictionary.
- Per-segment plotting has been removed.
- Previously, a "useSexChromosomes" flag was used to indicate whether "X" and "Y" contigs would be plotted.  This would fail if "chr" was present in the contig names.  Furthermore, the previous implementation required exactly 22 autosomal contigs, to which the sex chromosomes were optionally appended.  This flag has thus been removed, as now any undesired contigs can simply be removed from the dictionary.
- A cutoff on the minimum contig length can be specified via a parameter.  This is handy in the typical use case in which one wants to filter out mitochondrial contigs, unlocalized contigs, etc.  Otherwise, a separate dictionary file that has these contigs removed would be required solely for plotting purposes.

And a few notes on refactoring:

- I added more argument validation and unit tests.
- I removed the \*Plotter backend classes and moved most of the logic to private methods in the Plot\* frontend CLI classes.  My reasoning is that 1) the R scripts are essentially already backends, and 2) the \*Plotter" unit tests and the Plot\* integration tests were exactly redundant and thus a pain to maintain.
- There is a lot of almost-but-not-quite-duplicate code in PlotSegmentedCopyRatio and PlotACNVResults.  It's possible that some methods could be extracted, but I think this is not worth the trouble.  Might be worth revisiting once ReadCountCollection and AllelicCountCollection are refactored, which @asmirnov239 is working on (note also we might eventually want to have an analogous SegmentCollection, which would make some of the code in these CLIs even cleaner).

Closes #686.


src/main/java/org/broadinstitute/hellbender/tools/exome/plotting/PlotSegmentedCopyRatio.java
```diff
@@ -0,0 +1,242 @@
...
+        summary = "Create plots of denoised and segmented copy ratio.",
+        oneLineSummary = "Create plots of denoised and segmented copy ratio",
+        programGroup = CopyNumberProgramGroup.class
+)
+public final class PlotSegmentedCopyRatio extends CommandLineProgram {
+    private static final String CNV_PLOTTING_R_LIBRARY = "CNVPlottingLibrary.R";
+    private static final String COPY_RATIO_PLOTTING_R_SCRIPT = "CopyRatioPlotting.R";
+
+    private static final String CONTIG_DELIMITER = ":";
+    private static final int DEFAULT_MINIMUM_CONTIG_LENGTH = 1000000;
```
achevali: There should be a comment somewhere noting that this is mainly to remove MT regions etc.

samuelklee: This is mentioned in the parameter documentation, but I can put a comment here as well.

samuelklee:
Any more comments, @achevali?  @sooheelee?

src/main/java/org/broadinstitute/hellbender/tools/exome/plotting/PlotSegmentedCopyRatio.java
achevali: I think inline comment is helpful for devs.

sooheelee:
Just seeing this. Sorry, I'm super focused until this Thursday on M2 tutorial development and then vacation. 

- GRCh38 uses colons (`:`) in contig names. GATKv3.7+ handles these correctly so our tools going forward in GATK4 should too.
- Do we allow `-L` in plotting? So folks can zoom in on regions that interest them?

samuelklee:
Ah, OK, thanks @sooheelee!  Did not know that about GRCh38.  I'll just change the delimiter to the string "CONTIG_DELIMITER"---very unlikely that contig names will contain that string, right?  Although this is a bit hacky, the only other option I see is to use some nonstandard ASCII character, which is also kinda gross!

As for allowing "-L" to zoom, that's a good suggestion, but will probably have to wait until a future PR.  Baby steps!

PR #890, opened 2017-02-03
Tumor-only Mutect wdl

davidbenjamin:


davidbenjamin:
@LeeTL1220 Here it is, hack and all.

scripts/mutect2_wdl/mutect2.wdl
```diff
@@ -22,9 +22,9 @@ task M2 {
   File tumor_bam
```
LeeTL1220: I assume that this has been tested

Please put a note on the cromwell issue that you ahve a workaround.

scripts/mutect2_wdl/mutect2.wdl
```diff
@@ -33,21 +33,24 @@ task M2 {
   File? cosmic_index
 
   command {
+    if [ -s "$normal_sample_name" -a "$normal_sample_name" != "NO_NORMAL" ]; then
```
LeeTL1220: Is this proper syntax?

scripts/mutect2_wdl/mutect2.wdl
```diff
@@ -33,21 +33,24 @@ task M2 {
...
     ${"--cosmic " + cosmic} \
     ${"--normal_panel " + pon} \
     -L ${intervals} \
-    -O ${tumor_sample_name}-vs-${normal_sample_name}.vcf
+    -O "${tumor_sample_name}.vcf"
   }
 
   output {
-    File output_vcf = "${tumor_sample_name}-vs-${normal_sample_name}.vcf"
+    File output_vcf = "${tumor_sample_name}.vcf"
```
LeeTL1220: Please rename this output file to something more descriptive...

LeeTL1220:
@davidbenjamin Back to you

scripts/mutect2_wdl/mutect2.wdl
davidbenjamin: Yes, but I think I have made it nicer in the latest commit.

scripts/mutect2_wdl/mutect2.wdl
davidbenjamin: Done, but it requires a bit of a hack because the cromwell doesn't let you refer to a `String?` in output or input.  Since I couldn't use `File output_vcf = "${tumor_sample_name}{"-vs-" + normal_sample_name}.vcf"` I wrote a brief task to compute this in bash.

scripts/mutect2_wdl/mutect2.wdl
davidbenjamin: Yes, tested with and without the normal.

davidbenjamin: Put note on cromwell issue.

davidbenjamin:
Back to @LeeTL1220

scripts/mutect2_wdl/mutect2.wdl
```diff
@@ -22,32 +22,57 @@ task M2 {
...
+
+  output {
+    File output_vcf = "${output_vcf_name}.vcf"
+  }
+}
+
+# HACK: cromwell can't handle the optional normal sample name in output or input --
+# string interpolation of optionals only works inside a command block
+# thus we use this hack
+task MakeOutputVcfName {
```
LeeTL1220: Please make the ``normal_bam`` parameter a String ... Otherwise, you will localize a bam file on the cloud.

Alternatively, since normal bam and normal sample name are usually both blank or both populated, can't you just ditch the normal_bam parameter entirely?  Then just use normal_sample_name if that is populated?

scripts/mutect2_wdl/mutect2.wdl
```diff
@@ -71,14 +96,14 @@ task MergeVCFs {
...
+  String output_prepend
 
   command {
-  	java -Xmx4g -jar ${gatk4_jar} FilterMutectCalls -V ${unfiltered_calls} -O ${output_vcf_name}.vcf
+  	java -Xmx4g -jar ${gatk4_jar} FilterMutectCalls -V ${unfiltered_calls} -O ${output_prepend}-filtered.vcf
   }
 
   output {
-    File output_vcf = "${output_vcf_name}.vcf"
+    File output_vcf = "${output_prepend}-filtered.vcf"
```
LeeTL1220: Can you please rename this output parameter to something more specific?

LeeTL1220:
@davidbenjamin Back to you... a couple more items.  Apologies if I missed in the first go-around.

scripts/mutect2_wdl/mutect2.wdl
davidbenjamin: I could do the alternative, but the way it works now allows you to either leave the normal fields blank or put in some placeholder like `TUMOR_ONLY` or `TEMPORARY_FORMATTING_HACK`.  I'm loathe to make an empty field the only option because the difference between the correct thee-tabs-and-nothing-else format and an incorrect format could be invisible.

It's true that it's awkward to pass both the bam and the sample name to this task, but since it's not exposed to the user I'm comfortable with the redundancy.

I will make `normal_bam` a `String`.  Do you mean just for this particular task?

scripts/mutect2_wdl/mutect2.wdl
davidbenjamin: I understand you to be referring to the variable name `output_vcf` as opposed to its value `"${output_prepend}-filtered.vcf"`.  Let me know if I'm mixed up.

davidbenjamin:
Back to @LeeTL1220

scripts/mutect2_wdl/mutect2.wdl
```diff
@@ -103,7 +103,7 @@ task Filter {
   }
 
   output {
-    File output_vcf = "${output_prepend}-filtered.vcf"
+    File filtered_vcf = "${output_prepend}-filtered.vcf"
```
LeeTL1220: How about something more like ``m2_filtered_vcf``?

scripts/mutect2_wdl/mutect2.wdl
```diff
@@ -60,7 +60,7 @@ task M2 {
 # thus we use this hack
 task MakeOutputVcfName {
   String tumor_sample_name
-  File? normal_bam
+  String? normal_bam
```
LeeTL1220: Good ... you tested this change?

LeeTL1220:
@davidbenjamin Back to you

scripts/mutect2_wdl/mutect2.wdl
davidbenjamin: Always.

scripts/mutect2_wdl/mutect2.wdl
davidbenjamin: Sure.

davidbenjamin:
@LeeTL1220 Anything else?

scripts/mutect2_wdl/mutect2.wdl
davidbenjamin: done

scripts/mutect2_wdl/mutect2.wdl
```diff
@@ -246,14 +246,14 @@ workflow Mutect2 {
         input:
            gatk4_jar = gatk4_jar,
            output_prepend = MakeOutputVcfName.output_name,
-           m2_vcf = Filter.filtered_vcf,
+           m2_vcf = Filter._m2filtered_vcf,
```
LeeTL1220: Did you test this?  I doubt it....

davidbenjamin:
Tested!!!

davidbenjamin:
Failing after rebase. . . gotta fix that.

PR #900, opened 2017-02-14
Add "gradle localJar" as replacement for "gradle shadowJar", and adjust jar naming to be consistent with public

### LINKED ISSUE #873: gatk-launch fails after ./gradlew clean shadowJar
This is because the name for the shadow jar differs between public vs. protected, and `gatk-launch` looks for the name used in public. We should just have protected use the suffix `local.jar` like in public.

droazen:
-Add a new "gradle localJar" target that just runs shadowJar

-The localJar/shadowJar name now ends in a "local.jar" suffix, to
 better distinguish it from the spark jar. This is consistent with
 gatk public, and (critically) with the names that the gatk-launch
 script now expects. This fixes a failure in gatk-launch.

-Kept "gradle shadowJar" around for backwards compatibility, but removed
 mention from README

-Include the word "package" in the name for jars that contain all dependencies
 (again, for consistency with gatk public)

-Updated the README with basic information on building and running
 GATK4 copied from the public README, with links back to the public
 README for further details.

Resolves #873

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,11 @@ GATK4 development of the license-protected part of the toolkit
 
 This README is aimed at developers.  For user information, please see the [GATK 4 forum](http://gatkforums.broadinstitute.org/gatk/categories/gatk-4-alpha)
 
+Read GATK 4 README
+------------------------
+
+**Please refer to the [GATK 4 public repo README](https://github.com/broadinstitute/hellbender/blob/master/README.md) for general guidelines and how to setup your development environment.**
+
 Requirements
 ------------
 * R 3.1.3 see additional requirements below: [R package requirements](#r-required-packages)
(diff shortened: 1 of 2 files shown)


droazen:
@lbergelson or @LeeTL1220 please review

README.md
```diff
@@ -19,22 +24,66 @@ download and use an appropriate gradle version automatically.
...
+  ./gatk-launch PrintReads -I input.bam -O output.bam
+  ```
 
-`` ./gradlew shadowJar ``
+  ```
+  ./gatk-launch PrintReadsSpark -I input.bam -O output.bam
+  ```
 
-A jar file will appear in ``build/libs``.
+* See the [GATK4 public README](https://github.com/broadinstitute/hellbender/blob/master/README.md) for full instructions on using `gatk-launch` to run tools on a Spark-based cluster.
```
lbergelson: I think `Spark cluster` is clearer than `Spark-based cluster.

build.gradle
```diff
@@ -263,6 +263,10 @@ shadowJar {
     }
 }
 
+task localJar {
+    dependsOn shadowJar 
+}
+
 task sparkJar(type: ShadowJar) {
     configurations = [project.configurations.sparkConfiguration]
     classifier = 'spark'
```
LeeTL1220: Do the symlinks still work?  I.e. gatk-protected.jar is a symlink to the local jar?

droazen: Yes, I verified that the symlinks still work

README.md
droazen: Done

PR #906, opened 2017-02-16
Contamination tool

davidbenjamin:
Tool is tested on HCC1143 spike with 1%, 2%, 3%, 4%, and 5% NA12878.  Performance is great.  Also tested the WDL.

Note @takutosato @LeeTL1220 when you put the appropriate variants resource into the json for your own scripts, ask me for my ExAC common variants vcf.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/CalculateContamination.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/CalculateContamination.java
@@ -1,0 +1,131 @@
+package org.broadinstitute.hellbender.tools.walkers.contamination;
+
+import org.apache.commons.math3.stat.descriptive.rank.Median;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.broadinstitute.barclay.argparser.Argument;
+import org.broadinstitute.barclay.argparser.CommandLineProgramProperties;
+import org.broadinstitute.hellbender.cmdline.CommandLineProgram;
+import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
+import org.broadinstitute.hellbender.cmdline.programgroups.VariantProgramGroup;
+import org.broadinstitute.hellbender.tools.exome.HashedListTargetCollection;
+import org.broadinstitute.hellbender.tools.exome.TargetCollection;
+import org.broadinstitute.hellbender.utils.SimpleInterval;
+
+import java.io.File;
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+/**
+ * Given pileup data from {@link GetPileupSummaries} calculate the fraction of reads coming from exogenous human DNA.
+ *
+ * This tool is run as follows:
+ *
+ * java -jar gatk.jar GetPileupSummaries -V ExAC.vcf -V thousand_genomes.vcf -I input.bam -O pileups.table
+ * java -jar gatk.jar CalculateContamination -I pileups.table -O contamination.table
+ *
+ * Created by David Benjamin on 2/7/17.
+ */
+@CommandLineProgramProperties(
+        summary = "Calculate contamination",
+        oneLineSummary = "Calculate contamination",
+        programGroup = VariantProgramGroup.class
+)
+public class CalculateContamination extends CommandLineProgram {
+
+    private static final Logger logger = LogManager.getLogger(CalculateContamination.class);
+
+    @Argument(fullName = StandardArgumentDefinitions.INPUT_LONG_NAME,
+            shortName = StandardArgumentDefinitions.INPUT_SHORT_NAME,
+            doc="The input table", optional = false)
+    private File inputPileupSummariesTable;
+
+    @Argument(fullName= StandardArgumentDefinitions.OUTPUT_LONG_NAME,
+            shortName=StandardArgumentDefinitions.OUTPUT_SHORT_NAME,
+            doc="The output table", optional = false)
+    private final File outputTable = null;
+
+    private static final int CNV_SCALE = 1000000;
+
+    @Override
+    public Object doWork() {
+        final List<PileupSummary> pileupSummaries = PileupSummary.readPileupSummaries(inputPileupSummariesTable);
+        final List<PileupSummary> homAltSites = findConfidentHomAltSites(pileupSummaries);
+        final double contamination = homAltSites.isEmpty() ? 0.0 : calculateContamination(homAltSites);
+        ContaminationRecord.writeContaminationTable(Arrays.asList(new ContaminationRecord(ContaminationRecord.Level.WHOLE_BAM.toString(), contamination)), outputTable);
+
+        return "SUCCESS";
+    }
+
+    private double calculateContamination(List<PileupSummary> homAltSites) {
+
+        final long totalReadCount = homAltSites.stream().mapToLong(PileupSummary::getTotalCount).sum();
+        final long totalRefCount = homAltSites.stream().mapToLong(PileupSummary::getRefCount).sum();
+
+        // if eg ref is A, alt is C, then # of ref reads due to error is roughly (# of G read + # of T reads)/2
+        final long errorRefCount = homAltSites.stream().mapToLong(PileupSummary::getOtherAltCount).sum() / 2;
+        final long contaminationRefCount = Math.max(totalRefCount - errorRefCount, 0);
+        final double proportionRefReads = (double) contaminationRefCount / totalReadCount;
+        final double proportionOfContaminatingReadsThatAreRef = homAltSites.stream().mapToDouble(PileupSummary::getRefFrequency).average().getAsDouble();
+        final double contamination = proportionRefReads / proportionOfContaminatingReadsThatAreRef;
+
+        logger.info(String.format("In %d homozygous variant sites we find %d reference reads due to contamination and %d" +
+                        " due to to sequencing error out of a total %d reads.", homAltSites.size(), contaminationRefCount, errorRefCount, totalReadCount));
+        logger.info(String.format("We estimate the fraction of contaminating ref reads in these homAltSites to be %.3f." +
+                "Based on population data, we expect reference reads to comprise a fraction %.3f of all contaminating reads." +
+                "Therefore, we estimate a contamination of %.3f", proportionRefReads, proportionOfContaminatingReadsThatAreRef, contamination));
+
+        return contamination;
+    }
+
+    private static List<PileupSummary> findConfidentHomAltSites(List<PileupSummary> sites) {
+        if (sites.isEmpty()) {
+            return new ArrayList<>();
+        }
+        final TargetCollection<PileupSummary> tc = new HashedListTargetCollection<>(sites);
+        final double averageCoverage = sites.stream().mapToInt(PileupSummary::getTotalCount).average().getAsDouble();
+        final List<Double> smoothedCopyRatios = new ArrayList<>();
+        final List<Double> hetRatios = new ArrayList<>();
+
+        for (final PileupSummary site : sites) {
+            final SimpleInterval nearbySpan = new SimpleInterval(site.getContig(), Math.max(1, site.getStart() - CNV_SCALE), site.getEnd() + CNV_SCALE);
+            final List<PileupSummary> nearbySites = tc.targets(nearbySpan);
+
+            final double averageNearbyCopyRatio = nearbySites.stream().mapToDouble(s -> s.getTotalCount()/averageCoverage).average().orElseGet(() -> 0);
+            smoothedCopyRatios.add(averageNearbyCopyRatio);
+
+            final double expectedNumberOfNearbyHets = nearbySites.stream().mapToDouble(PileupSummary::getAlleleFrequency).map(x -> 2*x*(1-x)).sum();
+            final long numberOfNearbyHets = nearbySites.stream().mapToDouble(PileupSummary::getAltFraction).filter(x -> 0.4 < x && x < 0.6).count();
+            final double hetRatio = numberOfNearbyHets / expectedNumberOfNearbyHets;
+            hetRatios.add(hetRatio);
+        }
+
+        final double medianSmoothedCopyRatio = new Median().evaluate(smoothedCopyRatios.stream().mapToDouble(x->x).toArray());
+        final List<Integer> indicesWithAnomalousCopyRatio = IntStream.range(0, sites.size())
+                .filter(n -> smoothedCopyRatios.get(n) < 0.8 * medianSmoothedCopyRatio || smoothedCopyRatios.get(n) > 2 *medianSmoothedCopyRatio)
+                .boxed().collect(Collectors.toList());
+
+        final double meanHetRatio = hetRatios.stream().mapToDouble(x->x).average().getAsDouble();
+        final List<Integer> indicesWithLossOfHeterozygosity = IntStream.range(0, sites.size())
+                .filter(n -> hetRatios.get(n) < meanHetRatio * 0.5)
+                .boxed().collect(Collectors.toList());
+
+        //TODO: as extra security, filter out sites that are near too many hom alts
+
+        logger.info(String.format("Excluding %d sites with low or high copy ratio and %d sites with potential loss of heterozygosity",
+                indicesWithAnomalousCopyRatio.size(), indicesWithLossOfHeterozygosity.size()));
+
+        logger.info(String.format("The average ratio of hets within distance %d to theoretically expected number of hets is %.3f", CNV_SCALE, meanHetRatio));
+
+        final Set<Integer> badSites = new TreeSet<>();
+        badSites.addAll(indicesWithAnomalousCopyRatio);
+        badSites.addAll(indicesWithLossOfHeterozygosity);
+
+        return IntStream.range(0, sites.size())
+                .filter(n -> !badSites.contains(n))
+                .mapToObj(sites::get)
+                .filter(s -> s.getAltFraction() > 0.8)
+                .collect(Collectors.toList());
+    }
+}
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/ContaminationRecord.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/ContaminationRecord.java
@@ -1,0 +1,107 @@
+package org.broadinstitute.hellbender.tools.walkers.contamination;
+
+import org.broadinstitute.hellbender.exceptions.UserException;
+import org.broadinstitute.hellbender.utils.Utils;
+import org.broadinstitute.hellbender.utils.tsv.DataLine;
+import org.broadinstitute.hellbender.utils.tsv.TableColumnCollection;
+import org.broadinstitute.hellbender.utils.tsv.TableReader;
+import org.broadinstitute.hellbender.utils.tsv.TableWriter;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.List;
+
+/**
+ * Created by David Benjamin on 2/13/17.
+ */
+public class ContaminationRecord {
+    private String level;
+    private double contamination;
+
+    public ContaminationRecord(final String level, final double contamination) {
+        this.level = level;
+        this.contamination = contamination;
+    }
+
+    public String getLevel() {
+        return level;
+    }
+
+    public double getContamination() {
+        return contamination;
+    }
+
+    //----- The following two public static methods read and write contamination files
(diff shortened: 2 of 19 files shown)


src/test/java/org/broadinstitute/hellbender/tools/walkers/contamination/CalculateContaminationIntegrationTest.java
```diff
@@ -0,0 +1,55 @@
...
+                ps.add(n % 10 == 0
+                        ? new PileupSummary(contig, position, (int) (depth*contamination), (int) (depth * (1 - contamination)), 0, alleleFrequency)
+                        : new PileupSummary(contig, position, depth / 2, depth/2, 1, alleleFrequency));
+        }
+
+        // same idea, but loss of heterozygosity -- these sites should be thrown out
+        for (int n = 1000; n < 1100; n++) {
+            int position = n * spacing;
+            ps.add(n % 10 == 0
+                    ? new PileupSummary(contig, position, (int) (depth*contamination), (int) (depth * (1 - contamination)), 0, alleleFrequency)
```
takutosato: ...or you didn't. Most of variants for n in (1000,1100) are thrown out as loss of heterozygosity as expected. But how? How can having one LOH in every 10 variants drastically reduce the `hetRatio` for _all_ variants nearby such that they are _all_ considered_ LOH?

src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/GetPileupSummaries.java
```diff
@@ -0,0 +1,99 @@
...
+        programGroup = VariantProgramGroup.class)
+public class GetPileupSummaries extends MultiVariantWalker {
+
+    public final String MAX_SITE_AF_LONG_NAME = "maximumPopulationAlleleFrequency";
+    public final String MIN_SITE_AF_LONG_NAME = "minimumPopulationAlleleFrequency";
+    public final String MAX_SITE_AF_SHORT_NAME = "maxAF";
+    public final String MIN_SITE_AF_SHORT_NAME = "minAF";
+
+    private static final double DEFAULT_MIN_POPULATION_AF = 0.01;
+    private static final double DEFAULT_MAX_POPULATION_AF = 0.2;
```
takutosato: Why did you choose 0.01 and 0.2?

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2FilteringEngine.java
```diff
@@ -1,24 +1,50 @@
...
+
+
+    public Mutect2FilteringEngine(final M2FiltersArgumentCollection MTFAC) {
+        this.MTFAC = MTFAC;
+        contamination = MTFAC.contaminationTable == null ? 0.0 : ContaminationRecord.readContaminationTable(MTFAC.contaminationTable).get(0).getContamination();
+
+    }
+
+    // very naive M1-style contamination filter -- remove calls with AF less than the contamination fraction
+    private void applyContaminationFilter(final M2FiltersArgumentCollection MTFAC, final VariantContext vc, final Collection<String> filters) {
```
takutosato: Here we want to extract the tumor alt allele _fraction_ from the genotype field, not the population allele _frequency_ from the INFO field, right?

src/main/java/org/broadinstitute/hellbender/utils/GATKProtectedVariantContextUtils.java
```diff
@@ -263,4 +273,21 @@ public static void setGenotypeQualityFromPLs(final GenotypeBuilder builder, fina
             return translate.apply(String.valueOf(value));
         }
     }
+
+    /**
+     * Get the pileup of reads covering a locus.  This is useful, for example, in VariantWalkers, which work on
+     * ReadsContexts and not AlignmentContexts.
+     */
+    public static ReadPileup getPileup(final Locatable loc, Iterable<GATKRead> reads) {
```
takutosato: make `reads` final

src/main/java/org/broadinstitute/hellbender/utils/GATKProtectedVariantContextUtils.java
```diff
@@ -263,4 +273,21 @@ public static void setGenotypeQualityFromPLs(final GenotypeBuilder builder, fina
...
+        final List<PileupElement> pile = StreamSupport.stream(reads.spliterator(), false)
+                .filter(r -> !r.failsVendorQualityCheck())
```
takutosato: You might want to use static Predicates defined in `ReadFilterLibrary.java` like `ReadFilterLibrary.PassesVendorQualityCheckReadFilter`. They do the same thing, it's just a matter of style.

src/test/java/org/broadinstitute/hellbender/utils/GATKProtectedVariantContextUtilsUnitTest.java
```diff
@@ -0,0 +1,58 @@
...
+        final int readLength = 3;
+
+        //this read doesn't overlap {@code loc}
+        final GATKRead read1 = ArtificialReadUtils.createArtificialRead(header, "read1", 0, 1, readLength);
+        read1.setBases(Utils.dupBytes((byte) 'A', readLength));
+        read1.setBaseQualities(Utils.dupBytes((byte) 30, readLength));
+        read1.setCigar("3M");
+
+        //this read overlaps {@code loc} with a Q30 'A'
+        final GATKRead read2 = ArtificialReadUtils.createArtificialRead(header, "read1", 0, 10, readLength);
```
takutosato: typo: the read names (the second argument to `createArtificialRead`) should be read2, read3, and so on.

src/test/java/org/broadinstitute/hellbender/utils/GATKProtectedVariantContextUtilsUnitTest.java
```diff
@@ -0,0 +1,58 @@
...
+        read2.setBaseQualities(Utils.dupBytes((byte) 30, readLength));
+        read2.setCigar("3M");
+
+        //this read overlaps {@code loc} with a Q20 'C' (due to the deletion)
+        final GATKRead read3 = ArtificialReadUtils.createArtificialRead(header, "read1", 0, 7, readLength);
+        read3.setBases(Utils.dupBytes((byte) 'C', readLength));
+        read3.setBaseQualities(Utils.dupBytes((byte) 20, readLength));
+        read3.setCigar("1M1D2M");
+
+        //this read doesn't overlaps {@code loc} due to the deletion
```
takutosato: doesn't overlap

src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/ContaminationRecord.java
```diff
@@ -0,0 +1,107 @@
...
+        ContaminationTableColumn(final String columnName) {
+            this.columnName = Utils.nonNull(columnName);
+        }
+
+        @Override
+        public String toString() {
+            return columnName;
+        }
+
+        public static final TableColumnCollection COLUMNS = new TableColumnCollection((Object[]) values());
```
davidbenjamin: It turns out you do if you want to call `values()` -- not an error, but Travis will fail because it yields a compiler warning -- but I don't really care about that and so `new TableColumnCollection(LEVEL, CONTAMINATION)` avoids the cast.

src/test/java/org/broadinstitute/hellbender/tools/walkers/contamination/CalculateContaminationIntegrationTest.java
davidbenjamin: It's not one LoH per 10 variants -- it's 1 true hom var and 9 LoH hets per 10 variants, which produces a very small het ratio.

src/test/java/org/broadinstitute/hellbender/tools/walkers/contamination/CalculateContaminationIntegrationTest.java
```diff
@@ -0,0 +1,55 @@
...
+        final File contaminationTable = createTempFile("contamination", ".table");
+
+        final String[] args = {
+                "-I", psTable.getAbsolutePath(),
+                "-O", contaminationTable.getAbsolutePath(),
+        };
+        runCommandLine(args);
+
+        final double calculatedContamination = ContaminationRecord.readContaminationTable(contaminationTable).get(0).getContamination();
+        Assert.assertEquals(calculatedContamination, 0.05 / (1 - alleleFrequency), 0.01);
```
davidbenjamin: True, and I renamed `contamination` --> `refContamination` to be more precise.

src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/CalculateContamination.java
```diff
@@ -0,0 +1,131 @@
...
+        logger.info(String.format("In %d homozygous variant sites we find %d reference reads due to contamination and %d" +
+                        " due to to sequencing error out of a total %d reads.", homAltSites.size(), contaminationRefCount, errorRefCount, totalReadCount));
+        logger.info(String.format("We estimate the fraction of contaminating ref reads in these homAltSites to be %.3f." +
+                "Based on population data, we expect reference reads to comprise a fraction %.3f of all contaminating reads." +
+                "Therefore, we estimate a contamination of %.3f", proportionRefReads, proportionOfContaminatingReadsThatAreRef, contamination));
+
+        return contamination;
+    }
+
+    public static List<PileupSummary> findConfidentHomAltSites(List<PileupSummary> sites) {
```
davidbenjamin: true

src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/GetPileupSummaries.java
davidbenjamin: Back-of-the-envelope reasoning.  ExAC has about 6 million SNPs.  Most of them are very rare, and if you iterate over them the tool is too slow.  If you take only variants with allele fraction 0.01 and above, you will get about 1% of all ExAC variants, 6 x 10^4 sites.  Because most of these sites are closer to the low end of allele frequency (it's kind of a power law), let's wave our hands and say that these 6 x 10^4 sites all have allele frequency 0.05 = 5 x 10^-2.  The probability of being hom alt is the square of this.  Then we get about 6 x 25 x 10^4 x 10^-4 or about 150 hom vars in this range per exome.

Now if our estimate is based on the fraction of ref reads in these 150 sites, as long as the allele frequency is not too big the variance in the number of ref reads is roughly equal to the expectation (binomial variance is p(1-p)*n ~ pn if p is small).  Thus if our estimate of the contamination is (ref reads / total reads) and the depth is d and N = 150 is the number of hom alt sites, the variance is (contamination * Nd / (Nd)^2) = contamination/(Nd).  If the depth is 20 and the contamination is 10% for these 150 hom alt sites we get a standard deviation of sqrt(0.1 / (3000)) ~ 0.006, which is pretty good.  If we included rarer sites we would not get many more hom alts and the IO cost in run time would be huge.

As for the upper limit, it probably doesn't make that much of a difference.  It's probably superstitious, but I just like the idea that the estimate would still be good (i.e. only off by 20%) even if we naively assumed that all contamination was ref.  The non-superstitious defense of this is that common variants' allele frequency might vary a lot from population to population, throwing off the estimate, whereas rarer variants have less wiggle room.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2FilteringEngine.java
davidbenjamin: True!

src/main/java/org/broadinstitute/hellbender/utils/GATKProtectedVariantContextUtils.java
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/utils/GATKProtectedVariantContextUtils.java
davidbenjamin: I like `Predicates` much more. done

src/test/java/org/broadinstitute/hellbender/utils/GATKProtectedVariantContextUtilsUnitTest.java
davidbenjamin: done

src/test/java/org/broadinstitute/hellbender/utils/GATKProtectedVariantContextUtilsUnitTest.java
davidbenjamin: done

davidbenjamin:
Note that I also came up with a way to get the tumor and normal sample names from the VCF header -- see the stuff in `Mutect2Engine` where it used to be a `VCFSimpleHeaderLine` and is now a `VCFHeaderLine`.

davidbenjamin:
@takutosato Back to you if tests pass, which, contrary to my usual unwise practice, I *won't* stay up to find out about.

src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/ContaminationRecord.java
takutosato: I see. Got it.

src/test/java/org/broadinstitute/hellbender/tools/walkers/contamination/CalculateContaminationIntegrationTest.java
takutosato: ahhh that was silly of me. thanks.

src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/GetPileupSummaries.java
takutosato: Thanks David! I think I get all but the "only off by 20%" in the last paragraph. Can we discuss this in person?

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/FilterMutectCalls.java
```diff
@@ -41,6 +41,7 @@
     @Override
     public void onTraversalStart() {
         final VCFHeader inputHeader = getHeaderForVariants();
+        final String tumorSample = inputHeader.getMetaDataLine(Mutect2Engine.TUMOR_SAMPLE_KEY_IN_VCF_HEADER).getValue();
```
takutosato: `tumorSample` here is not used

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
```diff
@@ -201,11 +204,9 @@ public void writeHeader(final VariantContextWriter vcfWriter, final SAMSequenceD
     private Set<VCFHeaderLine> getSampleHeaderLines(){
         final Set<VCFHeaderLine> sampleLines = new HashSet<>();
         if (hasNormal()) {
-            final Map<String, String> normalSampleHeaderAttributes = ImmutableMap.of("ID", "NORMAL", "name", MTAC.normalSampleName);
-            sampleLines.add(new VCFSimpleHeaderLine("sample", normalSampleHeaderAttributes));
+            sampleLines.add(new VCFHeaderLine(NORMAL_SAMPLE_KEY_IN_VCF_HEADER, MTAC.normalSampleName));
```
takutosato: awesome!

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2FilteringEngine.java
```diff
@@ -27,23 +28,24 @@
...
+        this.tumorSample = tumorSample;
     }
 
     // very naive M1-style contamination filter -- remove calls with AF less than the contamination fraction
     private void applyContaminationFilter(final M2FiltersArgumentCollection MTFAC, final VariantContext vc, final Collection<String> filters) {
-        if (vc.hasAttribute(VCFConstants.ALLELE_FREQUENCY_KEY)) {
-            final double[] alleleFrequencies = getArrayAttribute(vc, VCFConstants.ALLELE_FREQUENCY_KEY);
-            final double maxFrequency = MathUtils.arrayMax(alleleFrequencies);
+        final Genotype tumorGenotype = vc.getGenotypes(tumorSample).get(0);
+            final double[] alleleFractions = GATKProtectedVariantContextUtils.getAttributeAsDoubleArray(tumorGenotype, VCFConstants.ALLELE_FREQUENCY_KEY,
```
takutosato: indentation

takutosato: `vc.getGenotype(tumorSample)` is simpler

src/main/java/org/broadinstitute/hellbender/tools/walkers/contamination/GetPileupSummaries.java
davidbenjamin: Sure, maybe by talking it out we'll realize that the 20% ceiling in unnecessary.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2FilteringEngine.java
takutosato: Instead of `VCFConstants.ALLELE_FREQUENCY_KEY` we want to get `DEPTH_KEY` and `GENOTYPE_ALLELE_DEPTHS` and compute the allele fraction. I _think_ the alleles are not sorted though (we only sort the tumor lods in VCF; we should fix it). Besides I don't think getting allele fractions directly from the genotype fields don't play well with multi-allelic sites. So I suggest we annotate each variant context with a new INFO field `TUMOR_ALT_AF` of type `double[]` when we compute them in `SomaticGenotypingEngine`. I was thinking of taking this approach for `Concordance.java` when I need to retrieve the tumor alt allele fraction of a variant.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2FilteringEngine.java
```diff
@@ -27,23 +28,24 @@
...
+                    () -> new double[] {1.0}, 1.0);
+            final double maxFrequency = MathUtils.arrayMax(alleleFractions);
```
takutosato: `maxFrequency` -> `maxFraction`. Sorry for being fussy about allele fraction vs frequency but I get confused when they are used interchangeably

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2FilteringEngine.java
takutosato: Also we should add an integration test for `FilterMutectCalls`. I think this line would've failed with a no such (genotype) field `VCFConstants.ALLELE_FREQUENCY_KEY`.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2FilteringEngine.java
davidbenjamin: done

davidbenjamin: Apparently I get confused, too!

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2FilteringEngine.java
davidbenjamin: `FilterMutectCalls` is invoked within `Mutect2IntegrationTest`.  It doesn't fail because `VCFConstants.ALLELE_FREQUENCY_KEY` is just the string `"AF".  But I agree that this is misleading. . .

davidbenjamin: It looks like there's no good way to sort the `ReadLikelihoods` before annotation, so instead I got rid of the sorting in `SomaticGenotypingEngine`, keeping the original order, and changed `Mutect2FilteringEngine` to not assume that alleles have been sorted by tumor lod.

davidbenjamin:
@takutosato What do you think?

takutosato:
I like it! 👍

takutosato:
Please merge when tests pass

davidbenjamin:
And now for a tricky rebase. . .

PR #930, opened 2017-03-06
crsp sensitivity wdl. Closes #878.

### LINKED ISSUE #878: CRSP sensitivity validation
The CRSP sensitivity pipeline was designed around M1, which outputs a MAF and not a vcf.  For M2, we can rewrite it as pure GATK + wdl.

It can be broken into two parts, generating truth and comparing calls to truth.

## Generating truth
* Given a list of samples from which the pooled bam was sequenced and an interval list, subset the enormous HapMap vcf to just these samples and intervals.  This can be done with `SelectVariants`.
* Remove all indels that are within 20 bp of the nearest other indel.
* Calculate the mixing ratios (it's not simply eg 20%, 20%, . . . for the 5plex pool -- more like 17.5%, 23.4%. . .) of samples as follows: 1) for each sample, find all its singleton het SNPs in the subsampled vcf. 2) divide the total alt read count by the total depth (from the pooled bam's pileup) over all these sites to get an unnormalized guess of the sample's mixing ratio. 3) Normalize these to sum to 1.
*  Annotate each variant with its expected allele fraction in the pool, defined as the sum over samples of the sample's mixing fraction times (0/0.5/1 if the sample is hom ref/het/hom alt).
* Annotate each variant in the HapMap subset with the depth of that locus in the pooled bam.

## Calculating sensitivity
* Run Mutect on the pooled bam (this will be called as a sub-workflow in the wdl).
* Binning by expected allele fraction and depth, divide the number of true positives called by Mutect by the total number of variants in the subsampled HapMap vcf to get the sensitivity for each bin, for SNPs and indels separately.

This is more complex that the specificity validation, but essentially it amounts to writing a few simple `VariantWalker`s.  Since `VariantWalker` is already set up to walk over an auxiliary bam, the above calculations where we walk over the sub-sampled HapMap vcf but need the pooled bam's pileup will not be difficult.

davidbenjamin:
Tested it and it works.  The Python script was adapted from broadinstitute/sock.

@takutosato can you review (not on your sick day, of course)?

### DIFF
--- a/scripts/mutect2_wdl/unsupported/calculate_sensitivity.py
+++ b/scripts/mutect2_wdl/unsupported/calculate_sensitivity.py
@@ -1,0 +1,123 @@
+from __future__ import division
+import numpy as np
+import pandas as pd
+import argparse
+from itertools import chain
+from statsmodels.stats.proportion import proportion_confint
+import matplotlib as mpl
+mpl.use('Agg')
+import matplotlib.pyplot as plt
+
+DEPTH_BINS = [0,50,100,150,200,250,300,350,400,450,500,550,600,650,700,750,800]
+DEPTH_JITTER = 25
+
+SNP_AF_BINS = [0.05, 0.1, 0.2, 0.4, 0.8, 1.0]
+INDEL_AF_BINS = [0.1, 0.2, 1]
+AF_JITTER = 0.05
+EPSILON = 0.0001 #regularizer for 0/0
+
+EXPECTED_AF_COLUMN = 'AF_EXP'
+DEPTH_COLUMN = 'BAM_DEPTH'
+STATUS_COLUMN = 'STATUS' # 1 if called, 0 otherwise
+
+def flatten(x):
+    return list(chain.from_iterable(x))
+
+def generate_sensitivity_table(df, snp_or_indel):
+
+    def assign_bin(x, bins, jitter):
+        """ Assigns 'x' into a bin +/- jitter. """
+        for _bin in bins:
+            lower, upper = _bin - jitter, _bin + jitter
+            if (x > lower and x <= upper):
+                return _bin
+        return np.NaN
+
+    def assign_depth_bin(depth):
+        return assign_bin(depth, DEPTH_BINS, DEPTH_JITTER)
+
+    def assign_snp_af_bin(af):
+        return assign_bin(af, SNP_AF_BINS, AF_JITTER)
+
+    def assign_indel_af_bin(af):
+        if (af > 0 and af <= 0.1):
+            return 0.1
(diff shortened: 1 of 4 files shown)


src/main/java/org/broadinstitute/hellbender/engine/AbstractConcordanceWalker.java
```diff
@@ -70,9 +70,16 @@
     // Overriding the superclass method to favor the truth variants' dictionary
     @Override
     public final SAMSequenceDictionary getBestAvailableSequenceDictionary() {
+        initializeTruthVariantsIfNecessary();
```
davidbenjamin: It's subtle, which is why I only put it in after discovering the null pointer exception.  `GATKTool` calls `getBestAvailableSequenceDictionary` inside `initializeIntervals` from `onStartup`.  Since this always occurs before we initialize truth variants anywhere else, I could just do so inside `getBestAvailableSequenceDictionary` but that's 1) obscure and 2) could break easily if the parent class changed.

scripts/mutect2_wdl/unsupported/calculate_sensitivity.py
```diff
@@ -0,0 +1,121 @@
...
+AF_JITTER = 0.005
+
+EXPECTED_AF_COLUMN = 'AF_EXP'
+DEPTH_COLUMN = 'BAM_DEPTH'
+STATUS_COLUMN = 'STATUS' # 1 if called, 0 otherwise
+
+def flatten(x):
+    return list(chain.from_iterable(x))
+
+def generate_sensitivity_table(df, type):
```
davidbenjamin: Nice catch!  Fixed it.

scripts/mutect2_wdl/unsupported/calculate_sensitivity.py
```diff
@@ -0,0 +1,121 @@
...
+
+    # make a matrix of binned AF vs binned depth where entries are the number of true positives
+    called = pd.pivot_table(df, index = 'depth_bin', columns = 'af_bin',
+                            values = ['is_tp'], aggfunc = sum, fill_value = 0)
+
+    # make a matrix of binned AF vs binned depth where entries are the number of true positives and false negatives
+    total = pd.pivot_table(df, index = 'depth_bin', columns = 'af_bin',
+                           values = [STATUS_COLUMN], aggfunc = len, fill_value = 0)
+
+    sensitivity = called.values / total.values
```
davidbenjamin: done

davidbenjamin: done; some zigzagging remains but less.

PR #931, opened 2017-03-07
Fixes NPE in FilterByOrientationBias when AD field is null

meganshand:


meganshand:
@davidbenjamin Could you please take a look at this?

PR #932, opened 2017-03-08
Simplified M2 isActive, increasing speed and sensitivity. Closes #801.

### LINKED ISSUE #801: Investigate Mutect2 active region determination
We suspect that M2 might have lost sensitivity to SNPs simply because SNPS with low minor allele fraction don't show up as active regions.  This ticket is to understand the current method, come with with a better one, and test it with our automatic validations.

davidbenjamin:
At least on our DREAM challenge integration tests, it turns out that `isActive()` in `Mutect2Engine` was: 1) too permissive of sites including, *some with zero evidence of variation whatsoever*; 2) prone to throwing away good variants; 3) sufficiently complex that the act of determining activity itself, as opposed to the consequence of genotyping too many inactive sites, took a significant amount of time.

This simplification improved sensitivity on DREAM 3 and DREAM 4by a bit while running about three times as fast.  Based on these results, as well as manual review in IGV of several hundred sites, I feel pretty confident in this change.  I am currently running the CRSP sensitivity validation with the old code, so we will have a point of reference.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
@@ -19,7 +19,7 @@ import org.broadinstitute.hellbender.tools.walkers.annotator.VariantAnnotatorEng
 import org.broadinstitute.hellbender.tools.walkers.genotyper.GenotypingOutputMode;
 import org.broadinstitute.hellbender.tools.walkers.haplotypecaller.*;
 import org.broadinstitute.hellbender.tools.walkers.haplotypecaller.readthreading.ReadThreadingAssembler;
-import org.broadinstitute.hellbender.utils.Utils;
+import org.broadinstitute.hellbender.utils.*;
 import org.broadinstitute.hellbender.utils.activityprofile.ActivityProfileState;
 import org.broadinstitute.hellbender.utils.downsampling.AlleleBiasedDownsamplingUtils;
 import org.broadinstitute.hellbender.utils.fasta.CachingIndexedFastaSequenceFile;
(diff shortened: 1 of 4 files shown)


davidbenjamin:
@takutosato Can you review?

takutosato:
Below is the histogram of the number of non-ref reads in sites that Mutect2 considered active, before and after this code change. I used the dream3 bam over a small region in chr 20 from the integration test. 

image

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
```diff
@@ -374,40 +326,37 @@ public ActivityProfileState isActive(final AlignmentContext context, final Refer
...
+        }
+
+        return new ActivityProfileState( refInterval, 1.0, ActivityProfileState.Type.NONE, null);
+    }
+
+    private boolean isNonRef(final byte refBase, final PileupElement p) {
+        return p.getBase() != refBase || p.isDeletion() || p.isBeforeDeletionStart() || p.isAfterDeletionEnd() || p.isBeforeInsertion() || p.isAfterInsertion() || p.isNextToSoftClip();
+    }
+
+    private int countNonRef(byte refBase, AlignmentContext normalContext) {
```
davidbenjamin: done

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
```diff
@@ -374,40 +326,37 @@ public ActivityProfileState isActive(final AlignmentContext context, final Refer
...
+        }
+
+        // since errors are rare, the number of errors (if reads are independent) is approximately a Poisson random variable,
+        // with mean equal to its variance
+        final double expectedTumorNonRefDueToError = StreamSupport.stream(tumorContext.getBasePileup().spliterator(), false)
+                .mapToDouble(pe -> QualityUtils.qualToErrorProb(pe.getQual()))
+                .sum();
+        final double tumorNonRefStdev = Math.sqrt(expectedTumorNonRefDueToError);
+
+        if (tumorNonRef < expectedTumorNonRefDueToError + MTAC.tumorStandardDeviationsThreshold * tumorNonRefStdev) {
```
takutosato: @davidbenjamin actually, should this be expected _minus_ standard deviation?

takutosato:
@davidbenjamin here is the result for all of chromosome 20. Y-axis is on log scale. We have more active sites after the change, especially in higher alt counts. The actual difference is about 7000 more sites, which is about 7000/64mb ~ 100 sites per mb. I think it's a huge win!


image

takutosato:
Also here is the speed comparison. Running single job on gsa5, dream3 over chromosome 20. 

Before: 68.76 minutes
After: 45.82 minutes

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
davidbenjamin: This is as intended -- it's saying that we need to be a few standard deviations above the expected number of alt counts to be called active.  Although you could always set `MTAC.tumorStandardDeviationsThreshold` to be negative. . .

davidbenjamin:
Hmmm, so my guess of a 2-3x speedup based on our integrations tests was overly optimistic, but I'll take it.

davidbenjamin:
@takutosato Thanks for that terrific analysis of active sites!  The more confidence we can have of a change like this, the better.  I could also imagine wanting to reuse this for future experiments.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2Engine.java
takutosato: Yea now I see I read it wrong. Thanks!

takutosato:
No problem! I just pushed my branch ts_active, which has the code I used to get the "before" data. I added similar code to the end of `isActive` in your branch. The gatk4 commands, R script, and the before/after jar are in `/humgen/gsa-hpprojects/dev/tsato/m2-evaluation/active-region`

PR #935, opened 2017-03-13
needed for Jenkins

davidbernick:


davidbernick:
eval test please

scripts/docker/Dockerfile.jenkins
```diff
@@ -0,0 +1,107 @@
+# VERSION 0.2 -- very basic setup of an image.
+#JENKINS expects this file to be copied as Dockerfile to the root of the repo when Jenkins job is run for building
```
LeeTL1220: Hey @davidbernick , If I make a change to the dockerfile for gatk-protected, do I have to repeat the change here?  Why can't you use the gatk-protected Dockerfile?

davidbernick:
I can absolutely make my changes to the Dockerfile. Happy to. I didn't want to do it because I didn't want to break what you had.

davidbernick:
If you're cool with my changes, I'll do another PR with my changes directly in your dockerfile.

davidbernick:
The main difference is my script doesn't do a git clone of itself. If you're cool with my Dockerfile for Jenkins being THE Dockerfile, then I'd prefer to use it.

davidbernick:
eval test please

davidbernick:
@LeeTL1220 feel free it merge this as well as the eval-gatk-protected Jenkins.

LeeTL1220:
@davidbernick Hey, this has failing unit tests.  And looks like an issue with changes you've made (as near as I can tell).

davidbernick:
It is.. this is why I made a separate Dockerfile for Jenkins. I'll change .travis file to match this new Dockerfile (because it assumes a few things).

davidbernick:
@LeeTL1220 i think this is ready to go

.travis.yml
```diff
@@ -43,9 +43,10 @@ script:
     echo "Running ACNV pipeline";
     bash scripts/acnv_cromwell_tests/run_acnv_pipelines.sh;
   elif [[ $TEST_DOCKER == true ]]; then
-    echo "Building docker image with unit tests...";
-    cd scripts/docker/;
-    sudo bash build_docker.sh  -e `git rev-parse ${TRAVIS_BRANCH}` -s;
+    echo "Building docker image with unit tests..." ;
+    echo "${TRAVIS_BRANCH}" ;
+    sudo cp -rfp scripts/docker/Dockerfile . ;
+    sudo docker build -t broadinstitute/gatk-protected:${TRAVIS_BRANCH} --build-arg GITHUB_TAG=${TRAVIS_BRANCH} --build-arg GITHUB_DIR=\  . ;
```
LeeTL1220: Can you continue to use the build_docker script?

.travis.yml
```diff
@@ -43,9 +43,10 @@ script:
     echo "Running ACNV pipeline";
     bash scripts/acnv_cromwell_tests/run_acnv_pipelines.sh;
   elif [[ $TEST_DOCKER == true ]]; then
-    echo "Building docker image with unit tests...";
-    cd scripts/docker/;
-    sudo bash build_docker.sh  -e `git rev-parse ${TRAVIS_BRANCH}` -s;
+    echo "Building docker image with unit tests..." ;
+    echo "${TRAVIS_BRANCH}" ;
+    sudo cp -rfp scripts/docker/Dockerfile . ;
```
LeeTL1220: I don't think sudo is needed (I might be wrong, though)

scripts/docker/Dockerfile.deprecated
```diff
@@ -0,0 +1,107 @@
+# VERSION 0.2 -- very basic setup of an image.
```
LeeTL1220: Do we need this at all?  I am not sure why

.travis.yml
davidbernick: It's not

.travis.yml
davidbernick: Not sure. Does it matter? Do you have downstream stuff? I'll test.

scripts/docker/Dockerfile.deprecated
davidbernick: No I'll remove

.travis.yml
davidbernick: Dockerfiles usually go in the root of the repo. In my Jenkins I just move this to root and run a regular docker build. No script needed. And that's how it should run. Can we just get rid of the docker build script you have and run that way?

davidbernick: Ok I just looked -- then build script is unnecessary because it's all stuff you can do from docker commandline. But given the old Dockerfile that this PR replaces it makes sense because you don't move the Dockerfile to root and thus you git clone the repo before building. If you move the new Dockerfile to root, as suggested, and then run regular docker commands it works. That's the preferred way to work on it, but it's fine to leave the Dockerfile where it is now just with the caveat that you have to move it to root before doing a build. Jenkins does all this in the background automatically. 

So it's your call: if you want to keep using the build script, I have to alter it a bit to keep up with the workflow that is now assumed and works in Jenkins.

This is why I had two different Dockerfiles originally in this PR -- I wanted to support your legacy functions.

davidbernick:
eval test please

davidbernick:
@LeeTL1220 if these are all green, we're good to go in my opinion. I kept your build_docker

.travis.yml
LeeTL1220: Can we just talk for a few minutes in person when you get a chance?

PR #943, opened 2017-03-20
Rev GATK public dependency to 4.alpha.2-184-g88c181d-20170320.193607-1

droazen:


droazen:
@lbergelson please review

droazen:
`HaplotypeCaller` tests are passing, but the `GenotypeGVCF` test cases that formerly used `--stand_emit_conf` are failing. @lbergelson, assigning to you to diagnose the problem, since you are currently in the process of modifying tests for that tool

droazen:
@lbergelson If it turns out that all we need to do here is update the expected outputs, it would be good to get this merged today, since it's blocking your fix in https://github.com/broadinstitute/gatk/pull/2500 from propagating to protected

droazen:
Can you name `standardConfTest.expected.vcf` to include the GATK 3.x version in its name?

PR #953, opened 2017-03-24
Minor changes to Mutect filtering and annotation

davidbenjamin:
* extracts parameters for some filters
* changes which annotations are turned in in M2 by default
* three new annotations  -- Arturo will experiment with these

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/BaseQuality.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/BaseQuality.java
@@ -1,0 +1,76 @@
+package org.broadinstitute.hellbender.tools.walkers.annotator;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.common.primitives.Doubles;
+import htsjdk.variant.variantcontext.Allele;
+import htsjdk.variant.variantcontext.VariantContext;
+import htsjdk.variant.vcf.VCFHeaderLineType;
+import htsjdk.variant.vcf.VCFInfoHeaderLine;
+import org.apache.commons.math3.stat.descriptive.rank.Median;
+import org.broadinstitute.hellbender.engine.ReferenceContext;
+import org.broadinstitute.hellbender.utils.Utils;
+import org.broadinstitute.hellbender.utils.genotyper.ReadLikelihoods;
+import org.broadinstitute.hellbender.utils.read.AlignmentUtils;
+import org.broadinstitute.hellbender.utils.read.GATKRead;
+import org.broadinstitute.hellbender.utils.read.ReadUtils;
+
+import java.util.*;
+
+/**
+ * Created by David Benjamin on 3/20/17.
+ */
+public class BaseQuality extends InfoFieldAnnotation {
+    public static final String REFERENCE_MEDIAN_QUALITY_KEY = "REF_MED_QUAL";
+    public static final String ALT_MEDIAN_QUALITY_KEY = "ALT_MED_QUAL";
+
+    @Override
+    public List<String> getKeyNames() { return Arrays.asList(REFERENCE_MEDIAN_QUALITY_KEY, ALT_MEDIAN_QUALITY_KEY); }
+
+    @Override
+    public Map<String, Object> annotate(final ReferenceContext ref,
+                                        final VariantContext vc,
+                                        final ReadLikelihoods<Allele> likelihoods) {
+
+        final List<Double> refQuals = new ArrayList<>();
+        final List<Double> altQuals = new ArrayList<>();
+
+        final int refLoc = vc.getStart();
+
+        for (final ReadLikelihoods<Allele>.BestAllele bestAllele : likelihoods.bestAlleles()) {
+            final GATKRead read = bestAllele.read;
+            final Allele allele = bestAllele.allele;
+            if (bestAllele.isInformative() && isUsableRead(read, refLoc)) {
+                final OptionalDouble value = getQual(read, refLoc);
+                // Bypass read if the clipping goal is not reached or the refloc is inside a spanning deletion
+                if ( value.isPresent() ) {
+                    (allele.isReference() ? refQuals : altQuals).add(value.getAsDouble());
+                }
+            }
+        }
+
+        final double refMedian = refQuals.size() == 0 ? -1 : new Median().evaluate(Doubles.toArray(refQuals));
+        final double altMedian = altQuals.size() == 0 ? -1 : new Median().evaluate(Doubles.toArray(altQuals));
+
+        return ImmutableMap.of(REFERENCE_MEDIAN_QUALITY_KEY, (int) refMedian, ALT_MEDIAN_QUALITY_KEY, (int) altMedian);
+    }
+
+    private OptionalDouble getQual(final GATKRead read, final int refLoc) {
+        Utils.nonNull(read);
+
+        final int offset = ReadUtils.getReadCoordinateForReferenceCoordinate(ReadUtils.getSoftStart(read), read.getCigar(), refLoc, ReadUtils.ClippingTail.RIGHT_TAIL, true);
+        return offset == ReadUtils.CLIPPING_GOAL_NOT_REACHED || AlignmentUtils.isInsideDeletion(read.getCigar(), offset) ?
+                OptionalDouble.empty() : OptionalDouble.of(read.getBaseQuality(offset));
+    }
+
+    @Override
+    public List<VCFInfoHeaderLine> getDescriptions() {
+        return Arrays.asList(new VCFInfoHeaderLine(REFERENCE_MEDIAN_QUALITY_KEY, 1, VCFHeaderLineType.Integer,
+                        "Median distance of reference alleles to end of read."),
+                new VCFInfoHeaderLine(ALT_MEDIAN_QUALITY_KEY, 1, VCFHeaderLineType.Integer,
+                        "Median distance of alt alleles to end of read."));
+    }
+
+    private boolean isUsableRead(final GATKRead read, final int refLoc) {
+        return read.getMappingQuality() != 0 && !read.isUnmapped() && ReadUtils.getSoftEnd(read) >= refLoc;
+    }
+}
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/ReferenceBases.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/ReferenceBases.java
@@ -1,0 +1,43 @@
+package org.broadinstitute.hellbender.tools.walkers.annotator;
+
+import htsjdk.variant.variantcontext.Allele;
+import htsjdk.variant.variantcontext.VariantContext;
+import htsjdk.variant.vcf.VCFHeaderLineType;
+import htsjdk.variant.vcf.VCFInfoHeaderLine;
+import org.broadinstitute.hellbender.engine.ReferenceContext;
+import org.broadinstitute.hellbender.utils.genotyper.ReadLikelihoods;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Annotate with local reference bases.
+ *
+ * Created by David Benjamin on 3/20/17.
+ */
+public class ReferenceBases extends InfoFieldAnnotation {
+    public static final String REFERENCE_BASES_KEY = "REF_BASES";
+
+    public static final int NUM_BASES_ON_EITHER_SIDE = 10;
+
+    @Override
+    public List<String> getKeyNames() { return Collections.singletonList(REFERENCE_BASES_KEY); }
+
+    @Override
+    public Map<String, Object> annotate(final ReferenceContext ref,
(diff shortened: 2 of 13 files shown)


src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/ReadPositionUnitTest.java
```diff
@@ -0,0 +1,70 @@
...
+        matrix.set(0,1, 10);
+        matrix.set(0,2, 10);
+        matrix.set(0,3, 10);
+        matrix.set(1,4, 10);
+        matrix.set(1,5, 10);
+        matrix.set(1,6, 10);
+
+        final Map<String, Object> annotations = new ReadPosition().annotate(null, vc, likelihoods);
+
+        Assert.assertEquals((int) annotations.get(ReadPosition.REFERENCE_MEDIAN_POSITION_KEY), 2);
```
takutosato: We are not so concerned about the distance to the _nearest_ end of a read. In fact the beginning of the read should be high quality, and thus we should trust non-ref bases there. It's those bases that the sequencer read last that we're concerned about. I think this annotation would be of more value if it measured the distance to the last base read.

takutosato: I will hold off on reviewing `getDistanceFromEnd` until I hear your opinion on this.

src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/BaseQualityUnitTest.java
```diff
@@ -0,0 +1,68 @@
...
+        // variant is a SNP at position 10
+        final int variantSite = 20;
+        final VariantContext vc = new VariantContextBuilder("source", "5", variantSite, variantSite, alleles).make();
+        final org.broadinstitute.hellbender.utils.genotyper.SampleList sampleList = new IndexedSampleList("SAMPLE");
+
+        //7 reads
+        final Map<String, List<GATKRead>> readMap = new LinkedHashMap<>();
+        final List<GATKRead> reads = new ArrayList<>();
+        final byte[] quals = new byte[]{30, 30, 30, 30, 25, 25, 25};
+        for (int r = 0; r < 7; r++) {
```
takutosato: Comments from `ReadPositionUnitTest` apply here too

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/BaseQuality.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Override
+    public List<VCFInfoHeaderLine> getDescriptions() {
+        return Arrays.asList(new VCFInfoHeaderLine(REFERENCE_MEDIAN_QUALITY_KEY, 1, VCFHeaderLineType.Integer,
+                        "Median distance of reference alleles to end of read."),
+                new VCFInfoHeaderLine(ALT_MEDIAN_QUALITY_KEY, 1, VCFHeaderLineType.Integer,
+                        "Median distance of alt alleles to end of read."));
+    }
+
+    private boolean isUsableRead(final GATKRead read, final int refLoc) {
```
takutosato: Ideally we would move `isUsableRead` to a utility class, perhaps to `ReadUtils`. But I see that it's a gatk-public class and the trouble may not be worth it. If you find another way to reduce the duplicated code that would be wonderful, but don't worry about it if it's not trivial.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/BaseQuality.java
```diff
@@ -0,0 +1,76 @@
...
+        final double refMedian = refQuals.size() == 0 ? -1 : new Median().evaluate(Doubles.toArray(refQuals));
+        final double altMedian = altQuals.size() == 0 ? -1 : new Median().evaluate(Doubles.toArray(altQuals));
+
+        return ImmutableMap.of(REFERENCE_MEDIAN_QUALITY_KEY, (int) refMedian, ALT_MEDIAN_QUALITY_KEY, (int) altMedian);
+    }
+
+    private OptionalDouble getQual(final GATKRead read, final int refLoc) {
+        Utils.nonNull(read);
+
+        final int offset = ReadUtils.getReadCoordinateForReferenceCoordinate(ReadUtils.getSoftStart(read), read.getCigar(), refLoc, ReadUtils.ClippingTail.RIGHT_TAIL, true);
```
takutosato: I don't understand this Clipping Right/Left business. It's also getting late and I don't have the energy to step through the nightmare that is `getReadCoordinateForReferenceCoordinate()`. I will just trust that it gives you the correct offset value in the read.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/ReferenceBases.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @Override
+    public Map<String, Object> annotate(final ReferenceContext ref,
+                                        final VariantContext vc,
+                                        final ReadLikelihoods<Allele> likelihoods) {
+
+        final int basesToDiscardInFront = Math.max(vc.getStart() - ref.getWindow().getStart() - NUM_BASES_ON_EITHER_SIDE, 0);
+        final String allBases = new String(ref.getBases());
+        final String localBases = allBases.substring(basesToDiscardInFront, basesToDiscardInFront + 2 * NUM_BASES_ON_EITHER_SIDE);
+        return Collections.singletonMap(REFERENCE_BASES_KEY, localBases );
```
davidbenjamin: I was thinking of using this annotation for plugging into machine learning models for detecting artifacts.  In that case, the idea would be to decide whether the original reference was predisposed to artifacts.  We'll get a better sense of what features we need as we go, but for now the simpler version is defensible.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2FilteringEngine.java
```diff
@@ -48,8 +48,8 @@ private void applyContaminationFilter(final M2FiltersArgumentCollection MTFAC, f
         }
     }
 
-    private static void applyTriallelicFilter(final VariantContext vc, final Collection<String> filters) {
-        if (vc.getNAlleles() > 2) {
+    private void applyTriallelicFilter(final VariantContext vc, final Collection<String> filters) {
+        if (vc.getNAlleles() > (MTFAC.numAltAllelesThreshold + 1)) {
```
davidbenjamin: Becasue `VariantContext` doesn't have a `getNAltAlleles` method, unfortunately.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/BaseQuality.java
davidbenjamin: I'm not sure I understand it either -- I took it from `ReadPosRankSumTest`.  Basically, this is the same as that annotation without emitting a rank sum, which I don't think is the most useful transformation of the data.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/BaseQuality.java
davidbenjamin: Once the annotations, filters etc we use in Mutect stabilize we will need to move a lot of stuff to public, the VCf header lines, for example.  Let's hold off for now.

src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/ReadPositionUnitTest.java
davidbenjamin: I took this from `ReadPositionRankSumTest` and kept it the way it is because I figure that if an artifact occurs only on the late end of a read it will still show up, and if an artifact occurs on the high-quality end for some reason we'll catch it, too.  So basically, if artifacts on the good end do occur, this way makes sense.  But that's an empirical question that I have never looked at.  I will add this to my list of questions for @kcibul.

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/BaseQuality.java
takutosato: Yup, makes sense

src/main/java/org/broadinstitute/hellbender/tools/walkers/annotator/ReferenceBases.java
takutosato: Sounds good!

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2FilteringEngine.java
takutosato: Aha

src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/ReadPositionUnitTest.java
```diff
@@ -0,0 +1,70 @@
...
+        final org.broadinstitute.hellbender.utils.genotyper.SampleList sampleList = new IndexedSampleList("SAMPLE");
+
+        //7 length-12 reads
+        final Map<String,List<GATKRead>> readMap = new LinkedHashMap<>();
+        final List<GATKRead> reads = new ArrayList<>();
+        final int[] positionsOfSiteWithinReads = new int[] {1, 2, 2, 3, 10, 10, 11};
+        final int[] alignmentStarts = Arrays.stream(positionsOfSiteWithinReads).map(n -> variantSite - n).toArray();
+        for (int r = 0; r < 7; r++) {
+            final GATKRead read = ArtificialReadUtils.createArtificialRead(SAM_HEADER,
+                    "RRR00" + r, 0, alignmentStarts[r], "ACGTACGTACGT".getBytes(), new byte[]{30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30, 30}, "12M");
```
davidbenjamin: done

src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/ReadPositionUnitTest.java
```diff
@@ -0,0 +1,70 @@
...
+        final int variantSite = 20;
+        final VariantContext vc = new VariantContextBuilder("source", "5", variantSite, variantSite, alleles).make();
+        final org.broadinstitute.hellbender.utils.genotyper.SampleList sampleList = new IndexedSampleList("SAMPLE");
+
+        //7 length-12 reads
+        final Map<String,List<GATKRead>> readMap = new LinkedHashMap<>();
+        final List<GATKRead> reads = new ArrayList<>();
+        final int[] positionsOfSiteWithinReads = new int[] {1, 2, 2, 3, 10, 10, 11};
+        final int[] alignmentStarts = Arrays.stream(positionsOfSiteWithinReads).map(n -> variantSite - n).toArray();
+        for (int r = 0; r < 7; r++) {
```
davidbenjamin: done

src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/ReadPositionUnitTest.java
```diff
@@ -0,0 +1,70 @@
...
+    public void test() {
+        final SAMFileHeader SAM_HEADER = ArtificialReadUtils.createArtificialSamHeader(10, 0, 1000);
+
+        final List<Allele> alleles = Arrays.asList(Allele.create((byte) 'A', true), Allele.create((byte) 'C', false));
+        final AlleleList<Allele> alleleList = new IndexedAlleleList<>(alleles);
+
+        // variant is a SNP at position 10
+        final int variantSite = 20;
+        final VariantContext vc = new VariantContextBuilder("source", "5", variantSite, variantSite, alleles).make();
+        final org.broadinstitute.hellbender.utils.genotyper.SampleList sampleList = new IndexedSampleList("SAMPLE");
```
davidbenjamin: done

src/test/java/org/broadinstitute/hellbender/tools/walkers/annotator/ReadPositionUnitTest.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+public class ReadPositionUnitTest extends BaseTest {
+    @Test
+    public void test() {
+        final SAMFileHeader SAM_HEADER = ArtificialReadUtils.createArtificialSamHeader(10, 0, 1000);
+
+        final List<Allele> alleles = Arrays.asList(Allele.create((byte) 'A', true), Allele.create((byte) 'C', false));
+        final AlleleList<Allele> alleleList = new IndexedAlleleList<>(alleles);
+
+        // variant is a SNP at position 10
```
davidbenjamin: done

davidbenjamin:
@takutosato I made all suggested changes.  I'm meeting with Kristian tomorrow and will ask his opinion on the distance from any end vs distance from the later end.

takutosato:
Looks good. Let's discuss the end position of a read after you meet with Kristian.

PR #954, opened 2017-03-24
fixing headers in README

lbergelson:
Github changed how it's markdown parser handles headers.  It used to accept `#SomeHeader` but now requires a space, i.e. `# SomeHeader`.

Updating to the readme to match the new markdown syntax.


lbergelson:
@LeeTL1220 This is an easy one here....

PR #962, opened 2017-03-29
Made changes to joint segmentation and added temporary AllelicCNVHMM CLI.

samuelklee:
Code in this branch is work-in-progress and should be reviewed accordingly.

Changes by @davidbenjamin 
-changes to docs
-got rid of weights and concentration
-got rid of constant states to simplify before CRP pre-training
-smarter transition matrix
-switched to binomial AF likelihoods for segmentation
-got rid of attempt big change in memory length
-fixed outlier likelihood
Changes by @samuelklee
-ACNV with joint segmentation
-tweaked convergence criteria and removed extraneous MCMC fit
-sorted acc in AF segmentation
-NaN fixes in binomial likelihood
-fixed some tests and added EXPERIMENTAL tags
-disabled JointAFCRSegmenterUnitTest


samuelklee:
@mbabadi, could you review?  This includes changes by both myself and @davidbenjamin.

This introduces a new command line (AllelicCNVHMM---@sooheelee, this command line is experimental and should not be used) that performs joint segmentation and then fits model parameters using MCMC.  It performs relatively well on some samples (and was used to generate results for the AACR poster), but others result in oversegmentation and convergence issues.  It's possible that this could be due to the naive copy-ratio model used.  @davidbenjamin may want to do some additional tweaking, but I think we will also explore other iHMM variants concurrently.

droazen:
@samuelklee Migration instructions for this branch: https://github.com/broadinstitute/gatk/wiki/Migrating-branches-from-gatk-protected-to-gatk

PR #967, opened 2017-03-31
Filter reads before assembly in Mutect. Closes #723.

### LINKED ISSUE #723: Should we filter reads before assembly?
HaplotypeCaller and Mutect filter reads (due to length < 10 or mapping quality < 20 or unmapped mate or mate on different contig) after assembly.  This seems odd.  Why would we assemble reads we don't trust?

davidbenjamin:
@takutosato You saw the improvement in sensitivity on DREAM.  HapMap showed the same.  Normal-normal false positives were about the same.  Most of the changes are straightforward.  The stuff about `perSampledFilteredReadList` is that we used to keep track of filtered reads in order to annotate based on those.  1) I don't think that makes much sense.  2) We can't do that any longer because those reads are filtered before `Mutect2Engine` see them.

@LeeTL1220 You might be interested in knowing that this improved sensitivity in all the DREAM challenges by about 3%.

@droazen @yfarjoun @ldgauthier HaplotypeCaller has the same comment as Mutect asking whether this change is worthwhile.  Might be worth investigating making the change there, too.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/M2ArgumentCollection.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/M2ArgumentCollection.java
@@ -85,12 +85,6 @@ public class M2ArgumentCollection extends AssemblyBasedCallerArgumentCollection 
     public float STRAND_ARTIFACT_LOD_THRESHOLD = 2.0f;
 
     /**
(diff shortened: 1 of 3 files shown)


davidbenjamin:
@takutosato My bad for letting this fester with a failing test for a while.  I re-ran the job and it was just a Travis thing.

takutosato:
@davidbenjamin sorry for the delay; it looks good.

PR #973, opened 2017-04-04
Move to a snapshot of gatk-public based on 4.alpha.2-188-g7332d10 with TandemRepeat fix applied

droazen:
This moves us to a snapshot of gatk-public that takes the version gatk-protected
currently depends on (4.alpha.2-188-g7332d10) and applies JUST @davidbenjamin 's fix
to the TandemRepeat annotation to it.

This is necessary to unblock @davidbenjamin 's work, because the HaplotypeCaller
tests are failing if we update protected to the latest public HEAD, and although we've
fixed some of the issues there are still some unexplained failures in the concordance tests.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -77,7 +77,7 @@ build.dependsOn installDist
 check.dependsOn installDist
 
 dependencies {
-    compile 'org.broadinstitute:gatk:4.alpha.2-188-g7332d10-20170321.172613-1'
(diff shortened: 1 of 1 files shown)


droazen:
Tests are failing due to some (hopefully temporary) git lfs issues that we've seen on other branches as well today. Will check again tomorrow morning to see if they've cleared up.

davidbenjamin:
I figured that was the case.  One of the failing tests passed when I restarted the job.

Thanks for doing this @droazen!

PR #974, opened 2017-04-05
Use more sub-workflows in M2 WDL

davidbenjamin:
@LeeTL1220 These changes let us run all the hapmap multiplexes and all the dream challenges from a single cromwell command.  They are tested and currently in use in my validations.

### DIFF
--- a/scripts/mutect2_wdl/mutect2.wdl
+++ b/scripts/mutect2_wdl/mutect2.wdl
@@ -22,12 +22,160 @@
 #   independent of what is in the docker file.  See the README.md for more info.
 #
 
+
+workflow Mutect2 {
+  # gatk4_jar needs to be a String input to the workflow in order to work in a Docker image
+  String gatk4_jar
+  File intervals
+  File ref_fasta
+  File ref_fasta_index
+  File ref_dict
+  File tumor_bam
+  File tumor_bam_index
+  String tumor_sample_name
+  File? normal_bam
+  File? normal_bam_index
+  String? normal_sample_name
+  File? pon
+  File? pon_index
+  Int scatter_count
+  File? dbsnp
+  File? dbsnp_index
+  File? cosmic
+  File? cosmic_index
+  File? variants_for_contamination
+  File? variants_for_contamination_index
+  Boolean is_run_orientation_bias_filter
+  Boolean is_run_oncotator
+  String m2_docker
+  String oncotator_docker
+  File? gatk4_jar_override
+  Int preemptible_attempts
+  File? onco_ds_tar_gz
+  String? onco_ds_local_db_dir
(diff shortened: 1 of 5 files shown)


scripts/mutect2_wdl/mutect2.wdl
```diff
@@ -80,48 +80,13 @@ task M2 {
   }
 }
 
-task CalculateContamination {
```
LeeTL1220: Why are you making any changes in mutect2.wdl?    How is it okay to remove the contamination calculation from mutect WDL?

Please explain....

LeeTL1220:
@davidbenjamin This PR confuses me.

scripts/mutect2_wdl/mutect2.wdl
davidbenjamin: I put it into the filtering task.

scripts/mutect2_wdl/mutect2.wdl
```diff
@@ -201,8 +168,27 @@ task Filter {
         GATK_JAR=${gatk4_jar_override}
     fi
 
-  	java -Xmx4g -jar $GATK_JAR FilterMutectCalls -V ${unfiltered_calls} -O ${output_prepend}-filtered.vcf \
-       ${"-contaminationTable " + contamination_table}
+    if [[ "${variants_for_contamination}" == *.vcf ]]; then
+        java -Xmx4g -jar $GATK_JAR GetPileupSummaries -I ${tumor_bam} -L ${intervals} -V ${variants_for_contamination} -O pileups.table
```
LeeTL1220: I can be convinced that combining these into one WDL task is a good idea .... But here are my concerns:

1)  At some point, we are going to want to allow users to specify a sequencing artifact metric file (optional) so that they do not have to rerun this task.  

2)  How long do these tasks take?  The SequencingArtifactMetrics can take a while.  I worry about failures costing a lot in terms of lost time.  

3)  Some of these can run in parallel, which save time.

What're your thoughts?

LeeTL1220:
@davidbenjamin Also, maybe I missed something, but combining commands into single WDL tasks is not really using sub-workflows.  Anyway, back to you...

davidbenjamin:
@LeeTL1220 The sub-workflows part of this PR is in the changes to the concordance and Hapmap wdls.

scripts/mutect2_wdl/mutect2.wdl
davidbenjamin: I think that a good compromise would be to put Sequencing artifacts as its own task but leave the rest, which are very quick, in one task.

LeeTL1220:
I like that.  Feel free to merge after

PR #976, opened 2017-04-05
Rev GATK dependency to 4.alpha.2-214-g09b380e-20170405.195304-1

droazen:
* Includes a fix to GenotypingEngine in public necessary to get the
  HaplotypeCaller tests passing with the latest gatk-public

* Updated HaplotypeCallerIntegrationTest to use "-stand_call_conf 30.0"
  for now, so that we don't have to deal with the migration to a default
  stand_call_conf of 10.0 just yet.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -77,7 +77,7 @@ build.dependsOn installDist
 check.dependsOn installDist
 
 dependencies {
-    compile 'org.broadinstitute:gatk:4.alpha.2-189-g056f196-20170404.231309-1'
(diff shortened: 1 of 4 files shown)


droazen:
@lbergelson Could you push one additional commit into this branch to fix the GenotypeGVCFs tests?

lbergelson:
@droazen update pushed

droazen:
Thanks @lbergelson -- I've pushed one (hopefully final) change to `HaplotypeCallerSparkIntegrationTest`. This should pass tests now.

Issue #580, opened 2016-06-21
enable --includeNonVariantSites GenotypeGVCS option


lbergelson:
Update genotype gvcfs so includeNonVariantSites can be enabled.  This may be easiest by implementing a different walker type to match the old LocusWalker behavoir but could also be done by accumulating sites as it goes.

lbergelson:
this is blocked by  broadinstitute/gatk#2429

droazen:
Issue moved to [broadinstitute/gatk #2865](https://github.com/broadinstitute/gatk/issues/2865) via [**ZenHub**](https://www.zenhub.com/)

PR #978, opened 2017-04-06
genotypeGVCFs update

lbergelson:
GenotypeGVCFs in pretty good shape now. It matches gatk3 closely

**minor differences that shouldn't be problematic:**
*  rawMQ is fixed with a hack that doesn't fix the rest of the reducible annotations, we don't use any others by default though so it should be ok
*   inbreeding coefficient is very close to gatk3
      there still seems to be some difference but its very highly correlated and I think it's likely because of precision issues
* there's some jitter in qual scores compared to gatk3 but as far as I can tell it's precision issues and not significant
* QD is still randomized so it's hard to compare to gatk3
sites with a large difference
* there are rare differences in the genotypes having to do with how gatk4 trims excess alleles vs gatk3
 gatk4 seems to be doing it better

**potential problems**
* there's a rare condition where we output a RankSumTest annotation with a small negative value while gatk3 puts out no rank sum annotation at the site
* FS seems to have some issue still at certain sites.
      The gatk3 fix for FS resolve almost all of the differences so it's wierd that we still see occasional large differences.  It doesn't seem to be numerical precision issues, somethings wrong with either gatk3 or gatk4's calculation
 
The first two commits were clean applications of changes made to gatk3 between the time I ported genotypeGVCFs.  The last commit is a blob of changes which were necessary to get things to match.


droazen:
@lbergelson This branch has test failures that seem unrelated to the recent travis flakiness. Eg.,

```
org.broadinstitute.hellbender.exceptions.UserException$MissingReference: The specified fasta file (/Users/louisb/data/Homo_sapiens_assembly19.fasta) does not exist.
```

```
org.broadinstitute.hellbender.exceptions.UserException$CouldNotReadInputFile: Couldn't read file file:///home/travis/build/broadinstitute/gatk-protected/src/test/resources/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFs/CEUTrio.20.21.missingIndel.g.vcf. Error was: It doesn't exist.
```

```
Gradle suite > Gradle test > org.broadinstitute.hellbender.tools.walkers.GenotypeGVCFsIntegrationTest.testEntireVariantContext[6](src/test/resources/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFs/combined_genotype_gvcf_exception.nocall.vcf, src/test/resources/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFs/combined_genotype_gvcf_exception.output.vcf, [], /home/travis/build/broadinstitute/gatk-protected/src/test/resources/large/human_g1k_v37.20.21.fasta) FAILED
    java.lang.AssertionError: sample names set expected [true] but found [false]
```

etc etc.

Can you fix these?

droazen:
I'll do a review pass on the changes to the tool now while you're working on the test failures.

src/main/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFs.java
```diff
@@ -169,36 +187,79 @@ public void apply(VariantContext vc, ReadsContext reads, ReferenceContext ref, F
...
+     *
+     * @param vc  the VariantContext to evaluate
+     * @return true if it has proper alternate alleles, false otherwise
+     */
+    private boolean isProperlyPolymorphic(final VariantContext vc) {
+        return ( vc != null &&
+                !vc.getAlternateAlleles().isEmpty() &&
+                (!vc.isBiallelic() ||
+                        (!vc.getAlternateAllele(0).equals(Allele.SPAN_DEL) &&
+                                !vc.getAlternateAllele(0).equals(GATKVCFConstants.SPANNING_DELETION_SYMBOLIC_ALLELE_DEPRECATED))
```
droazen: There is an extra `vc.isSymbolic()` call in the GATK3 equivalent of this boolean expression -- did you intend to leave it out?

Can you please make this method static and `@VisibleForTesting`, and add an exhaustive unit test for it?

src/main/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFs.java
```diff
@@ -169,36 +187,79 @@ public void apply(VariantContext vc, ReadsContext reads, ReferenceContext ref, F
     private VariantContext regenotypeVC(final VariantContext originalVC, final ReferenceContext ref, final FeatureContext features, boolean includeNonVariants) {
         Utils.nonNull(originalVC);
 
-        VariantContext result = originalVC;
```
droazen: Can you confirm that all cases in the previous code that would have ultimately resulted in `result == originalVC` are preserved?

src/main/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFs.java
```diff
@@ -169,36 +187,79 @@ public void apply(VariantContext vc, ReadsContext reads, ReferenceContext ref, F
...
+    private VariantContext calculateGenotypes(VariantContext vc){
+
+        final VariantContext.Type type = vc.getType();
+        final GenotypeLikelihoodsCalculationModel model;
+        /*
+         * Query the VariantContext for the appropriate model.  If type == MIXED, one would want to use model = BOTH.
+         * However GenotypingEngine.getAlleleFrequencyPriors throws an exception if you give it anything but a SNP or INDEL model.
+         */
+        if ( type == VariantContext.Type.INDEL ) {
+            model = GenotypeLikelihoodsCalculationModel.INDEL;
```
droazen: Erm, can you confirm that the INDEL model is functional in this minimal version of the genotyping engine? Do we have a representative sampling of indels in our test inputs?

src/main/java/org/broadinstitute/hellbender/tools/walkers/ReferenceConfidenceVariantContextMerger.java
```diff
@@ -194,18 +201,18 @@ private VCWithNewAlleles(final VariantContext vc, final List<Allele> newAlleles,
             this.isSpanningEvent = isSpanningEvent;
         }
 
-        public List<Allele> filterAllelesForFinalSet() {
+        public Stream<Allele> filterAllelesForFinalSet() {
```
droazen: It seems weird to change this public method to return a `Stream` rather than a more conventional `List` -- and if you're only doing this to make your streaming expressions below infinitesimally more elegant then I particularly disapprove :) Does this method have to be public in the first place?

src/main/java/org/broadinstitute/hellbender/tools/walkers/ReferenceConfidenceVariantContextMerger.java
```diff
@@ -194,18 +201,18 @@ private VCWithNewAlleles(final VariantContext vc, final List<Allele> newAlleles,
...
                     .filter(a -> !a.isReference())
                     .filter(a -> !(a.isSymbolic() && vc.isSymbolic())) // skip <*DEL> if there isn't a real alternate allele.
-                    .filter(Allele::isCalled) // skip NO_CALL
-                    .collect(Collectors.toList());
+                    .filter(Allele::isCalled) ; // skip NO_CALL
         }
 
         // record whether it's also a spanning deletion/event (we know this because the VariantContext type is no
         // longer "symbolic" but "mixed" because there are real alleles mixed in with the symbolic non-ref allele)
         public boolean isSpanningDeletion() {
```
droazen: Can this method be independently unit tested? Given that the criteria have changed it seems like a good idea..

src/test/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFsIntegrationTest.java
```diff
@@ -15,20 +19,24 @@
...
 import java.util.function.BiConsumer;
+import java.util.stream.Collectors;
 
 public class GenotypeGVCFsIntegrationTest extends CommandLineProgramTest {
 
-    //This value was originally 30 but was changed to 10.  We're setting most tests to use 30 to temporarily avoid having
-    //to update the expected outputs
-    private static final List<String> STAND_CALL_CONF_30 = Arrays.asList("-stand_call_conf", "30");
+    private static final List<String> NO_EXTRA_ARGS = Collections.emptyList();
+    private static final String b38_reference_20_21 =largeFileTestDir + "Homo_sapiens_assembly38.20.21.fasta";
```
droazen: Hmm, ideally this would be a constant in `BaseTest` in public. Can we check this reference into public as well (if it's not already), and add the constant to `BaseTest`? (not a blocker for merging this PR though)

src/test/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFsIntegrationTest.java
```diff
@@ -38,47 +46,101 @@
...
-                {basePairGVCF, "maxAltAllelesTest.expected.vcf", Arrays.asList("--maxAltAlleles", "1", "-stand_call_conf", "30")}, //restricting the max number of alt alleles
-                {basePairGVCF, "standardConfTest.expected.vcf", Arrays.asList("-stand_call_conf", "300")}, //changing call confidence
-                {"spanningDel.combined.g.vcf", "spanningDel.combined.g.vcf.expected.vcf", STAND_CALL_CONF_30},
-                {"spanningDel.delOnly.g.vcf", "spanningDel.delOnly.g.vcf.expected.vcf", STAND_CALL_CONF_30},
-                {"CEUTrio.20.21.gatk3.4.g.vcf", "CEUTrio.20.21.expected.vcf", Arrays.asList("--dbsnp", "src/test/resources/large/dbsnp_138.b37.20.21.vcf", "-stand_call_conf", "30")},
-                //{basePairGVCF, "gvcf.basepairResolution.includeNonVariantSites.expected.vcf", Collections.singletonList("--includeNonVariantSites")
+//                {"combine.single.sample.pipeline.1.vcf", null, Arrays.asList("-V", getTestFile("combine.single.sample.pipeline.2.vcf").toString() , "-V", getTestFile("combine.single.sample.pipeline.3.vcf").toString()), b37_reference_20_21}, //combine not supported yet
```
droazen: This TODO should be captured in a ticket on github

src/test/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFsIntegrationTest.java
```diff
@@ -38,47 +46,101 @@
...
+                {getTestFile("CEUTrio.20.21.gatk3.4.g.vcf"), getTestFile( "CEUTrio.20.21.expected.vcf"), Arrays.asList("--dbsnp", "src/test/resources/large/dbsnp_138.b37.20.21.vcf"), b37_reference_20_21},
+                {getTestFile("CEUTrio.20.21.missingIndel.g.vcf"), getTestFile( "CEUTrio.20.21.missingIndel.expected.vcf"), Arrays.asList("--dbsnp", "src/test/resources/large/dbsnp_138.b37.20.21.vcf"), b37_reference_20_21},
+                {new File(largeFileTestDir + "gvcfs/cutDown.24_sample.21.g.vcf"), new File( largeFileTestDir + "gvcfs/cutDown.24_sample.21.expected.vcf"), NO_EXTRA_ARGS, b38_reference_20_21},
+                {getTestFile("chr21.bad.pl.g.vcf"), getTestFile( "chr21.bad.pl.expected.vcf"), Arrays.asList("-L", "chr21:28341770-28341790"), b38_reference_20_21},
+                {new File(largeFileTestDir + "gvcfs/combined.gatk3.7.g.vcf.gz"),  new File(largeFileTestDir + "gvcfs/combined.gatk3.7.expected.vcf.gz"), NO_EXTRA_ARGS, b38_reference_20_21}
+                //{getTestFile(basePairGVCF), getTestFile( "gvcf.basepairResolution.includeNonVariantSites.expected.vcf"), Collections.singletonList("--includeNonVariantSites") //allsites not supported yet
```
droazen: Capture TODO in a github ticket

src/test/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFsIntegrationTest.java
```diff
@@ -38,47 +46,101 @@
...
+                {getTestFile("spanningDel.depr.delOnly.g.vcf"), getTestFile( "spanningDel.depr.delOnly.g.vcf.expected.vcf" ), NO_EXTRA_ARGS, b37_reference_20_21},
+                {getTestFile("ad-bug-input.vcf"), getTestFile( "ad-bug-output.vcf"), NO_EXTRA_ARGS, b37_reference_20_21}, //Bad AD Propagation Haploid Bug
+                {getTestFile("CEUTrio.20.21.gatk3.4.g.vcf"), getTestFile( "CEUTrio.20.21.expected.vcf"), Arrays.asList("--dbsnp", "src/test/resources/large/dbsnp_138.b37.20.21.vcf"), b37_reference_20_21},
+                {getTestFile("CEUTrio.20.21.missingIndel.g.vcf"), getTestFile( "CEUTrio.20.21.missingIndel.expected.vcf"), Arrays.asList("--dbsnp", "src/test/resources/large/dbsnp_138.b37.20.21.vcf"), b37_reference_20_21},
+                {new File(largeFileTestDir + "gvcfs/cutDown.24_sample.21.g.vcf"), new File( largeFileTestDir + "gvcfs/cutDown.24_sample.21.expected.vcf"), NO_EXTRA_ARGS, b38_reference_20_21},
+                {getTestFile("chr21.bad.pl.g.vcf"), getTestFile( "chr21.bad.pl.expected.vcf"), Arrays.asList("-L", "chr21:28341770-28341790"), b38_reference_20_21},
```
droazen: Some expected outputs here are versioned (as, eg., coming from a particular version of GATK 3.x) while others are not. Where did the unversioned ones come from, and can we record their provenance in their names?

src/test/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFsIntegrationTest.java
```diff
@@ -38,47 +46,101 @@
...
+                {new File(largeFileTestDir + "gvcfs/combined.gatk3.7.g.vcf.gz"),  new File(largeFileTestDir + "gvcfs/combined.gatk3.7.expected.vcf.gz"), NO_EXTRA_ARGS, b38_reference_20_21}
+                //{getTestFile(basePairGVCF), getTestFile( "gvcf.basepairResolution.includeNonVariantSites.expected.vcf"), Collections.singletonList("--includeNonVariantSites") //allsites not supported yet
         };
     }
 
+
+    @Test(dataProvider = "gvcfsToGenotype", enabled = false) // this is useful for development
```
droazen: If this test is only for development and should always be disabled under normal circumstances, can you state this more clearly, and provide some instructions on how to use it?

src/test/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFsIntegrationTest.java
```diff
@@ -38,47 +46,101 @@
...
+            Files.createDirectory(expectedResultsDir);
+        }
+        Files.copy(gatk3Result.toPath(), expectedResultsDir.resolve(outputFile.getName()), StandardCopyOption.REPLACE_EXISTING);
+
+        assertGenotypesMatch(input, gatk3Result, extraArgs, reference);
+        assertVariantContextsMatch(input, gatk3Result, extraArgs, reference);
+    }
+
+
+    private static void runProcess(ProcessController processController, String[] command) {
```
droazen: Surely we have an existing utility method somewhere that does this?

src/test/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFsIntegrationTest.java
```diff
@@ -38,47 +46,101 @@
...
+        assertVariantContextsMatch(input, expected, extraArgs, reference);
     }
 
-    private void assertVariantContextsMatch(File input, File expected, List<String> extraArgs) throws IOException {
+    private void assertVariantContextsMatch(File input, File expected, List<String> extraArgs, String reference) throws IOException {
         runGenotypeGVCFSAndAssertSomething(input, expected, extraArgs, (a, e) -> VariantContextTestUtils.assertVariantContextsAreEqual(a, e,
-                Arrays.asList("FS", //TODO There's a bug in GATK 3 computing FS
-                        "QD", //TODO QD has a cap value and anything that reaches that is randomized.  It's difficult to reproduce the same random numbers accross gatk3 -> 4
-                        "InbreedingCoeff")));
+                Arrays.asList("QD",//TODO QD has a cap value and anything that reaches that is randomized.  It's difficult to reproduce the same random numbers accross gatk3 -> 4
```
droazen: Add comment explaining the purpose of this scary-looking list (eg., are these annotations excluded from the comparison?)

src/test/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFsIntegrationTest.java
```diff
@@ -38,47 +46,101 @@
...
-    private void runGenotypeGVCFSAndAssertSomething(File input, File expected, List<String> additionalArguments, BiConsumer<VariantContext, VariantContext> assertion) throws IOException {
-        final File output = createTempFile("genotypegvcf", "vcf");
+    private void runGenotypeGVCFSAndAssertSomething(File input, File expected, List<String> additionalArguments, BiConsumer<VariantContext, VariantContext> assertion, String reference) throws IOException {
+        final File output = createTempFile("genotypegvcf", ".vcf");
 
         final ArgumentsBuilder args = new ArgumentsBuilder();
-        args.addReference(new File(b37_reference_20_21))
+        args.addReference(new File(reference))
                 .addVCF(input)
                 .addOutput(output);
```
droazen: There is a TODO below in a comment: "TODO Inbreeding calculation changed between 3.4 and now"  Is this still valid? Did you fix this?

src/test/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFsUnitTest.java
```diff
@@ -121,9 +122,17 @@ public void testPhasingUpdate(){
         Assert.assertEquals(homVarAfterCleanup.getAnyAttribute(GATKVCFConstants.HAPLOTYPE_CALLER_PHASING_GT_KEY), GenotypeGVCFs.PHASED_HOM_VAR_STRING);
         final Genotype homRefAfterCleaning = GenotypeGVCFs.cleanupGenotypeAnnotations(withPhasing, true).get(1);
         Assert.assertEquals(homRefAfterCleaning.getAnyAttribute(GATKVCFConstants.HAPLOTYPE_CALLER_PHASING_GT_KEY), "something");
-
     }
 
-
+    @Test
+    public void testRGQ0IsNoCall(){
```
droazen: `RGQ0`? What does the `R` stand for?

src/test/resources/large/.gitattributes
```diff
@@ -27,3 +27,9 @@ mutect/dream_synthetic_bams/normal_4.bam filter=lfs diff=lfs merge=lfs -text
 mutect/dream_synthetic_bams/tumor_3.bam filter=lfs diff=lfs merge=lfs -text
 mutect/dream_synthetic_bams/tumor_4.bam filter=lfs diff=lfs merge=lfs -text
 ACNV_pipeline_test_files/human_g1k_v37.chr-20.truncated.fasta filter=lfs diff=lfs merge=lfs -text
+Homo_sapiens_assembly38.20.21.fasta filter=lfs diff=lfs merge=lfs -text
+gvcfs/cutDown.24_sample.21.g.vcf filter=lfs diff=lfs merge=lfs -text
+gvcfs/combined.gatk3.7.g.vcf.gz filter=lfs diff=lfs merge=lfs -text
+gvcfs/combined.gatk3.7.g.vcf.gz.tbi filter=lfs diff=lfs merge=lfs -text
+gvcfs/combined.gatk3.7.expected.vcf.gz filter=lfs diff=lfs merge=lfs -text
```
droazen: There are a number of files with "gatk3.7" in their names -- were these generated with the release version of GATK 3.7, or with a development version? If the latter, can you put the full version (with git hash) in the file names?

src/main/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFs.java
```diff
@@ -74,8 +72,6 @@
             doc="File to which variants should be written", optional=false)
     private File outputFile;
 
-    private static String GVCF_BLOCK = "GVCFBlock";
```
lbergelson: At some point I had made the change to expose it in public, but then decided it wasn't worth making  separate pr to public and inlined it and forgot to reextract the constant.  Will fix..

src/main/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFs.java
```diff
@@ -87,8 +83,23 @@
      * Which annotations to recompute for the combined output VCF file.
      */
     @Advanced
```
lbergelson: this one was advanced, annotations to exclude didn't exist

src/main/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFs.java
```diff
@@ -87,8 +83,23 @@
...
+
+
+    @Advanced
+    @Argument(fullName="annotationsToExclude", shortName="AX", doc="One or more specific annotations to exclude from recomputation.", optional=true)
+    List<String> annotationsToExclude = new ArrayList<>();
+
+    /**
+     * Which groups of annotations to add to the output VCF file. The single value 'none' removes the default group. See
+     * the VariantAnnotator -list argument to view available groups. Note that this usage is not recommended because
+     * it obscures the specific requirements of individual annotations. Any requirements that are not met (e.g. failing
```
lbergelson: Literally a copy and paste of the comment in gatk3.  Looks like it needs amending though.  I'm not sure "none" works.  And we don't have VariantAnnotator...

src/main/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFs.java
lbergelson: fixed

src/main/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFs.java
lbergelson: updated comment

src/main/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFs.java
```diff
@@ -105,6 +116,11 @@
 
     private VariantContextWriter vcfWriter;
 
+    //todo: remove this when the reducible annotation framework is in place
+    //this is ok because RMS_MAPPING is stateless, but it needs to be instantiated in order to call its
+    //overridden instance methods
+    private static final RMSMappingQuality RMS_MAPPING_QUALITY = new RMSMappingQuality();
```
lbergelson: can't override static methods :(  We could probably fix that problem if we move to scala...

src/main/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFs.java
```diff
@@ -116,26 +132,28 @@ public void onTraversalStart() {
         final SampleList samples = new IndexedSampleList(inputVCFHeader.getGenotypeSamples()); //todo should this be getSampleNamesInOrder?
 
         genotypingEngine = new MinimalGenotypingEngine(createUAC(), samples, new GeneralPloidyFailOverAFCalculatorProvider(genotypeArgs));
-        annotationEngine = VariantAnnotatorEngine.ofSelectedMinusExcluded(Collections.emptyList(), annotationsToUse, Collections.<String>emptyList(), dbsnp.dbsnp, Collections.emptyList());
+
+        annotationEngine = VariantAnnotatorEngine.ofSelectedMinusExcluded(annotationGroupsToUse, annotationsToUse, annotationsToExclude, dbsnp.dbsnp, Collections.emptyList());
```
lbergelson: yes, and it even fails if you take dbsnp away!

src/main/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFs.java
```diff
@@ -147,14 +165,14 @@ private void setupVCFWriter(VCFHeader inputVCFHeader, SampleList samples) {
...
+        final VCFHeader vcfHeader = new VCFHeader(headerLines, new TreeSet<>(sampleNameSet));
         vcfWriter.writeHeader(vcfHeader);
     }
 
     @Override
-    public void apply(VariantContext vc, ReadsContext reads, ReferenceContext ref, FeatureContext features ) {
-        ref.setWindow(10,10);
-        final VariantContext mergedVC = merger.merge(Collections.singletonList(vc), vc, includeNonVariants ? ref.getBase() : null, true, false);
+    public void apply(VariantContext variant, ReadsContext reads, ReferenceContext ref, FeatureContext features) {
+        ref.setWindow(10, 10);
```
lbergelson: I'm not sure why it's that way... I just copied it from gatk3.  I can check the annotations though.

lbergelson: It looks like it's only actually used by `TandemRepeat` which is smart about getting the right base.

lbergelson: It's seems very possible that we could remove the setWindow, but I'd rather leave it for now and reevaluate later.

src/main/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFs.java
lbergelson: We definitely have indels in our test inputs.  I think they're representative, but maybe not exhaustive. As far as I can tell, the only thing the model is used for in genotypeGVCFs is in that it's passed through to `getAlleleFrequencyPriors` which is a method on `GenotypingEngine` that was ported to gatk4. As far as I can see everythign we need is here.  The model seems to be used in a number of methods on UnifiedGenotypingEngine that we DON'T call from GenotypeGVCFs.

src/main/java/org/broadinstitute/hellbender/tools/walkers/ReferenceConfidenceVariantContextMerger.java
lbergelson: So this is a public method on a private inner class.  The scope is really irrelevant because java allows the outer class to reach into the inner class private methods anyway, but I figured it's better to still signal which are internal to the class and which are meant to be accessed by the outer class.  That's why it's public.

I changed it to avoid materializing the list at the end of the call because it seemed like a wasteful allocation.  It's a micro-optimization but since the only use case is in a stream operation in the outer class it seems silly to change it back.

Let me  know what you want me to do.  I think it's fine how it is, but I'm willing to change it if you prefer.

src/main/java/org/broadinstitute/hellbender/tools/walkers/ReferenceConfidenceVariantContextMerger.java
lbergelson: I can make the inner class package protected in order to test this I guess.  Will do.

lbergelson: I've added an exhaustive test. This method is now better tested than the majority of the genotyping engine.

src/test/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFsIntegrationTest.java
lbergelson: added a ticket https://github.com/broadinstitute/gatk/issues/2583

src/test/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFsIntegrationTest.java
lbergelson: this is https://github.com/broadinstitute/gatk/issues/2429

lbergelson: opened to be more explicit about it https://github.com/broadinstitute/gatk/issues/2584

src/test/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFsIntegrationTest.java
lbergelson: this is already captured here https://github.com/broadinstitute/gatk-protected/issues/580

src/main/java/org/broadinstitute/hellbender/tools/walkers/ReferenceConfidenceVariantContextMerger.java
droazen: Ah, I didn't notice it was in a private inner class -- in that case, this is fine.

src/test/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFsIntegrationTest.java
lbergelson: They were all regenerated with gatk3.7_30_ga4f720357.  I've relabelled them all.

src/test/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFsIntegrationTest.java
lbergelson: added a comment with explanation

src/test/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFsIntegrationTest.java
lbergelson: I would have thought so, if you can find it, let me know and I'll change it..

src/test/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFsIntegrationTest.java
lbergelson: I've extracted these to a named constant at the top to make it clear what these are.  They're excluded from comparison.

src/test/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFsIntegrationTest.java
lbergelson: removed the comment since we fixed the differences in inbreeding coefficient

src/test/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFsUnitTest.java
lbergelson: ReferenceGenotypeQuality.  I ported this commit, which had an integration test we can't run, so I added this unit test instead.  Weirdly, none of the changes mention RGQ, so I don't know how it affects that... This unit test tests the changes that were made in the commit though.

src/test/resources/large/.gitattributes
lbergelson: updated

src/main/java/org/broadinstitute/hellbender/tools/walkers/GenotypeGVCFs.java
lbergelson: I don't think I removed that deliberately?  I'll add it in and see if it breaks anything...

droazen:
@lbergelson Are you going to be able to merge this before you leave? If not, let me know and leave me a list of what remains to be done.

lbergelson:
@droazen Working on it right now.

lbergelson:
adding a new method `isSpanningDeletion` and opening https://github.com/broadinstitute/gatk/issues/2589 to port it to public

lbergelson:
@droazen I think I've addressed your comments.   Do you want to take another look at the changes or should I merge?

droazen:
@lbergelson If you've addressed all the comments and tests are passing, go ahead and merge.

Issue #961, opened 2017-03-29
GATK CNV/ACNV automated WDL tests should use docker


LeeTL1220:

LeeTL1220:
Currently, the tests do not use docker.  The M2 WDL does use docker.

droazen:
Issue moved to [broadinstitute/gatk #2963](https://github.com/broadinstitute/gatk/issues/2963) via [**ZenHub**](https://www.zenhub.com/)

Issue #905, opened 2017-02-16
Add preemptible to runtime blocks of all GATK CNV and ACNV tasks.


LeeTL1220:
See issue #904

droazen:
Issue moved to [broadinstitute/gatk #2943](https://github.com/broadinstitute/gatk/issues/2943) via [**ZenHub**](https://www.zenhub.com/)

Issue #825, opened 2016-12-14
Use sub-workflows to iterate over single pairs in case CNV/ACNV workflow


LeeTL1220:
The current case workflow wdl for CNV/ACNV simply calls a pair workflow multiple times.  Support teams and workbench would prefer to have a single pair workflow that they can run and the GATK dev team can run a WDL that simply runs the pair workflow over a list of pairs.

LeeTL1220:
If you need to populate entity_id from a bam filepath:

``entity_id=sub(sub(p.right.left, "[/]*.*/", ""), "\\.bam$", "")``

This will give the basename.  

This has not been thoroughly tested.

LeeTL1220:
Better version:

``    String entity_id=sub(sub(sub(tumorBam, "gs://",""), "[/]*.*/", ""), "\\.bam$", "")``

samuelklee:
See #988.

PR #983, opened 2017-04-10
Revised WDL for CNV somatic panel and single-pair workflows.

### LINKED ISSUE #793: ``pon_gatk_workflow.wdl`` needs to have a parameter for memory on the CreatePanelOfNormals Task.
Currently, hardcoded at 4GB, which is rarely enough.

### LINKED ISSUE #820: ``GetBamFileName`` should leverage method to extract the bam filename in the WDL, instead of a separate task.
This is seems a bit wasteful, even for local/SGE backends.

This is a big problem for the cloud backends.  These backends will copy the file, just to call ``basename ...`` on it.  This task could be rolled into coverage.

samuelklee:
This contains panel and single-pair workflows as well as subworkflows for copy-ratio and allele-fraction cases.  Travis tests six paths:

-Panel WES
-Panel WGS
-Pair WES
-Pair WGS
-Tumor-only WES
-Tumor-only WGS

In principle, the CNV/CNV+ACNV branching could be tested as well, but I felt this was probably overkill.  However, I've verified that this branching does work properly.

Parameter json templates contain only the minimal parameters.  All optional/advanced parameters have been relegated to the task level, with documentation showing a few examples.

I will leave the addition of a multi-pair workflow with corresponding tests (#825) and Dockerizing (#905, #961) to a later PR.

Closes #793, closes #820.


samuelklee:
@asmirnov239 @mbabadi perhaps you can structure the germline WDL similarly?

scripts/acnv_cromwell_tests/case_gatk_acnv_workflow_wgs.json
```diff
@@ -1,37 +0,0 @@
...
-  "case_gatk_acnv_workflow.seg_param_alpha": "0.01",
-  "case_gatk_acnv_workflow.seg_param_nperm": "10000",
-  "case_gatk_acnv_workflow.seg_param_pmethod": "HYBRID",
-  "case_gatk_acnv_workflow.seg_param_minWidth": "2",
-  "case_gatk_acnv_workflow.seg_param_kmax": "25",
-  "case_gatk_acnv_workflow.seg_param_nmin": "200",
-  "case_gatk_acnv_workflow.seg_param_eta": "0.05",
-  "case_gatk_acnv_workflow.seg_param_trim": "0.025",
-  "case_gatk_acnv_workflow.seg_param_undoSplits": "NONE",
-  "case_gatk_acnv_workflow.seg_param_undoPrune": "0.05",
```
LeeTL1220: For WGS runs can you use the recommended values?  See the README for the GATK CNV WDL.  If you can't find it just ask.

scripts/cnv_cromwell_tests/somatic/cnv_somatic_pair_wes_tumor-only_workflow.json
```diff
@@ -0,0 +1,11 @@
+{
```
LeeTL1220: Can you rename this file using the word ``case`` if it is a case workflow?

scripts/cnv_wdl/somatic/README.md
```diff
@@ -0,0 +1,58 @@
+## Running the Somatic CNV WDL
```
LeeTL1220: This file was moved, right?  Is there anything in the content (other than renamed workflow) that is different?  I did not see anything...

scripts/cnv_wdl/somatic/cnv_somatic_allele_fraction_pair_workflow.wdl
```diff
@@ -0,0 +1,262 @@
+# Subworkflow for running GATK ACNV on tumor/normal or tumor-only cases. Supports both WGS and WES.
+#
+# Notes:
+#
+# - A normal BAM (normal_bam) is requrired for the tumor/normal workflow.  If not provided, the tumor-only workflow
+#   will be run.
+#
+# - ACNV is run only on the tumor.
```
LeeTL1220: ACNV or CNV?

scripts/cnv_wdl/somatic/cnv_somatic_allele_fraction_pair_workflow.wdl
```diff
@@ -0,0 +1,262 @@
...
+    File ref_fasta_dict
+    File ref_fasta_fai
+    String gatk_jar
+    Int? mem
+
+    # Sample names are derived from the bam filenames
+    String tumor_base_filename = sub(sub(sub(tumor_bam, "gs://", ""), "[/]*.*/", ""), "\\.bam$", "")
+    String normal_base_filename = sub(sub(sub(select_first([normal_bam, ""]), "gs://", ""), "[/]*.*/", ""), "\\.bam$", "") # Evaluates to "" if tumor-only and is not used
+
+    # TODO the below if-else block can be cleaned up once Cromwell support for optionals/conditionals improves
```
LeeTL1220: This is an improvement on how we've done it in the past.

LeeTL1220: Are you sure this works?  My understanding is that you actually need to check if the value is ``null`` not ``""``

scripts/cnv_wdl/somatic/cnv_somatic_pair_workflow.wdl
```diff
@@ -0,0 +1,112 @@
+# Workflow for running GATK CNV (and optionally, ACNV) on tumor/normal or tumor-only cases. Supports both WGS and WES.
+#
+# Notes:
+#
+# - The target file (targets) is required for the WES workflow and should be a TSV file with the column headers:
+#    contig    start    stop    name
+#   These targets will be padded on both sides by the amount specified by PadTargets.padding (default 250).
+#
+# - If a target file is not provided, then the WGS workflow will be run instead and the specified value of
+#   wgs_bin_size (default 10000) will be used.
```
LeeTL1220: If I run WGS, how do I specify the bin size?  Can I?

LeeTL1220: So the bin size is not exposed in this workflow?

LeeTL1220:
@samuelklee Back to you

scripts/acnv_cromwell_tests/case_gatk_acnv_workflow_wgs.json
samuelklee: Yup, the old defaults are used if not specified by setting the task-level parameters in the json.  See the command block of the PerformSegmentation task in cnv_somatic_copy_ratio_bam_workflow.wdl.

scripts/cnv_cromwell_tests/somatic/cnv_somatic_pair_wes_tumor-only_workflow.json
samuelklee: Can we chat in person to clarify the differences between case, sample, pair, etc.?  Also note that I called the copy-ratio workflow a "bam_workflow", since it can be run on a single normal or tumor bam, but I don't know if this is the appropriate term either.

scripts/cnv_wdl/somatic/README.md
samuelklee: Yes, it's different.  I only kept descriptions of the workflow-level parameters and examples of a few task-level parameters.

scripts/cnv_wdl/somatic/cnv_somatic_allele_fraction_pair_workflow.wdl
samuelklee: ACNV is run only on the tumor.  CNV is run on both the tumor and the normal, since this is how we did it before.  We could change this to be more parallel if you like (i.e., either run ACNV on both, or run CNV only on the tumor).

scripts/cnv_wdl/somatic/cnv_somatic_allele_fraction_pair_workflow.wdl
samuelklee: The select_first call in the definition of normal_base_filename takes care of this.

In general, I found using optionals to be a bit confusing and non-intuitive...for example, here I would have preferred to have something like

`String? normal_base_filename = sub(sub(sub(normal_bam, "gs://", ""), "[/]*.*/", ""), "\\.bam$", "")`

and then use

`${"--normalHets " + normal_base_filename}`

in the command block, which would obviate the need for the if-else.  Another possibility would be to use an if-then-else instead.  However, it seems both of these options don't work as you'd naturally expect.

scripts/cnv_wdl/somatic/cnv_somatic_pair_workflow.wdl
samuelklee: You can set it as a task-level parameter to CollectCoverage in the json.  Note that you will have to do this separately for both the tumor and the normal.  I'll add this to the readme as an example.

scripts/cnv_cromwell_tests/somatic/cnv_somatic_pair_wes_tumor-only_workflow.json
LeeTL1220: Sure!  I may have been a little confused.

scripts/cnv_wdl/somatic/cnv_somatic_allele_fraction_pair_workflow.wdl
LeeTL1220: Mimicking the old way is fine by me.

scripts/cnv_wdl/somatic/cnv_somatic_allele_fraction_pair_workflow.wdl
LeeTL1220: ``else`` is not implemented.

scripts/cnv_wdl/somatic/cnv_somatic_pair_workflow.wdl
LeeTL1220: ok

PR #984, opened 2017-04-10
Rev gatk-public dependency to 4.alpha.2-221-g2ecdef4-20170410.183103-1

droazen:


droazen:
@cmnbroad please review

PR #1007, opened 2017-04-24
Upgrade testNG and tests, add force resolution for some dependencies.

cmnbroad:
To match those used by GATK public.


droazen:
:+1 merge when test pass

PR #1010, opened 2017-04-26
New, simpler tool for splitting intervals and changes to M2 wdl.

davidbenjamin:
@LeeTL1220 Here you go.  The wdl changes are tested, with and without intervals specified.

### DIFF
--- a/scripts/mutect2_wdl/mutect2.wdl
+++ b/scripts/mutect2_wdl/mutect2.wdl
@@ -26,7 +26,7 @@
 workflow Mutect2 {
   # gatk4_jar needs to be a String input to the workflow in order to work in a Docker image
   String gatk4_jar
-  File intervals
+  File? intervals
   File ref_fasta
   File ref_fasta_index
   File ref_dict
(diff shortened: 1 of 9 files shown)


scripts/mutect2_wdl/mutect2.wdl
```diff
@@ -413,7 +407,7 @@ task SplitIntervals {
   }
 
   output {
-    Array[File] interval_files = glob("*.interval_list")
+    Array[File] interval_files = glob("*.intervals")
```
LeeTL1220: This extension is not going to cause any issues, right?

LeeTL1220:
@davidbenjamin Back at you.  Not much here.  

*I'm assuming that bed and interval_list are tested elsewhere.*

scripts/mutect2_wdl/mutect2.wdl
davidbenjamin: Nope, it works.

davidbenjamin:
@LeeTL1220 I have tested bed and interval_list inputs.

LeeTL1220:
Great.  Merge away!

PR #1016, opened 2017-04-27
Rev gatk-public dependency to 4.alpha.2-236-g1dde31b-20170427.202410-1

droazen:
This brings in a bug fix to the FS annotation to make it match R,
as well as some GenomicsDB-related test utilities.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -77,7 +77,7 @@ build.dependsOn installDist
 check.dependsOn installDist
 
 // NOTE: When changing the GATK version, these versions should be changed to specify the same version as used by GATK
-final gatkVersion = '4.alpha.2-234-gc3a44c9-20170427.144324-1'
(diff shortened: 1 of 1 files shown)


droazen:
@meganshand Once this is merged, how do you want to proceed with the evaluation? Is it worth waiting for GATK3 to get your FS fix as well?

meganshand:
If you are ok with me using a jar I built off of that gatk3 branch, then I wouldn't have to wait.

droazen:
I think we should wait until it's merged into gatk3/master -- otherwise we might tie out against a commit hash that won't exist in the future! It seems like you've already updated the tests, so feel free to merge that branch as soon as unit/integration tests are green in http://gsabamboo.broadinstitute.org/

meganshand:
Ok, I'm keeping my eye on it.

PR #1017, opened 2017-04-28
making GenotypeGVCFs write index files

lbergelson:
it previously didn't honor the engine arguments for creating index files but it does now

fixes broadinstitute/gatk#2635


droazen:
:+1: merging

PR #1027, opened 2017-05-08
Rev gatk public to 4.alpha.2-260-ge8d00ca-20170508.215840-1

droazen:
-Updates to GenomicsDBImport (add --batchSize and --consolidate,
 and provide a way to avoid downloading all headers up-front)

-Fix to RMSMappingQuality (don't blow up when 0 reads)

-Close threads properly in NIO code

-Allow HomRefBlock to handle GQs > 99 without blowing up.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerArgumentCollection.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerArgumentCollection.java
@@ -50,10 +50,10 @@ public class HaplotypeCallerArgumentCollection extends AssemblyBasedCallerArgume
      *
      * This argument allows you to set the GQ boundaries. HC expects a list of multiple GQ threshold values. To pass
      * multiple values, you provide them one by one with the argument, as in `-GQB 10 -GQB 20 -GQB 30` and so on. Note
-     * that GQ values are capped at 99 in the GATK.
+     * that GQ values are capped at 99 in the GATK, so values must be integers in the range [1, 99].
      */
     @Advanced
-    @Argument(fullName = "GVCFGQBands", shortName = "GQB", doc= "GQ thresholds for reference confidence bands", optional = true)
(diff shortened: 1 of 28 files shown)


droazen:
@lbergelson please review

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerArgumentCollection.java
```diff
@@ -50,10 +50,10 @@
      *
      * This argument allows you to set the GQ boundaries. HC expects a list of multiple GQ threshold values. To pass
      * multiple values, you provide them one by one with the argument, as in `-GQB 10 -GQB 20 -GQB 30` and so on. Note
-     * that GQ values are capped at 99 in the GATK.
+     * that GQ values are capped at 99 in the GATK, so values must be integers in the range [1, 99].
```
lbergelson: I think the max may be 100 if you want to capture 99's in a band with anything else.

lbergelson:
Thumbs up from me. We can address that comment later since we need this now.

src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/HaplotypeCallerArgumentCollection.java
samuelklee: These specify the lower inclusive bounds, so I think 99 is correct (perhaps the old documentation should be updated if this is not clear).  I believe that putting in 100 causes an exception to be thrown, but I'm on my phone so I can't easily check this.

samuelklee: Back at a computer---actually, it looks like I updated the documentation and changed the meaning of the GQB arguments in https://github.com/broadinstitute/gsa-unstable/pull/1451, along with changing the necessary tests.  These changes were not ported over in the GATK4 public/protected PRs by @ronlevine. so we should issue an additional PR to add these changes.

samuelklee: Filed https://github.com/broadinstitute/gatk-protected/issues/1029.

PR #1022, opened 2017-05-03
Limit GQ to fix GVCF block merging

ronlevine:
Implements https://github.com/broadinstitute/gatk-protected/issues/602

Issue #602, opened 2016-07-06
Port bug fix for GVCF block merging.


samuelklee:
See https://github.com/broadinstitute/gsa-unstable/pull/1431.

samuelklee:
Also see broadinstitute/gsa-unstable#1440.

samuelklee:
@lbergelson were these fixes ported?

lbergelson:
@samuelklee I don't think so, not by me at least.

ronlevine:
@samuelklee Would you like me to do the port? It will require both gatk and gatk-protected changes.

samuelklee:
If you don't mind, @ronlevine---I probably won't get around to it for a bit.  Not high priority though, I was just going through and cleaning up issues.

ronlevine:
Will do.

ronlevine:
Closing since https://github.com/broadinstitute/gatk/pull/2619 is merged.

ronlevine:
Reopening since need to do the protected part.

samuelklee:
Also note that https://github.com/broadinstitute/gsa-unstable/pull/1451 should be included in this port.

ronlevine:
The HaplotypeCaller tests that compare the output GVCFs between the current code and GATK 3.5 will not pass. This is due the block merging fix. So, I generated output files with with a version of GATK 3 that has this fix, 3.7-42-gbe6a37c. If it's premature for these tests, I can disable them.

droazen:
Please don't either disable these tests or move them to GATK 3.7 outputs just yet -- this would be disruptive to my current efforts to figure out what's behind some annotation differences in the GATK4 HaplotypeCaller. I do plan to migrate the tests to 3.7 outputs soon, but I need to study the diffs a bit first.

droazen:
Issue moved to [broadinstitute/gatk #2870](https://github.com/broadinstitute/gatk/issues/2870) via [**ZenHub**](https://www.zenhub.com/)

PR #1028, opened 2017-05-08
Isolate a problematic change to ReferenceConfidenceModel for further discussion

droazen:
I've extracted this change to ReferenceConfidenceModel from the now-closed PR
https://github.com/broadinstitute/gatk-protected/pull/1022, since it causes `HaplotypeCaller` tests
to fail spectacularly (including concordance tests against GATK3), so that we can
unblock the badly-needed merge of https://github.com/broadinstitute/gatk-protected/pull/1027

Let's review and test this change in isolation to be sure we understand it fully
before accepting it into master.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/ReferenceConfidenceModel.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/haplotypecaller/ReferenceConfidenceModel.java
@@ -213,8 +213,9 @@ public final class ReferenceConfidenceModel {
         // as our GLs for the site.
         final GenotypeLikelihoods leastConfidenceGLs = getGLwithWorstGQ(indelGLs, snpGLs);
 
-        gb.GQ((int) (-10 * getGQForHomRef(leastConfidenceGLs)));
-        gb.PL(leastConfidenceGLs.getAsPLs());
+        final int[] leastConfidenceGLsAsPLs = leastConfidenceGLs.getAsPLs();
+        gb.GQ(GATKVariantContextUtils.calculateGQFromPLs(leastConfidenceGLsAsPLs));
+        gb.PL(leastConfidenceGLsAsPLs);
 
         vcb.genotypes(gb.make());
         return vcb.make();


droazen:
@samuelklee and @ronlevine, since you two were associated with the original branch this proposed change came from, please weigh in here, including an explanation of:

* exactly what this patch does
* why it's needed
* why it changes test results so much
* what the consequences are of running off of the latest gatk public without this patch

I had to isolate this particular change from the others in order to be able to update protected to the latest public snapshot and unblock our work.

samuelklee:
This patch is part of port of a bug fix that resolved https://github.com/broadinstitute/gsa-unstable/issues/1345.  See discussion of original bug and details there.

This bug fix was included in GATK 3.7.  Tests against results generated by previous version of GATK 3 will fail because merging of hom-ref blocks by GQ band was performed incorrectly in those versions.  Without this patch, many hom-ref blocks that appear to be in the same GQ band (but are in fact not, due to rounding peculiarities) will remain unmerged.

A proper, complete port of the bug fix should encompass https://github.com/broadinstitute/gsa-unstable/pull/1431, https://github.com/broadinstitute/gsa-unstable/pull/1440, and https://github.com/broadinstitute/gsa-unstable/pull/1451 and would close #602.  Currently, various portions and iterations of the bug fix have been ported in both merged PRs (https://github.com/broadinstitute/gatk/pull/2619, https://github.com/broadinstitute/gatk/pull/2659, and https://github.com/broadinstitute/gatk-protected/pull/1027) and unmerged PRs (https://github.com/broadinstitute/gatk-protected/pull/1022 and the current PR).

droazen:
Ok, thanks for the explanation! So I take it that we still don't have the complete fix? What's missing, exactly, once this branch goes in? I found that we had to do an additional fix in public (https://github.com/broadinstitute/gatk/pull/2681) to get protected to stop blowing up over high GQs.

One nice feature of our artifactory is that it contains snapshots for every commit of every branch of gatk public that runs through travis -- so it would have been possible to craft complete, passing branches in both public and protected for this fix in advance, without having to first merge a partial fix into public/master and put us into this intermediate state where protected is not yet passing tests.

ronlevine:
What's missing: https://github.com/broadinstitute/gsa-unstable/pull/1451.

droazen:
Closing -- this has been moved to https://github.com/broadinstitute/gatk as https://github.com/broadinstitute/gatk/pull/2807

PR #1030, opened 2017-05-10
SplitIntervals now gives non-zero exit code if the gatk command fails

### LINKED ISSUE #918: When wget fails in Oncotator for M2 the task doesn't fail
It would be better for the task to fail if any of the internal steps fail.

### LINKED ISSUE #1006: M2 SplitInterval WDL returns 0 when it fails
Simply add a ``set -e`` to the command.  Otherwise, it will have rc 0 even when a step fails.

davidbenjamin:
Closes #918 and #1006.  @LeeTL1220 housekeeping using your proposed solution.

### DIFF
--- a/scripts/mutect2_wdl/mutect2.wdl
+++ b/scripts/mutect2_wdl/mutect2.wdl
@@ -388,6 +388,9 @@ task SplitIntervals {
   String m2_docker
 
   command {
+    # fail if *any* command below (not just the last) doesn't return 0, in particular if GATK SplitIntervals fails
+    set -e
+
     # Use GATK Jar override if specified
     GATK_JAR=${gatk4_jar}
     if [[ "${gatk4_jar_override}" == *.jar ]]; then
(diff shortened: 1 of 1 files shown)


scripts/mutect2_wdl/mutect2.wdl
```diff
@@ -397,6 +397,9 @@ task SplitIntervals {
     mkdir interval-files
     java -jar $GATK_JAR SplitIntervals -R ${ref_fasta} ${"-L " + intervals} -scatter ${scatter_count} -O interval-files
     cp interval-files/*.intervals .
+
+    # fail if any command above didn't return 0 -- that way if the gatk command fails, the task fails, even though cp doesn't
+    set -e
```
LeeTL1220: Ummm ... this needs to be at the beginning of the command block.  Not the end.

scripts/mutect2_wdl/mutect2.wdl
```diff
@@ -442,7 +445,11 @@ task oncotate_m2 {
           fi
 
 
-        /root/oncotator_venv/bin/oncotator --db-dir onco_dbdir/ -c $HOME/tx_exact_uniprot_matches.AKT1_CRLF2_FGFR1.txt  -v ${m2_vcf} ${entity_id}.oncotated.vcf hg19 -i VCF -o VCF --infer-onps --collapse-number-annotations --log_name oncotator.log
+        /root/oncotator_venv/bin/oncotator --db-dir onco_dbdir/ -c $HOME/tx_exact_uniprot_matches.AKT1_CRLF2_FGFR1.txt  \
```
LeeTL1220: Ditto

PR #1034, opened 2017-05-10
Add ability to generate tool documentation.

cmnbroad:
Until we resolve how we're going to integrate gatk-protected doc with gatk-public, the comms team needs a way to generate/review doc for protected tools. This PR adds the same docgen process we use for public, with the exception that you also need to have a GATK clone to run it, since javadoc from any dependent engine classes needs to be available.

I also tagged a handful of tools with @DocumentedFeatureObject so I could test the process manually.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -280,32 +280,37 @@ javadoc {
 }
 
 // Generate GATK Online Doc
-task gatkDoc(type: Javadoc, dependsOn: classes) {
+task gatkDoc(type: Javadoc, dependsOn: ['cleanGatkDoc', classes]) {
+    // A GATK source folder is required to resolve any dependent classes that will be part of the generated doc.
+    final String gatkSourceDir = project.hasProperty('gatkSourceDir') ?
+            project.getProperty('gatkSourceDir') :
+            '../gatk'
     final File gatkDocDir = new File("build/docs/gatkdoc")
-    String gatkSourceDir;
+    final String pathToGATKClasses = gatkSourceDir + '/src/main/java/org/broadinstitute/hellbender/'
+    final String pathToGATKHelpTemplates = gatkSourceDir + '/src/main/resources/org/broadinstitute/hellbender/utils/helpTemplates/'
     doFirst {
         // make sure the output folder exists or we can create it
         if (!gatkDocDir.exists() && !gatkDocDir.mkdirs()) {
-            throw new GradleException(String.format("Failure creating folder (%s) for GATK doc output in task (%s)",
+            throw new InvalidUserDataException(String.format("Failure creating folder (%s) for GATK doc output in task (%s)",
                     gatkDocDir.getAbsolutePath(),
                     it.name));
         }
-        // In order to generate docs for gatk-protected, a GATK source folder is required to resolve any dependent
-        // classes that will be part of the generated doc.
-        gatkSourceDir = project.hasProperty('gatkSourceDir') ?
-                project.getProperty('gatkSourceDir') :
-                '../gatk'
         if (!file(gatkSourceDir).exists()) {
             throw new GradleException("A GATK source directory is required to generate doc for gatk-protected: "
                     + gatkSourceDir + ". Set the gatkSourceDir property to the GATK source directory.")
         }
+        println("Note: The GATK source folder must contain the same version of GATK that is used to build gatk-protected")
     }
 
-    // We need to add the source for any GATK classes that are referenced by the local DocumentedFeatures
-    source = sourceSets.main.allJava +
-            files { file(gatkSourceDir + '/src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections').listFiles() } +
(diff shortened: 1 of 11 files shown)


build.gradle
```diff
@@ -258,6 +268,61 @@ task wrapper(type: Wrapper) {
...
+task gatkDoc(type: Javadoc, dependsOn: classes) {
+    final File gatkDocDir = new File("build/docs/gatkdoc")
+    String gatkSourceDir;
+    doFirst {
+        // make sure the output folder exists or we can create it
+        if (!gatkDocDir.exists() && !gatkDocDir.mkdirs()) {
+            throw new GradleException(String.format("Failure creating folder (%s) for GATK doc output in task (%s)",
+                    gatkDocDir.getAbsolutePath(),
+                    it.name));
+        }
```
droazen: Maybe delete the `gatkDocDir` if it exists?

build.gradle
```diff
@@ -258,6 +268,61 @@ task wrapper(type: Wrapper) {
...
+            throw new GradleException("A GATK source directory is required to generate doc for gatk-protected: "
+                    + gatkSourceDir + ". Set the gatkSourceDir property to the GATK source directory.")
+        }
+    }
+
+    // We need to add the source for any GATK classes that are referenced by the local DocumentedFeatures
+    source = sourceSets.main.allJava +
+            files { file(gatkSourceDir + '/src/main/java/org/broadinstitute/hellbender/cmdline/argumentcollections').listFiles() } +
+            files { file(gatkSourceDir + '/src/main/java/org/broadinstitute/hellbender/cmdline/GATKPlugin/').listFiles() } +
+            files { file(gatkSourceDir + '/src/main/java/org/broadinstitute/hellbender/engine/filters/').listFiles() }
```
droazen: This is super-brittle, but ok on a temporary basis.

build.gradle
```diff
@@ -258,6 +268,61 @@ task wrapper(type: Wrapper) {
...
+    // The gatkDoc process instantiates any documented feature classes, so to run it we need the entire
+    // runtime classpath, as well as jdk javadoc files such as tools.jar, where com.sun.javadoc lives.
+    classpath = sourceSets.main.runtimeClasspath + javadocJDKFiles
+    options.docletpath = classpath.asType(List)
+    options.doclet = "org.broadinstitute.hellbender.utils.help.GATKHelpDoclet"
+
+    outputs.dir(gatkDocDir)
+    options.destinationDirectory(gatkDocDir)
+
+    options.addStringOption("settings-dir", "src/main/resources/org/broadinstitute/hellbender/utils/helpTemplates");
```
droazen: Some of these paths should be declared as named constants

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2.java
```diff
@@ -90,6 +91,7 @@
         oneLineSummary = "Call somatic SNPs and indels via local re-assembly of haplotypes",
         programGroup = VariantProgramGroup.class
 )
+@DocumentedFeature
```
droazen: Does the comms team want to reserve this annotation for tools whose docs are complete and up-to-date?

Even if so, it's good to have some output initially for the gatkDoc task in protected.

droazen:
:+1: looks good, merge after addressing comments.

build.gradle
cmnbroad: Factored these strings out, as well as one for common the base path.

build.gradle
cmnbroad: I'll add a dependency on the (automatically generated) 'cleanGatkDoc' task.

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2.java
vdauwera: No, I'd rather have all tools show up unless explicitly tagged to be ignored. But conversely it would be good to be able to tag tools as being experimental or in development, eg specifying alpha or beta status. 

Actually, do/can we have a label system that allows us to use arbitrary tags? Something that we could use web-side to would allow users to filter tools by tag, eg "only show me alpha tools" or "only show me production-worthy tools", or "show me coverage tools"... Would like like eg `programTags = [alpha, coverage]` where alpha and coverage are defined in some curated class somewhere (so it doesn't become a wild wild west of alternative phrasings or spellings).

cmnbroad: In the current doc system, you need to use the DocumentedFeature annotation to get a tool to show up in doc. We do also have the ability to use embedded (custom) javadoc tags - the doc system will populate the freemarker template properties with these, and you can use them in html however you like. However, the command line help parser doesn't look at the javadoc, so that doesn't help with anything that needs to surface in command line help. There is a [Barclay PR](https://github.com/broadinstitute/barclay/pull/51) that implements an annotation that is integrated with both the CLP and the doc system to tag "experimental" tools (its actually called "EvaluationOnly" in the PR - but we can change that). It causes a property to be emitted to the freemarker template, and fixed string to be embedded in the command line output. Of course, any of it can be changed...

PR #1035, opened 2017-05-11
Concordance emits an additional vcf of filtered true and false negatives

davidbenjamin:
@takutosato could you review this addition to the concordance tool?

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/validation/Concordance.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/validation/Concordance.java
@@ -16,6 +16,7 @@ import org.broadinstitute.hellbender.engine.ReadsContext;
 import org.broadinstitute.hellbender.engine.ReferenceContext;
 import org.broadinstitute.hellbender.exceptions.UserException;
 import org.broadinstitute.hellbender.tools.walkers.variantutils.VariantsToTable;
+import org.broadinstitute.hellbender.utils.Utils;
 
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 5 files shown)


src/main/java/org/broadinstitute/hellbender/tools/walkers/validation/Concordance.java
```diff
@@ -75,11 +78,18 @@
...
+            shortName = FILTERED_TRUE_NEGATIVES_AND_FALSE_NEGATIVES_SHORT_NAME,
+            optional = true)
+    protected File filteredTrueNegativesAndFalseNegativesVcf = null;
+
     // we count true positives, false positives, false negatives for snps and indels
     private final EnumMap<ConcordanceState, MutableLong> snpCounts = new EnumMap<>(ConcordanceState.class);
     private final EnumMap<ConcordanceState, MutableLong> indelCounts = new EnumMap<>(ConcordanceState.class);
     private VariantContextWriter truePositivesAndFalseNegativesVcfWriter;
     private VariantContextWriter truePositivesAndFalsePositivesVcfWriter;
+    private VariantContextWriter filteredTrueNegativesAndFalseNegativesVcfWriter;
```
takutosato: Do we ever write filtered _true_ negatives to `filteredTrueNegativesAndFalseNegativesVcfWriter`? If not we should rename this variable (and its associated File) to `FalseNegativesVcfWriter`

src/main/java/org/broadinstitute/hellbender/tools/walkers/validation/Concordance.java
```diff
@@ -112,14 +128,51 @@ protected void apply(final TruthVersusEval truthVersusEval, final ReadsContext r
             indelCounts.get(concordanceState).increment();
         }
 
-        if (truePositivesAndFalseNegativesVcfWriter != null && concordanceState != ConcordanceState.FALSE_POSITIVE) {
-            final VariantContext vc = annotateWithConcordanceState(truthVersusEval.getTruth(), concordanceState);
-            truePositivesAndFalseNegativesVcfWriter.add(vc);
+        if (concordanceState == ConcordanceState.TRUE_POSITIVE) {
```
takutosato: I think a switch statement may be cleaner and might communicate the fact that we're going through possible cases (i.e. enum values)

takutosato:
@davidbenjamin back to you

src/main/java/org/broadinstitute/hellbender/tools/walkers/validation/Concordance.java
davidbenjamin: We do (only for filtered true negatives, as in this output file).

takutosato: I ask because on line 137 we just `return`; do we write elsewhere?

```java
        } else if (concordanceState == ConcordanceState.FILTERED_TRUE_NEGATIVE) {
            return;
```

davidbenjamin: Hmmm, maybe I forgot to push that commit. . .

src/main/java/org/broadinstitute/hellbender/tools/walkers/validation/Concordance.java
davidbenjamin: I think you're right, especially with the two new values of the enum.

PR #1036, opened 2017-05-11
Use Utils methods to improve clunky code

davidbenjamin:
@lbergelson now that these methods are in the snapshot, would you mind reviewing?


droazen:
@davidbenjamin Migration instructions for this branch: https://github.com/broadinstitute/gatk/wiki/Migrating-branches-from-gatk-protected-to-gatk

PR #1041, opened 2017-05-16
separate emitting from filtering based on tumor lod

davidbenjamin:
@takutosato This change will emit and then filter variants with low but not too low tumor lods.  This will be useful for analyzing false negatives and making ROC curves.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2FilteringEngine.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2FilteringEngine.java
@@ -51,9 +51,16 @@ public class Mutect2FilteringEngine {
     }
 
     private void applyTriallelicFilter(final VariantContext vc, final Collection<String> filters) {
-        if (vc.getNAlleles() > (MTFAC.numAltAllelesThreshold + 1)) {
-            filters.add(GATKVCFConstants.TRIALLELIC_SITE_FILTER_NAME);
+        if (vc.hasAttribute(GATKVCFConstants.TUMOR_LOD_KEY)) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2FilteringEngine.java
```diff
@@ -51,9 +51,16 @@ private void applyContaminationFilter(final M2FiltersArgumentCollection MTFAC, f
     }
 
     private void applyTriallelicFilter(final VariantContext vc, final Collection<String> filters) {
-        if (vc.getNAlleles() > (MTFAC.numAltAllelesThreshold + 1)) {
-            filters.add(GATKVCFConstants.TRIALLELIC_SITE_FILTER_NAME);
+        if (vc.hasAttribute(GATKVCFConstants.TUMOR_LOD_KEY)) {
+            final double[] tumorLods = getArrayAttribute(vc, GATKVCFConstants.TUMOR_LOD_KEY);
+            final long numPassingAltAlleles = Arrays.stream(tumorLods).filter(x -> x > MTFAC.TUMOR_LOD_THRESHOLD).count();
+
+            if (numPassingAltAlleles > (MTFAC.numAltAllelesThreshold + 1)) {
```
takutosato: Now that we use the number of passing alt alleles, we can get ride of that +1

takutosato:
@davidbenjamin please feel free to merge when ready

src/main/java/org/broadinstitute/hellbender/tools/walkers/mutect/Mutect2FilteringEngine.java
davidbenjamin: Nice catch, thank you!

PR #1046, opened 2017-05-17
Fixing issue in OB filter that caused crash when erroneously detectin…

LeeTL1220:
…g high diploid organism.


LeeTL1220:
@davidbenjamin Once merged into master, I'll apply the commit(s) to the alpha 1.2.6 release to create 1.2.7.

PR #1049, opened 2017-05-17
CNV AllelicCNV: Shl doc alleliccnv

sooheelee:
Please let me know who to assign these doc changes to @LeeTL1220 @vdauwera.

This is the first of the gatk-protected doc changes I've made so the changes are prototypical. If you object to what I've decided to do, speak up soon before I make more doc changes @vdauwera @samuelklee. I've some forty tools to go.

In addition to adding the doc tag, I went ahead and added in the WDL commands as example commands. For this tool, there were two example commands, one for WGS and another for other data types, e.g. WES. Because I use the WDL commands _as is_, I also added in a codeblock section to show the variable types (Int, File etc). This _as is_ format I think is understandable by folks who will not use WDLs, quick for me to incorporate and also enables folks who want to WDL'ize the commands, which we expect many will want to do.

I removed the `Boolean is_wgs` from the doc variable list, as I also omit the if/then component of the script--`if [ ${is_wgs} = true ]`.

The commands are taken from:
https://github.com/broadinstitute/gatk-protected/blob/158ac7fdb16622b41f706540e03a5f4fe2da9db8/scripts/cnv_wdl/somatic/cnv_somatic_allele_fraction_pair_workflow.wdl

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/AllelicCNV.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/AllelicCNV.java
@@ -2,6 +2,7 @@ package org.broadinstitute.hellbender.tools.exome;
 
 import org.apache.spark.api.java.JavaSparkContext;
 import org.broadinstitute.barclay.argparser.Argument;
+import org.broadinstitute.barclay.help.DocumentedFeature;
 import org.broadinstitute.hdf5.HDF5Library;
 import org.broadinstitute.barclay.argparser.CommandLineProgramProperties;
 import org.broadinstitute.hellbender.cmdline.ExomeStandardArgumentDefinitions;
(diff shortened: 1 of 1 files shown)


samuelklee:
:+1: from me.  I have no strong opinions about using WDL commands as examples, so I'll leave that up to @vdauwera.

vdauwera:
I'm concerned that using WDL commands as-is will cause much confusion in terms of syntax, and listing types adds clutter. In principle the commands should be copy-pastable and runnable if one had the same files.

sooheelee:
The commands will eventually change, I'm informed, to replace `java -jar gatkjar` to `gatk-launch`. I'd like to suggest that when the `gatk-launch` syntax gets incorporated, then we can overhaul the WDL-format commands. How do you feel about this @vdauwera?

vdauwera:
We won't release until the gatk-launch syntax is ready, so might as well write all of this like we want it to look when the release happens. In any case it's simple enough to make a one-word change automatically across all files in IntelliJ, compared to converting the WDL syntax.

sooheelee:
Ok, so go ahead and incorporate future changes, e.g.?

gatk-launch AllelicCNV \
 + *   --tumorHets somefilename.ext \
 + *   --tangentNormalized somefilename.ext \
...

sooheelee:
Just thinking this would be a great way to nudge users to WDL.

sooheelee:
Ok. Incorporated @vdauwera's suggestion for the example commands.

sooheelee:
I'm not sure why, but examining the code that has changed, I see more than the file I touched changed!

ronlevine:
@sooheelee Because the master branch has changed. Rebasing should fix that issue.

sooheelee:
Expert helped me with cobbled branch issues and now I only have one commit that shows my changes to the one document. This is ready for review.

LeeTL1220:
Based on @samuelklee  and my own review.  Feel free to merge @sooheelee

PR #1052, opened 2017-05-17
Update to 4.alpha.2-272-gf4d928f-20170517.201327-1

lbergelson:
This includes an important change to allele subsetting code that was causing differences with gatk3.


lbergelson:
@cmnbroad Really just a formality here since there no interpretable changes.

cmnbroad:
@lbergelson Looks good - merge at will.

PR #1060, opened 2017-05-18
Eliminated other erroneous ploidy errors.

### LINKED ISSUE #1043: M2 WDL does not fail if a single command line fails in the Filter task
Need ``set -e`` in the command block.  Otherwise, the Filter task continues even if one of the CLI invocations fails.  This makes workflows really hard to debug.

### LINKED ISSUE #1044: FilterByOrientationBias will fail when it (erroneously) determines that it is running on a non-diploid organism.
This caused a failure in a tumor sample.  In OrientationBiasFilterer.java, remove (or change to a warning) on roughly line 60-62. 

```
         if (genotype.getPloidy() != 2) {
                throw new UserException.BadInput("This tool will not run with non-diploid organisms.  Saw GT: " + genotype.getGenotypeString());
            }
```

LeeTL1220:
- Accidentally missed the same error occurring in other places.  Fixed here.

- Also, fixed WDL.

Closes #1043 
Closes #1044

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasUtils.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasUtils.java
@@ -90,8 +90,7 @@ public class OrientationBiasUtils {
         Utils.nonNull(g, "Genotype cannot be null");
         Utils.nonNull(transition, "Artifact mode cannot be null");
 
-        return IntStream.range(1,g.getAlleles().size()).anyMatch(i -> g.getAllele(0).getDisplayString().equals(String.valueOf(transition.ref())) &&
-                g.getAllele(i).getDisplayString().equals(String.valueOf(transition.call())));
+        return IntStream.range(1,g.getAlleles().size()).anyMatch(i -> g.getAllele(0).basesMatch(g.getAllele(i)));
     }
 
     /** See {@link #isGenotypeInTransition}, except that this will take into account complements.
@@ -105,15 +104,7 @@ public class OrientationBiasUtils {
         Utils.nonNull(g, "Genotype cannot be null");
         Utils.nonNull(transition, "Transition cannot be null");
 
-        final boolean isInTransition = isGenotypeInTransition(g, transition);
-
-        if (isInTransition) {
-            return true;
-        }
-
-        final Transition transitionComplement = transition.complement();
-        return isGenotypeInTransition(g, transitionComplement);
-
+        return isGenotypeInTransition(g, transition) || isGenotypeInTransition(g, transition.complement());
     }
 
     /** Is this genotype in any of the specified artifact modes (or complements)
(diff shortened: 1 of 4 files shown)


src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasUtils.java
```diff
@@ -78,26 +79,23 @@ public static String getGenotypeString(final Genotype g, final String fieldName)
...
     public static boolean isGenotypeInTransition(final Genotype g, final Transition transition) {
         Utils.nonNull(g, "Genotype cannot be null");
         Utils.nonNull(transition, "Artifact mode cannot be null");
 
-        if (g.getAlleles().size() != 2) {
-            throw new UserException.BadInput("Cannot run this method on non-diploid genotypes.");
-        }
-        return g.getAllele(0).getDisplayString().equals(String.valueOf(transition.ref())) &&
-                g.getAllele(1).getDisplayString().equals(String.valueOf(transition.call()));
+        return IntStream.range(1,g.getAlleles().size()).anyMatch(i -> g.getAllele(0).getDisplayString().equals(String.valueOf(transition.ref())) &&
```
davidbenjamin: This should be doable with `Allele::basesMatch` without converting the allele and the transition base to `String`s.

src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasUtils.java
```diff
@@ -107,19 +105,14 @@ public static boolean isGenotypeInTransitionWithComplement(final Genotype g, fin
         Utils.nonNull(g, "Genotype cannot be null");
         Utils.nonNull(transition, "Transition cannot be null");
 
-        if (g.getAlleles().size() != 2) {
-            throw new UserException.BadInput("Cannot run this method on non-diploid genotypes.");
-        }
-        final boolean isInTransition = g.getAllele(0).getDisplayString().equals(String.valueOf(transition.ref())) &&
-                g.getAllele(1).getDisplayString().equals(String.valueOf(transition.call()));
+        final boolean isInTransition = isGenotypeInTransition(g, transition);
```
davidbenjamin: This would look more readable to me as `return isGenotypeInTransition(g, transition) || is GenotypeInTransition(g, transition.complement())`.

src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasUtils.java
LeeTL1220: Done

src/main/java/org/broadinstitute/hellbender/tools/exome/orientationbiasvariantfilter/OrientationBiasUtils.java
LeeTL1220: Done

PR #1067, opened 2017-05-19
CNV CalculateTargetCoverage: tag doc; example WES command; change defaults

sooheelee:
Example command is only for WES data. For WGS, I point users to SparkGenomeReadCounts.

- I nudge users to use PadTargets to pad their targets and tell them the target intervals do not overlap. I also state NOT to use BED format and point them to ConvertBedToTargetFile.
- I leave in the optional reference to safeguard newbie users. I do state that 
    ```
    The reference is optional if --disableSequenceDictionaryValidation is set to true.
    ```
- description for targets param changed to specify 1-based indexing. 
- Default params changed for:
   --groupBy SAMPLE 
   --transform PCOV 
   --targetInformationColumns FULL

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
@@ -47,13 +47,13 @@ import java.util.stream.IntStream;
  *
  * <pre>
  * java -Xmx4g -jar $gatk_jar CalculateTargetCoverage \
- *   --input bam.bam \
+ *   --input sample.bam \
  *   --targets padded_targets.tsv \
  *   --groupBy SAMPLE \
  *   --transform PCOV \
  *   --targetInformationColumns FULL \
  *   --disableReadFilter NotDuplicateReadFilter \
- *   --output base_filename.coverage.tsv
+ *   --output sample.coverage.tsv
  * </pre>
  *
  * <p>
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
@@ -58,11 +58,11 @@ import java.util.stream.IntStream;
  *
  * <p>
  *     The interval targets are exome target intervals padded, e.g. with 250 bases on either side.
- *     Target intervals do NOT overlap. Use the PadTargets tool to generate non-overlapping padded intervals from exome targets.
+ *     Target intervals do NOT overlap. Use the {@link PadTargets} tool to generate non-overlapping padded intervals from exome targets.
  *     Do NOT use BED format. See {@link ConvertBedToTargetFile}.
  * </p>
  *
- * <p>For whole genome sequencing (WGS) data, use SparkGenomeReadCounts instead.</p>
+ * <p>For whole genome sequencing (WGS) data, use {@link SparkGenomeReadCounts} instead.</p>
  *
  */
 @CommandLineProgramProperties(
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
@@ -49,6 +49,9 @@ import java.util.stream.IntStream;
  * java -Xmx4g -jar $gatk_jar CalculateTargetCoverage \
  *   --input bam.bam \
  *   --targets padded_targets.tsv \
+ *   --groupBy SAMPLE \
+ *   --transform PCOV \
+ *   --targetInformationColumns FULL \
  *   --disableReadFilter NotDuplicateReadFilter \
  *   --output base_filename.coverage.tsv
  * </pre>
@@ -134,7 +137,7 @@ public final class CalculateTargetCoverage extends ReadWalker {
             shortName = GROUP_BY_SHORT_NAME,
             fullName = GROUP_BY_FULL_NAME,
             optional = true)
-    protected GroupBy groupBy = GroupBy.SAMPLE;
+    protected GroupBy groupBy = GroupBy.COHORT;
 
     @Argument(
             doc = "Cohort name used to name the read count column when the groupBy " +
@@ -151,7 +154,7 @@ public final class CalculateTargetCoverage extends ReadWalker {
             fullName = TRANSFORM_FULL_NAME,
             optional = true
     )
-    protected Transform transform = Transform.PCOV;
+    protected Transform transform = Transform.RAW;
 
     @Argument(
             doc = "TSV file listing 1-based genomic intervals with specific column headers. Do NOT use BED format.",
@@ -167,7 +170,7 @@ public final class CalculateTargetCoverage extends ReadWalker {
             fullName = TARGET_OUT_INFO_FULL_NAME,
             optional = true
     )
-    protected TargetOutInfo targetOutInfo = TargetOutInfo.FULL;
+    protected TargetOutInfo targetOutInfo = TargetOutInfo.COORDS;
 
     /**
      * Writer to the main output file indicated by {@link #output}.
@@ -693,7 +696,7 @@ public final class CalculateTargetCoverage extends ReadWalker {
         RAW((count, columnTotal) -> Integer.toString(count)),
 
         /**
-         * Default proportional coverage transformation.
+         * Proportional coverage transformation.
          * <p>Individual counts are transformed into the fraction of the total
          * count across the enclosing column.</p>
          */
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
@@ -32,7 +32,7 @@ import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
 /**
- * Calculate read-counts across targets given their reference coordinates.
+ * Calculates read-counts across targets for the exome copy number variant (CNV) calling workflow.
  *
  * @author Valentin Ruano-Rubio &lt;valentin@broadinstitute.org&gt;
  *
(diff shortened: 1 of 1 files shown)


sooheelee:
Given changing default parameters cause integration tests to fail, and the less than two week time frame I have been given to document 40-some tools in gatk-protected, and the presumed unavailability of developers to then fix the integration tests (either the test values or the compared to results), it may be best if I do NOT change any default parameters going forward for these tools. @vdauwera?

---
cnorman [3:03 PM] 
The travis tests for your CalculateTargetCoverage branch are failing (the tests run automatically when you submit a branch or make a PR). I'd guess more than likely its due to the changes to the defaults.

[3:03] 
since its the tests for that tool that are failing

shlee [3:04 PM] 
Do you mean the integration tests or unit tests?

[3:04] 
Do I need to change the test parameters?

cnorman [3:05 PM] 
I don't know - there ar 10 failing - not sure if they're all integration tests

shlee [3:06 PM] 
Because I have set the defaults to allow the most parsimonious example command, which in turn is based on recommended parameters.

cnorman [3:06 PM] 
I think whoever owns the tool/tests needs to look at them to see what needs to be done

[3:07] 
Yes, but when the tests run, the check the results, and changing the default values can change the results of running the tool. So most likely the test checks now fail since the results are not what is expected.

shlee [3:07 PM] 
I see. So if for the test commands we call for the parameters that were previously default, then these tests would again succeed.

[3:08] 
Or we need to change the compared to results

cnorman [3:08 PM] 
exactly.

shlee [3:08 PM] 
Since it makes sense to test the most used parameters.

cnorman [3:09 PM] 
thats up to the dev (which way they want to make the tests work). But either way the change can't go in until the tests succeed

shlee [3:10 PM] 
I will talk to the devs

sooheelee:
After discussing with Hellbender team, removed reference from the command.

sooheelee:
After discussion with DR and Geraldine, decided going forward, at least for the next two weeks, not to change any default parameters. 

However, CalculateTargetCoverage's integration test remains broken because of the changed default parameters. This is the only tool that needs a fix in the integration test during this doc update @samuelklee @mbabadi @achevali.

vdauwera:
Recommend changing the defaults back to the original value in that tool.

sooheelee:
Ok, will do.

LeeTL1220:
@sooheelee The automated tests are still failing.  I'll send out for review when that is fixed.

sooheelee:
@LeeTL1220 checks are passing now.

davidbenjamin:
@sooheelee I can review this.

src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
```diff
@@ -34,12 +35,42 @@
  * Calculate read-counts across targets given their reference coordinates.
  *
  * @author Valentin Ruano-Rubio &lt;valentin@broadinstitute.org&gt;
+ *
+ * <h3>Examples</h3>
+ *
+ * <p>
+ *     The command encompasses empirically determined parameters for TCGA project data.
+ *     You may obtain better results with different parameters.
```
davidbenjamin: This might not be clear to some users.  You should make clear that this tool is an input to our exome CNV pipeline, and so different parameters may give better results *for exome CNV calling*.

src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
```diff
@@ -34,12 +35,42 @@
...
+ *   --transform PCOV \
+ *   --targetInformationColumns FULL \
+ *   --disableReadFilter NotDuplicateReadFilter \
+ *   --output sample.coverage.tsv
+ * </pre>
+ *
+ * <p>
+ *     The interval targets are exome target intervals padded, e.g. with 250 bases on either side.
+ *     Target intervals do NOT overlap. Use the PadTargets tool to generate non-overlapping padded intervals from exome targets.
+ *     Do NOT use BED format. See ConvertBedToTargetFile.
```
davidbenjamin: Should this be `See {@link ConvertBedToTargetFile}`?

src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
```diff
@@ -34,12 +35,42 @@
...
+ * <p>For whole exome sequencing (WES) data: </p>
+ *
+ * <pre>
+ * java -Xmx4g -jar $gatk_jar CalculateTargetCoverage \
+ *   --input sample.bam \
+ *   --targets padded_targets.tsv \
+ *   --groupBy SAMPLE \
+ *   --transform PCOV \
+ *   --targetInformationColumns FULL \
+ *   --disableReadFilter NotDuplicateReadFilter \
```
davidbenjamin: Would removing this make the integration tests fail?

src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
```diff
@@ -110,7 +141,7 @@
 
     @Argument(
             doc = "Cohort name used to name the read count column when the groupBy " +
-                    "argument is set to COHORT (default)",
+                    "argument is set to COHORT",
```
davidbenjamin: This doesn't need to be on a new line.

src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
```diff
@@ -34,12 +35,42 @@
...
+ *     The command encompasses empirically determined parameters for TCGA project data.
+ *     You may obtain better results with different parameters.
+ * </p>
+ *
+ * <p>For whole exome sequencing (WES) data: </p>
+ *
+ * <pre>
+ * java -Xmx4g -jar $gatk_jar CalculateTargetCoverage \
+ *   --input sample.bam \
+ *   --targets padded_targets.tsv \
```
davidbenjamin: It might help to explicitly write down the header and a line or two of the input targets file format.  I think some of our other CNV tools do this in their javadoc.

src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
```diff
@@ -660,7 +691,7 @@ protected CountColumns countColumns(final CalculateTargetCoverage tool) {
     protected enum Transform {
 
         /**
-         * Default read-count raw value (non-)transformation.
+         * Read-count raw value (non-)transformation.
```
davidbenjamin: I think `Raw integer read count (non-)transformation` would make this a tad clearer.

davidbenjamin:
@sooheelee Back to you.

src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
sooheelee: Because the defaults were changed so that they could be removed. Defaults are now changed back and these parameters have to be put back in.

src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
sooheelee: Sure.

src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
sooheelee: Modified top to "Calculates read-counts across targets for the exome copy number variant (CNV) calling workflow."

src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
sooheelee: I explain the format in ConvertBedToTargetFile's tool doc and will now link to this doc.

sooheelee: When I say explain, I mean I give an example 3 lines from a file.

sooheelee:
@davidbenjamin I've incorporated your feedback and answered questions. Thank you for reviewing.

src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
sooheelee: I get it now. Fixed.

src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
sooheelee: Will do.

davidbenjamin:
@sooheelee looks good!

src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
```diff
@@ -31,15 +32,45 @@
...
+ *   --groupBy SAMPLE \
+ *   --transform PCOV \
+ *   --targetInformationColumns FULL \
+ *   --disableReadFilter NotDuplicateReadFilter \
+ *   --output sample.coverage.tsv
+ * </pre>
+ *
+ * <p>
+ *     The interval targets are exome target intervals padded, e.g. with 250 bases on either side.
+ *     Target intervals do NOT overlap. Use the PadTargets tool to generate non-overlapping padded intervals from exome targets.
```
samuelklee: Javadoc link to PadTargets?

src/main/java/org/broadinstitute/hellbender/tools/exome/CalculateTargetCoverage.java
```diff
@@ -31,15 +32,45 @@
...
+ *     Do NOT use BED format. See {@link ConvertBedToTargetFile}.
+ * </p>
+ *
+ * <p>For whole genome sequencing (WGS) data, use SparkGenomeReadCounts instead.</p>
```
samuelklee: Same for SparkGenomeReadCounts?

sooheelee:
@samuelklee 

> discuss the issue with the default parameters and integration tests in person, it should be an easy fix.

Sounds good.

sooheelee:
@samuelklee I will merge this now unless you envision the fixes to be a part of this PR?

samuelklee:
Go ahead and merge!

PR #1068, opened 2017-05-19
CNV SparkGenomeReadCounts tag doc; example WGS command

sooheelee:
### Questions yet to be answered:

- Who is the author of SparkGenomeReadCounts?
- Doesn't this tool need at least contig intervals, e.g. passed in via -L, to parallelize over Spark workers? 
- I will use the following opening, is it ok?
```
/**
 * Calculates read coverage on whole genome sequencing (WGS) alignments using Spark.
```

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/genome/SparkGenomeReadCounts.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/genome/SparkGenomeReadCounts.java
@@ -42,9 +42,9 @@ import java.util.stream.Collectors;
  *
  * <pre>
  * java -Xmx4g -jar $gatk_jar SparkGenomeReadCounts \
- *   --input bam.bam \
+ *   --input sample.bam \
  *   --disableReadFilter NotDuplicateReadFilter \
- *   --outputFile base_filename.coverage.tsv
+ *   --outputFile sample.coverage.tsv
  * </pre>
  *
  * <p>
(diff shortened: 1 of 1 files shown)


sooheelee:
After discussing with Hellbender team, removed reference from the command.

sooheelee:
Just curious if the tool spends a lot of time over synthetic regions of the genome, e.g. alpha satellite repeat arrays of GRCh38.

achevali:
@LeeTL1220 Is there a reason that the input bam file is an optional and not a required argument?

LeeTL1220:
@achevali How do you figure that it is optional?  @sooheelee Can you confirm that the input bam is not optional?  It seemed required to me, but I might be missing something.
@sooheelee Not sure of the answer.  We need to test on 38...

> Who is the author of SparkGenomeReadCounts?

Originally, @achevali and myself.

> Doesn't this tool need at least contig intervals, e.g. passed in via -L, to parallelize over Spark workers?

No, in fact we may want to fail if -L is passed in.  This tool creates the interval file on your behalf.

> I will use the following opening, is it ok?

Yes, fine by me.

LeeTL1220:
@sooheelee Back to you.

achevali:
@LeeTL1220 It's in Optional Arguments:
```--input,-I:String             BAM/SAM/CRAM file containing reads  This argument may be specified 0 or more times.  Default value: null. ```

sooheelee:
As much as I'd love to help on your recent questions, I leave all code related questions to you coders @achevali @LeeTL1220. Might I remind you that I'm your _technical writer_ who doesn't know how to code.... If you want to change the code, please feel free to use this PR. I've made my changes to the doc portion and would like to merge this today if you guys give me the thumbs up.

LeeTL1220:
@achevali Please file this as a bug.  Input file is required.  Thanks.

@sooheelee I'm not sure of SOP, but the doc changes look fine, except that the input file should be required.

sooheelee:
@LeeTL1220 This goes without saying. I think the users will know that it is required without me stating the obvious despite the bug. 

The example command shows an input file.

PR #1073, opened 2017-05-19
CNV CreatePanelOfNormals: tag doc, add example command

sooheelee:
Questions ~~yet to be~~ answered:

4. For CreatePanelOfNormals is the spark option now working? I see in the WDL we do not use the --disableSpark option.
5. How can I explain the target_weights.txt file?

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/CreatePanelOfNormals.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/CreatePanelOfNormals.java
@@ -2,6 +2,7 @@ package org.broadinstitute.hellbender.tools.exome;
 
 import org.apache.spark.api.java.JavaSparkContext;
 import org.broadinstitute.barclay.argparser.CommandLineProgramProperties;
+import org.broadinstitute.barclay.help.DocumentedFeature;
 import org.broadinstitute.hdf5.HDF5File;
 import org.broadinstitute.hdf5.HDF5Library;
 import org.broadinstitute.barclay.argparser.Argument;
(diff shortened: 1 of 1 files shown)


LeeTL1220:
> For CreatePanelOfNormals is the spark option now working? I see in the WDL we do not use the --disableSpark option.

Yup, it works (for 1-1.5 years).  There is not much incentive not to use it, so it should be the default.

> How can I explain the target_weights.txt file

The target weights file is an optional output file that can be leveraged in ``CallSegments``.  We have found that it does not help the results much, so we generate it, but I think we rarely use it.

src/main/java/org/broadinstitute/hellbender/tools/exome/CreatePanelOfNormals.java
```diff
@@ -39,13 +40,38 @@
  * </p>
  *
  * @author Valentin Ruano-Rubio &lt;valentin@broadinstitute.org&gt;
+ *
+ * <h3>Examples</h3>
+ * <p>
+ *     The command encompasses empirically determined parameters for TCGA project data.
```
LeeTL1220: This is not accurate.  This has been run on many sources of data.  And in fact, much was done on internal test samples unrelated to TCGA.

sooheelee: Our imperative is to nudge users to determine the parameters that work best for their own data and not for them to blindly think what we give are best practices (which they are not).

The most informative thing we can give our users is the representative technology that the parameters were determined to work on, e.g. data circa 2012 or TCGA-type. I'm informed that most of the CNV test data is TCGA.

I opt to keep these statements. If you feel strongly otherwise @LeeTL1220, please take this up with @samuelklee and also inform me what data type was used to determine the parameters.

sooheelee: We wait for @samuelklee's return from vacation.

sooheelee: Nope, Sam doesn't get back until Wednesday and we need to get moving. So I need to go with what Lee is saying.

sooheelee: @LeeTL1220 This is ready for your review.

LeeTL1220:
@sooheelee Looks good to me.

src/main/java/org/broadinstitute/hellbender/tools/exome/CreatePanelOfNormals.java
```diff
@@ -39,13 +40,34 @@
...
+ *
+ * <h3>Examples</h3>
+ *
+ * <p>
+ *     The following command is for either whole exome sequencing (WES) or whole genome sequencing (WGS) data.
+ * </p>
+ *
+ * <pre>
+ * java -Xmx4g -jar $gatk_jar CreatePanelOfNormals \
+ *   --input corrected_coverages.tsv \
```
samuelklee: Perhaps `gc_corrected_coverages.tsv` instead.

droazen:
@sooheelee For future reference, you should use "Squash and Merge" instead of "Merge" when merging a PR into master. This combines the commits on your branch into a single commit, and lets you edit the final commit message that ends up in the git history. Otherwise we end up with commits like "fixed typo" in our permanent commit history.

PR #1074, opened 2017-05-19
CNV NormalizeSomaticReadCounts: tag doc and provide example command

sooheelee:
### Pending questions

6. For NormalizeSomaticReadCounts, in the WDL I see two additional optional output files that I've not seen before--the Fnt and betaHats:
```
*   --input coverage.tsv \
*   --panelOfNormals cnv_panel_of_normals.pon \
*   --tangentNormalized ${entity_id}.tn.tsv \
*   --factorNormalizedOutput ${entity_id}.fnt.tsv \
*   --preTangentNormalized ${entity_id}.preTN.tsv \
*   --betaHatsOutput ${entity_id}.betaHats.tsv
```
Do we now recommend these be output as well? Or shall I stick to the Vancouver demo's simplified command with just PTN and TN?

7. For NormalizeSomaticReadCounts, in contrast to the WDL, I've removed the --targets option. If I recall, this is no longer necessary because it was already applied upstream in the workflow. Is this okay?

8. For NormalizeSomaticReadCounts, there is a note to developers that could use clarification in the choice of words, e.g. "case", "sample" and "use cases":

A note to developers:  If this is extended to use Spark, please be wary that the parallelization in tangent normalization is
* by case sample, which may not yield benefits for most use cases (which are one sample)

---
### Dependent suggestion
If we will recommend all four outputs, I think NormalizeSomaticReadCounts could also use an --outputPrefix option. See AllelicCNV:

```
--outputPrefix,-pre:String    Prefix for output files. Will also be used as the sample name in downstream plots.(Note:
                              if this is a file path or contains slashes (/), the string after the final slash will be
                              used as the sample name in downstream plots.)  Required. 
```

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/NormalizeSomaticReadCounts.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/NormalizeSomaticReadCounts.java
@@ -36,7 +36,7 @@ import java.util.List;
  * <pre>
  * java -Xmx4g -jar $gatk_jar NormalizeSomaticReadCounts \
  *   --input tumor.coverage.tsv \
- *   --targets padded_targets.tsv
+ *   --targets padded_targets.tsv \
  *   --panelOfNormals panel_of_normals.pon \
  *   --tangentNormalized tumor.tn.tsv \
  *   --preTangentNormalized tumor.preTN.tsv
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/NormalizeSomaticReadCounts.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/NormalizeSomaticReadCounts.java
@@ -30,11 +30,6 @@ import java.util.List;
  *
  * <h3>Examples</h3>
  * <p>
- *     The command encompasses empirically determined parameters for TCGA project data.
- *     You may obtain better results with different parameters.
- * </p>
- *
- * <p>
  *     The following command is for either whole exome sequencing (WES) or whole genome sequencing (WGS) data.
  * </p>
  *
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/NormalizeSomaticReadCounts.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/NormalizeSomaticReadCounts.java
@@ -36,11 +36,10 @@ import java.util.List;
  * <pre>
  * java -Xmx4g -jar $gatk_jar NormalizeSomaticReadCounts \
  *   --input tumor.coverage.tsv \
+ *   --targets padded_targets.tsv
  *   --panelOfNormals panel_of_normals.pon \
  *   --tangentNormalized tumor.tn.tsv \
- *   --factorNormalizedOutput tumor.fnt.tsv \
- *   --preTangentNormalized tumor.preTN.tsv \
- *   --betaHatsOutput tumor.betaHats.tsv
+ *   --preTangentNormalized tumor.preTN.tsv
  * </pre>
  */
 @CommandLineProgramProperties(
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/hellbender/tools/exome/NormalizeSomaticReadCounts.java
```diff
@@ -20,18 +21,39 @@
...
  *
- * <p> Dev note:  If this is extended to use spark, please be wary that the parallelization in tangent normalization is
+ * <p> A note to developers:  If this is extended to use Spark, please be wary that the parallelization in tangent normalization is
  * by case sample, which may not yield benefits for most use cases (which are one sample)  </p>
  *
  * @author Valentin Ruano-Rubio &lt;valentin@broadinstitute.org&gt;
+ *
+ * <h3>Examples</h3>
+ * <p>
+ *     The command encompasses empirically determined parameters for TCGA project data.
```
LeeTL1220: This does not make a lot of sense here.  Especially, since the pon file in the example does not actually exist.  And it's not really accurate anyway.

LeeTL1220:
> Do we now recommend these be output as well? Or shall I stick to the Vancouver demo's simplified command with just PTN and TN?

Stick with simplified command, if it works.

> For NormalizeSomaticReadCounts, in contrast to the WDL, I've removed the --targets option. If I recall, this is no longer necessary because it was already applied upstream in the workflow. Is this okay?

I'd have to run the command.  I think it is not okay.

> For NormalizeSomaticReadCounts, there is a note to developers that could use clarification in the choice of words, e.g. "case", "sample" and "use cases":

I am not totally sure how to address this.  Maybe replace "use cases" with "usage patterns"?  As for case and sample ,we may have to introduce this terminology earlier, but I am not sure how best to address that.

src/main/java/org/broadinstitute/hellbender/tools/exome/NormalizeSomaticReadCounts.java
sooheelee: Here I will agree with you so will remove this line.

sooheelee:
@LeeTL1220 I don't know what actually works. We don't have the time to test the commands. I can go by what is listed in the _Required_ field.

sooheelee:
@LeeTL1220 did you get a chance to run the commands?

LeeTL1220:
@sooheelee The issue is not "scientific" -- it's practical.  In general, the WDL is actually tested, so I would suggest what is in the WDL.  

Looks like the targets file is in fact required and the label as optional is a bug.  I've filed an issue.  I would add the targets parameter back and then merge this.

sooheelee:
Alright, @LeeTL1220, I've added in the targets 

and removed these two from the outputs since you say these are only useful for developers:

--factorNormalizedOutput ${entity_id}.fnt.tsv \
--betaHatsOutput ${entity_id}.betaHats.tsv

LeeTL1220:
@sooheelee Seems reasonable to me.  Merge away.

src/main/java/org/broadinstitute/hellbender/tools/exome/NormalizeSomaticReadCounts.java
sooheelee: I incorporated this feedback. Let me see what this repo buttons do.

sooheelee: Not what I wanted. I will use the `dismiss` button for this instance.

sooheelee:
Note for future iterations: I did not address the following because the comment in question talks about Sparkification of the tool but the tool parameters do not show an option to Sparkify nor to disable Spark. 

>> For NormalizeSomaticReadCounts, there is a note to developers that could use clarification in the choice of words, e.g. "case", "sample" and "use cases":
> I am not totally sure how to address this. Maybe replace "use cases" with "usage patterns"? As for case and sample ,we may have to introduce this terminology earlier, but I am not sure how best to address that.

The actual section in question reads:
```
<p> A note to developers:  If this is extended to use Spark, please be wary that the parallelization in tangent normalization is
 * by case sample, which may not yield benefits for most use cases (which are one sample)  </p>
```

samuelklee:
Looks good to me.  We can just remove the dev note if it causes confusion.

PR #1075, opened 2017-05-19
CNV PerformSegmentation: tag doc, note R requirement and give example commands

sooheelee:
### ~~Pending questions~~:
- Is the R and R component requirement still the case?
- Does the script at https://github.com/broadinstitute/gatk-protected/blob/master/scripts/install_R_packages.R
still apply?

---
### Default parameters that should be changed (and the corresponding integration test results):

--log2Input WES/WGS=true default=false
--undoSD default=3 WES=2 WGS=3


achevali:
@sooheelee Yes and yes.

sooheelee:
Thanks. I fixed a typo and this should be ready if @achevali says all is good.

sooheelee:
I think no need for further review. I will just merge. We will be going over all these docs again in the near future.

PR #1077, opened 2017-05-22
CNV PlotSegmentedCopyRatio: tag doc, add example command, clarify -SD parameter

sooheelee:
### Pending questions
For PlotSegmentedCopyRatio, what are the "CONTIG_DELIMITER"s? The --sequenceDictionaryFile parameter mentions that contig names should not have these but doesn't define what these are:

```
--sequenceDictionaryFile,-SD:File

                              File containing the reference sequence dictionary (used to determine relative contig

                              lengths). Contigs will be plotted in the order given. Contig names should not include

                              "CONTIG_DELIMITER".Data on contigs not in the dictionary will not be plotted and warning

                              will be thrown.  Required. 
```

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/plotting/PlotSegmentedCopyRatio.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/plotting/PlotSegmentedCopyRatio.java
@@ -90,7 +90,7 @@ public final class PlotSegmentedCopyRatio extends CommandLineProgram {
                     "The tool only considers contigs in the given dictionary for plotting, and " +
                     "data for contigs absent in the dictionary generate only a warning. In other words, you may " +
                     "modify a reference dictionary for use with this tool to include only contigs for which plotting is desired, " +
-                    "and sort the contigs to the order in which the plots should display the contigs."
+                    "and sort the contigs to the order in which the plots should display the contigs.",
             shortName = StandardArgumentDefinitions.SEQUENCE_DICTIONARY_SHORT_NAME,
             optional = false
     )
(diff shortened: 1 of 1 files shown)


sooheelee:
Hey @achevali, any word on this?

achevali:
@sooheelee I'm not sure what those are. I believe Sam Lee added that argument when we upgraded the plotting to be no longer tied explicitly to hg19.

sooheelee:
Thanks @achevali, I will check with @samuelklee then.

src/main/java/org/broadinstitute/hellbender/tools/exome/plotting/PlotSegmentedCopyRatio.java
```diff
@@ -66,7 +87,8 @@
             doc = "File containing the reference sequence dictionary (used to determine relative contig lengths). " +
                     "Contigs will be plotted in the order given. " +
                     "Contig names should not include \"" + CONTIG_DELIMITER + "\"." +
-                    "Data on contigs not in the dictionary will not be plotted and a warning will be thrown.",
+                    " Only data for contigs represented by the dictionary will be plotted. " +
```
samuelklee: Just a minor nitpick, but I would add the missing space to the end of the previous line (rather than the beginning of this line) for consistency.

samuelklee:
See discussion in https://github.com/broadinstitute/gatk-protected/pull/889 for explanation of the `CONTIG_DELIMITER` business.

samuelklee:
Other than the minor nitpick (and unless you want to change `CONTIG_DELIMITER` to something else or expand upon the doc), this is good to go!

sooheelee:
I will just merge this when checks pass.

sooheelee:
Why do my integration tests keep failing @lbergelson @droazen?

sooheelee:
Nevermind, I scoured the Travis log and found the source of the error.

PR #1078, opened 2017-05-22
CNV CallSegments: tag doc and give example command

sooheelee:
### Pending questions

For CallSegments, the --legacy parameter needs clarification.
- Is python ReCapSeg the main legacy data? 
- Can folks use non-python ReCapSeg? Other data? 
- Why is there a "NOTE" that the output is in the tool's format? Is there reason to expect a different format? I'd like to omit this note. Is that okay?

```
--legacy,-LEG:Boolean         (Advanced) Assume that the input seg file is using the legacy format (e.g. generated by
                              python ReCapSeg).  NOTE:  The output will be in the format used by this program -- i.e. no
                              preservation of legacy field names, etc.  Default value: false. Possible values: {true,
                              false} 
```

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/CallSegments.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/CallSegments.java
@@ -3,6 +3,7 @@ package org.broadinstitute.hellbender.tools.exome;
 import org.broadinstitute.barclay.argparser.*;
 import org.broadinstitute.barclay.argparser.Argument;
 import org.broadinstitute.barclay.argparser.CommandLineProgramProperties;
+import org.broadinstitute.barclay.help.DocumentedFeature;
 import org.broadinstitute.hellbender.cmdline.*;
 import org.broadinstitute.hellbender.cmdline.programgroups.CopyNumberProgramGroup;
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/hellbender/tools/exome/CallSegments.java
```diff
@@ -12,17 +13,29 @@
 import java.util.List;
 
 /**
- * Calls segments as amplified, deleted, or copy number neutral given files containing tangent-normalized
- * read counts by target and a list of segments
+ * Calls segments as amplified, deleted or copy number neutral given files containing tangent-normalized
```
LeeTL1220: I would leave the comma.  But I will not stop merging this change over it.

LeeTL1220:
Answers to questions:

> Is python ReCapSeg the main legacy data?

Yes

> Can folks use non-python ReCapSeg? Other data?

There is no such thing as non-python ReCapSeg data.  If you have data in the same format as ReCapSeg you can use it.  The tool is agnostic of how it was generated. 

> Why is there a "NOTE" that the output is in the tool's format? Is there reason to expect a different format? I'd like to omit this note. Is that okay?

I think that is okay.

PR #1079, opened 2017-05-22
CNV AnnotateTargets: Shl doc annotate targets

sooheelee:
### Pending questions

Where can I find a list of the annotations that AnnotateTargets can call upon and their dependencies?


src/main/java/org/broadinstitute/hellbender/tools/exome/CallSegments.java
```diff
@@ -12,17 +13,29 @@
 import java.util.List;
 
 /**
- * Calls segments as amplified, deleted, or copy number neutral given files containing tangent-normalized
- * read counts by target and a list of segments
+ * Calls segments as amplified, deleted or copy number neutral given files containing tangent-normalized
```
davidbenjamin: Is it the GATK standard to omit the Oxford comma after "deleted"?

sooheelee:
Great! Thanks David.

PR #1083, opened 2017-05-22
Mutect2: only output calls that *start* in the current shard

### LINKED ISSUE #1081: duplicate haplotype caller start position bug fix in mutect

droazen:
Previously the tool was outputting calls that were completely contained within
the current shard.

Resolves #1081


droazen:
@davidbenjamin please review

PR #1093, opened 2017-05-23
CNV PadTargets: tag doc and give example command

sooheelee:
### ~~Pending~~ questions
What exactly does the tool do to "behaves intelligently when padding consecutive targets naively would yield overlapping targets"? I think it's important to mention how the tool resolves overlaps in padding.


sooheelee:
I now have the answer this my question -- it divides the overlapping region in half and gives each half to the upstream and downstream targets.

sooheelee:
This is ready for review.

LeeTL1220:
@sooheelee So long as the tests pass, squash and merge away.

PR #1096, opened 2017-05-23
CNV GermlineCNVCaller: tag doc, give two example commands, point to results folders

sooheelee:
### Pending questions

- For @vdauwera:
GermlineCNVCaller's doc-- some of the parameters are described up top and should be moved to the actual parameter's section. I didn't do this because the descriptions up top link to e.g. {@link ReadCountCollectionUtils}, which when put into quotes, I'm not sure how they will appear, given we cannot yet generate the docs via autodoc (getting fixed currently).

It would be great when we will finally be able to generate the docs so I can understand how the formatting appears.

- @cmnbroad clarifies that we will indeed have a tag for experimental tools so I will not add this in the tool doc text itself and will wait to add this tag, which this tool needs.

- [ ] Add experimental tag?
- [x] Upload copyNumberTransitionPriors resources to the bundle (the beta folder).

Now lives at  #`/humgen/gsa-scr1/pub/bundle/beta/GermlineCNVCaller`:
screenshot 2017-05-23 17 33 56

Many thanks to MB for having these on hand.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/coveragemodel/germline/GermlineCNVCaller.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/coveragemodel/germline/GermlineCNVCaller.java
@@ -7,6 +7,7 @@ import org.apache.spark.api.java.JavaSparkContext;
 import org.broadinstitute.barclay.argparser.Argument;
 import org.broadinstitute.barclay.argparser.ArgumentCollection;
 import org.broadinstitute.barclay.argparser.CommandLineProgramProperties;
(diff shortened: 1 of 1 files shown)


sooheelee:
We are waiting for Stacey Donelly's green light to share the resources as I state in the tool doc that example files can be had in the GATK Resource Bundle.

sooheelee:
@mbabadi Just need your review that the wording is to your liking.

sooheelee:
In light of tomorrow's repo merging, have asked and given permission to just merge away.

PR #1103, opened 2017-05-26
Polishing M2 wdls

davidbenjamin:
@takutosato can you review this?


ldgauthier:
Does this do the indel sensitivity too?  If it does, I can throw out the unmerged Palantir branch with my crufty scripts.

davidbenjamin:
@ldgauthier Yes, it does indels, too.

PR #1105, opened 2017-05-26
Change docgen field search order.

cmnbroad:
This works around an issue where the docgen field search code tries to reflect on inaccessible JVM classes such as Throwable, and allows the docgen process to be run locally. This change affects dogcen only.


cmnbroad:
Fixes https://github.com/broadinstitute/gatk-protected/issues/1048. We'll also make Barclay robust to this since it can still happen in error paths, but we should make this change anyway. Also, it will allow the comms team to look at doc locally, and reduces the runtime of docgen.

PR #1115, opened 2017-05-30
Add command line/program to generated VCFs.

cmnbroad:
This covers only tools that are already CommandLineProgram-derived (including HaplotypeCaller/Spark, Mutect2, GenotypeGVCF). Requires a GATK public upgrade.


src/main/java/org/broadinstitute/hellbender/tools/walkers/validation/Concordance.java
```diff
@@ -102,19 +105,22 @@ public void onTraversalStart() {
...
             final VCFHeader truthHeader = getTruthHeader();
             truthHeader.addMetaDataLine(TRUTH_STATUS_HEADER_LINE);
+            defaultToolHeaderLines.forEach(truthHeader::addMetaDataLine);
             truePositivesAndFalseNegativesVcfWriter.writeHeader(truthHeader);
         }
 
         if (truePositivesAndFalsePositivesVcf != null) {
             truePositivesAndFalsePositivesVcfWriter = createVCFWriter(truePositivesAndFalsePositivesVcf);
             final VCFHeader evalHeader = getEvalHeader();
+            defaultToolHeaderLines.stream().forEach(evalHeader::addMetaDataLine);
```
lbergelson: we should be consistent between `forEach` vs `stream().forEach`

droazen:
:+1: merge when ready

src/main/java/org/broadinstitute/hellbender/tools/walkers/validation/Concordance.java
cmnbroad: Yeah, this is unnecessary and inconsistent. Plain forEach will do.

PR #1117, opened 2017-05-30
CNV ConvertGSVariantsToSegments: tag doc, give example command and example output

sooheelee:
Just incorporated what I learned, despite also learning this will be an internal tool.

Please let me know the information is correct @vruano. Thank you.


sooheelee:
I'm going to merge this now because I hear that gatk4 and gatk-protected repos are being merged today and we should close out all PRs that we can. I will come back for review of the documentation changes in the second round.

PR #1118, opened 2017-05-30
CNV CorrectGCBias: Shl doc correct gc bias

sooheelee:
- Besides the usual doc changes, updated the following:
```

        oneLineSummary = "Correct for per-sample GC bias effects",
        summary = "Correct coverage in read counts by estimating per-sample bias as a function of target GC " +
                "content and dividing input coverage by these derived bias curves.",
```

to refer to _per-target_ instead of per-sample.

- In addition, updated the output name of `AnnotateTargets.java` to match the input of CorrectGCBias: targets.annotated.tsv

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/eval/ConvertGSVariantsToSegments.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/eval/ConvertGSVariantsToSegments.java
@@ -7,6 +7,7 @@ import org.apache.commons.lang.math.IntRange;
 import org.broadinstitute.barclay.argparser.Argument;
 import org.broadinstitute.barclay.argparser.ArgumentCollection;
 import org.broadinstitute.barclay.argparser.CommandLineProgramProperties;
+import org.broadinstitute.barclay.help.DocumentedFeature;
 import org.broadinstitute.hellbender.cmdline.StandardArgumentDefinitions;
 import org.broadinstitute.hellbender.cmdline.programgroups.CopyNumberProgramGroup;
 import org.broadinstitute.hellbender.engine.FeatureContext;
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/hellbender/tools/exome/gcbias/CorrectGCBias.java
```diff
@@ -17,24 +18,36 @@
...
+ * java -Xmx4g -jar $gatk_jar CorrectGCBias \
+ *   --input coverage.tsv \
+ *   --targets targets.annotated.tsv \
+ *   --output coverage.gc_corrected.tsv
+ * </pre>
  *
  * @author David Benjamin &lt;davidben@broadinstitute.org&gt;
  */
 @CommandLineProgramProperties(
-        oneLineSummary = "Correct for per-sample GC bias effects",
```
samuelklee: There are arguments for both "per-sample" and "per-target".  Roughly speaking, GC bias is a multiplicative bias that affects coverage at a target as a function of the GC content of that target, but the particular function differs from sample to sample (as is indicated by the `summary` line below).  I think "Correct for sample-specific GC bias" is fine for the `oneLineSummary`.

src/main/java/org/broadinstitute/hellbender/tools/exome/gcbias/CorrectGCBias.java
```diff
@@ -17,24 +18,36 @@
...
-        summary = "Correct coverage in a read counts files by estimating per-sample bias as a function of target GC" +
-                " content and dividing input coverage by these derived bias curves.",
+        oneLineSummary = "Correct for per-target GC bias effects",
+        summary = "Correct coverage in read counts by estimating per-sample bias as a function of per-target GC " +
+                "content and by dividing input coverage by derived bias curves.",
```
samuelklee: The additional `by` feels extraneous to me.  We correct coverage by 1) estimating... and 2) dividing... in sequence.

samuelklee:
Also, don't forget to squash before merging!

src/main/java/org/broadinstitute/hellbender/tools/exome/eval/ConvertGSVariantsToSegments.java
```diff
@@ -33,15 +34,51 @@
 import java.util.stream.Stream;
 
 /**
- * Tool to convert Genotype Strip Variant call file into {@link HiddenStateSegmentRecord} table file.
+ * Converts Genome STRiP (GS) variant call format (VCF) variants into {@link HiddenStateSegmentRecord} CopyNumberTriState format calls.
```
sooheelee: No. I don't git add these extra files so it's not clear to me why they appear in my commit.

src/main/java/org/broadinstitute/hellbender/tools/exome/gcbias/CorrectGCBias.java
```diff
@@ -17,24 +18,36 @@
...
+ * <p>
  * Inputs are
  * 1. read counts file (format described in {@link ReadCountCollectionUtils}).  Since this tool corrects
  *    for multiplicative GC bias effects, these counts should not be in log space i.e. they should represent
  *    raw coverage or proportional coverage upstream of tangent normalization.
  * 2. targets file containing GC content annotation as produced by {@link AnnotateTargets}.  Every target in the input
  *    read counts must be present in the targets file but the reverse need not be true.
  *
- * Output is a read counts file with the same targets (rows) and samples (columns) as the input, corrected for GC bias.
+ * The resulting read counts file has the same targets (rows) and samples (columns) as the input and GC-bias-corrected coverage.
```
sooheelee: Sounds like the tool additionally marks those coverages that it corrected. Ok.

src/main/java/org/broadinstitute/hellbender/tools/exome/gcbias/CorrectGCBias.java
sooheelee: ok

src/main/java/org/broadinstitute/hellbender/tools/exome/gcbias/CorrectGCBias.java
sooheelee: ok

PR #1124, opened 2017-05-31
CNV getbayesianhetcoverage: tag doc, give example command and interval_list, massage summary, prioritize modes

sooheelee:
- Food for thought: The snpIntervals should also accept a VCF of snps.
- For future iteration: should probably mention where the Bayesian math comes in. I assume at the inference.

### DIFF
--- a/src/main/java/org/broadinstitute/hellbender/tools/exome/GetBayesianHetCoverage.java
+++ b/src/main/java/org/broadinstitute/hellbender/tools/exome/GetBayesianHetCoverage.java
@@ -24,35 +24,111 @@ import org.broadinstitute.hellbender.utils.read.ReadConstants;
 import java.io.File;
 
 /**
- * Calls heterozygous sites and outputs information about the called sites.
+ * Calls heterozygous SNPs, counts ref and alt alleles and gives genotype likelihoods.
  *
- * The tool has three modes. The mode is automatically selected based on the provided arguments:
+ * <p>The tool has three modes. The mode is automatically selected based on the provided arguments as follows. </p>
  *
  * <dl>
- *     <dt>NORMAL_ONLY</dt><dd>This mode is chosen if just the normal BAM file is provided. The balanced prior
- *     is used for estimating the likelihood of het sites (see {@link BalancedHeterozygousPileupPriorModel}).</dd>
- *
- *      <dt>TUMOR_ONLY</dt><dd>This mode is chosen if just the tumor BAM file is provided. The heterogeneous prior
- *      is used for estimating the likelihood of het sites (see {@link HeterogeneousHeterozygousPileupPriorModel}).</dd>
- *
- *      <dt>MATCHED_NORMAL_TUMOR</dt><dd>This mode is chosen if both normal and tumor BAM files are provided.
- *      The balanced prior (see {@link BalancedHeterozygousPileupPriorModel}) is used to estimate the likelihood of
- *      Het sites on the normal reads. The base counts on the same het sites are collected and reported from the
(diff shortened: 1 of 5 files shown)


sooheelee:
Ignore this one. See instead https://github.com/broadinstitute/gatk-protected/pull/1125

src/main/java/org/broadinstitute/hellbender/tools/exome/GetBayesianHetCoverage.java
```diff
@@ -24,35 +24,111 @@
...
+ *     READ_DEPTH = total pileup depth at the SNP site
+ *     HET_LOG_ODDS = log odds of the SNP site being heterozygous
+ *     </pre>
+ * </p>
+ *
+ * <p>
+ *     The required --snpIntervals file gives the genomic positions of common population Single Nucleotide Polymorphisms (SNPs)
+ *     in Picard-style interval list format. See https://broadinstitute.github.io/picard/command-line-overview.html#BedToIntervalList for details.
+ *     Here is an example file:
+ * </p>
```
sooheelee: @vdauwera Here is where the autodoc ends and the info below doesn't show.

cmnbroad: @sooheelee Its getting confused by the use of "@" below in "@HD" and "@SQ", which it tries to interpret as javadoc tags. Try replacing the uses of "@" with "{@literal @}", ie, replace "@HD" with "{@literal @}HD".

cmnbroad: Oh, now github interpreted my at-signs as users...

cmnbroad: Also, now that I'm looking at these, I see a few other patterns we should talk about.  I'll add some comments.

src/main/java/org/broadinstitute/hellbender/tools/exome/GetBayesianHetCoverage.java
```diff
@@ -24,35 +24,111 @@
...
+ *     READ_DEPTH,
+ *     HET_LOG_ODDS
+ *     READ_DEPTH = total pileup depth at the SNP site
+ *     HET_LOG_ODDS = log odds of the SNP site being heterozygous
+ *     </pre>
+ * </p>
+ *
+ * <p>
+ *     The required --snpIntervals file gives the genomic positions of common population Single Nucleotide Polymorphisms (SNPs)
+ *     in Picard-style interval list format. See https://broadinstitute.github.io/picard/command-line-overview.html#BedToIntervalList for details.
```
cmnbroad: You might want to put external links like this in anchor tags, so in the end result we;ll get real links instead of seeing the url:

<a href=https://broadinstitute.github.io/picard/command-line-overview.html#BedToIntervalList>BedToIntervalList</a>

src/main/java/org/broadinstitute/hellbender/tools/exome/GetBayesianHetCoverage.java
sooheelee: Thank you.

src/main/java/org/broadinstitute/hellbender/tools/exome/GetBayesianHetCoverage.java
sooheelee: Learning on the job has new meaning.

PR #1125, opened 2017-05-31
CNV GetBayesianHetCoverage: tag doc, give example command and interval_list, massage summary, pri…

sooheelee:
…oritize modes

- Food for thought: The snpIntervals should also accept a VCF of snps.
- For future iteration: should probably mention where the Bayesian math comes in. I assume at the inference.


sooheelee:
Had discussion with @mbabadi for the updated content.

There will be a chance for discussion and more review in the second round of edits. For now, need to merge as many PRs as possible since gatk4 and gatk-protected repos are being merged today.

PR #1126, opened 2017-05-31
CNV GetHetCoverage: mention tool is non-bayesian version of GetBeyesianHetCoverage

sooheelee:
Already had doc tag.


sooheelee:
See https://github.com/broadinstitute/gatk-protected/pull/1127 instead

